PR #34, opened 2015-03-05
split: allow more than one split point

RasmusWL:
In the external language you can now do
`split( (1,1,3), [ 1, 2, 3, 4 ]) = {[1], [], [2, 3], [4]}`

Here indexes are used to split the array, and you will get an
array with the leftover elements.

The index expressions are called "splitexps" in the code.

---

In the internal language, we use sizes to build "multi-take"
```
a = [1,2,3,4]
split( (1,0,2) , a ) = {[1], [], [2,3]}
```

Note that you do _not_ get an extra array with leftover elements.

The size expressions are called "sizeexps" in the code.


Athas:
At dinner today, Cosmin and I discussed that what we need is probably actually not this multi-way `split` in the internal language, but a simpler one that takes a single offset and a single length.  There is no need for more flexibility because `split` does not generate any instructions anyway, so we gain nothing from more flexibility.

Still, we may want the multiway split in the external language, so I am going to merge this.

PR #49, opened 2015-04-21
Troels flattening fixes

Athas:


Athas:
Merged with master.  Works on GHC 7.10.

PR #93, opened 2015-10-21
Merge with master.

Athas:
The biggest difference is that I have created specialised modules for each invariant of the imperative code, including a new one that contains OpenCL (`Futhark.CodeGen.ImpCode.OpenCL`).  This is the one you will use when you get to OpenCL code generation (and also the main motivation for this merge).

I modified your Python code generator to import `Futhark.CodeGen.ImpCode.Sequential`, because right now it's not really a generic code generator.  Once you get further, we will move back to importing the generic `Futhark.CodeGen.ImpCode`, but for now, let's pretend imperative code is our only care in the world.

### DIFF
--- a/docs/language-reference.rst
+++ b/docs/language-reference.rst
@@ -5,8 +5,9 @@ Language Reference
 
 The builtin types in Futhark are ``int``, ``real``, ``bool`` and
 ``char``, as well as their combination in tuples and arrays.  An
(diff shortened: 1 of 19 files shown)


Athas:
(I assigned @DanielGavin because only one person can be assigned, but any of you should merge this pull request as soon as you are able.)

PR #129, opened 2016-03-04
const, closes #123

### LINKED ISSUE #123: Add a `const` construct
I'm thinking [Go-style constants](https://blog.golang.org/constants), implemented initially, as mere nullary functions. So e.g.

    const f64 e = 2.718281828459045235360287471352662497757247093699959574966

Instead of

    fun f64 e() = 2.718281828459045235360287471352662497757247093699959574966

oleks:
for now, implemented as nullary functions

@Athas

### DIFF
--- a/data/tests/vasicek/iobound-mc2.fut
+++ b/data/tests/vasicek/iobound-mc2.fut
@@ -10,11 +10,11 @@ fun float mean([float,n] xs) = sum(map(/float(n), xs))
 
 
 -- Vasicek model parameters.
-fun float r0() = 0.03       -- initial interest rate
-fun float thetaP() = 0.03   -- short-rate mean
-fun float thetaQ() = 0.045  -- long-rate mean
-fun float kappa() = 0.1     -- speed of mean reversion
-fun float sigma() = 0.01    -- interest rate volatility
+const float r0 = 0.03       -- initial interest rate
+const float thetaP = 0.03   -- short-rate mean
+const float thetaQ = 0.045  -- long-rate mean
+const float kappa = 0.1     -- speed of mean reversion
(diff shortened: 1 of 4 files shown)


Athas:
Fair enough, but this is the last such parser hack we'll be accepting, I think.  Tis' not the way to implement a programming language.

oleks:
On Fri, Mar 4, 2016 at 9:45 PM, Troels Henriksen
<notifications@github.com> wrote:
> parser hack

You don't have to merge if you don't agree! That's the point of pull requests.

Anyway, I actually thought it was kind of neat till proven bad. I
often feel constrained by the limited scope of constants in most
languages, while finding the compiler benefits of said constraints
questionable. Not that I've had a look at your in-liner..

Athas:
Oleks <notifications@github.com> writes:

> On Fri, Mar 4, 2016 at 9:45 PM, Troels Henriksen
> <notifications@github.com> wrote:
>> parser hack
>
> You don't have to merge if you don't agree! That's the point of pull
> requests.

Well, I think we should have some form of constants, and now we will be
motivated to improve them.  If we cannot figure something out within a
few months, we'll just cut it again.

> Anyway, I actually thought it was kind of neat till proven bad. I
> often feel constrained by the limited scope of constants in most
> languages, while finding the compiler benefits of said constraints
> questionable. Not that I've had a look at your in-liner..

That's exactly it - we need to figure out a good design for constants.
Right now they have weird quirks:

  * Using them is done with function call syntax (and you can pass
    parameters).

  * They can be mutually recursive.

  * You cannot use them in shape declarations, which to me would
    otherwise be the obvious use case for a constant.

-- 
\  Troels
/\ Henriksen

PR #138, opened 2016-04-11
Add new construct 'write'

nqpz:
Construct: `write(indexes, values, array)`.

Tests are located in `data/tests/write/`.

There are bugs and missing features, all marked with a FIXME in the code.  They include:

  * No certificates for, among other things, bounds checking are generated when internalising.  Instead, just bounds checking is generated when building the kernel.
  * Only partial type checking for the new `Write` "SOAC": For example, the type checker currently does not check if the `values` array aliases the `array` array (which is not allowed, since `write` is in-place).
  * No type checking for the `WriteKernel` kernel.
  * Superfluous allocation of result array.  Currently the Futhark compiler generates code that allocates space for every `write` return array, but the space is not used, since `write` is in-place; see the FIXME in `CodeGen/ImpCode/Kernels.hs`.
  * Not clear to the compiler that the `array` array is always consumed.  In some cases the Futhark compiler has been too clever for its own good, apparently because it didn't know that it was going to consume `array`.  For an example of an ad hoc fix, see the `magic_unique` function in <https://github.com/HIPERFIT/futhark-benchmarks/blob/master/rodinia/bfs/bfs_main.fut>.


nqpz:
Also, two write tests currently fail because of Troels' new memory management runtime.

Athas:
"Niels G. W. Serup" <notifications@github.com> writes:

> Also, two write tests currently fail because of Troels' new memory management runtime.

These fail because you write out of bounds.  The only reason you didn't
notice before was because there was nothing important stored there.

-- 
\  Troels
/\ Henriksen

PR #145, opened 2016-04-19
Fix for OSX

maccam912:
This compiles fine for me on OSX. Travis-CI should confirm it can still compile fine on Linux.


Athas:
You have an unnecessary paranthesis in `(System.Info.os == "darwin")`.  Travis will complain about that (it's yelled at me plenty of times).  If you fix that, I'll merge your patch.

And thanks!

Athas:
Thanks for the patch!

PR #154, opened 2016-04-26
finished working on type aliasing

MikkelStorgaard:
Merge this to add the type aliasing feature to Futhark


Athas:
Good enough.  I'll take it from here.

Athas:
Actually, this code is pretty good.

PR #166, opened 2016-05-04
zip0-test: fix expected output

fenollp:


Athas:
Oh, yes, I guess I was lazy when I wrote that test.  The two forms are identical after desugaring, so they both work, but your version is neater.  Thanks!

PR #197, opened 2016-06-04
Extend 'write'

nqpz:
This pull request encompasses several changes and additions:

  + The `write` construct in the outer language can now take tuples of arrays as input.
  + `Write` and `WriteKernel` now uses a Lambda to make fusion easier.
  + The `Fusion.LoopKernel` module now also does horisontal write-write fusion and producer-consumer map-write fusion.
  + The `SOACS.Simplify` module now also eliminates `replicate` expressions from `Write` expressions.
  + The `Kernels.Simplify` module now also eliminates `iota` expressions from `WriteKernel` kernels.

I have also updated the language reference documentation for `write`.

There are also some minor write-related fixes.


Athas:
Could you please squash this commit and rebase it on master?  Currently it would be hell to bisect.

nqpz:
See https://github.com/HIPERFIT/futhark/pull/198 instead.

PR #198, opened 2016-06-04
Extend 'write'

nqpz:
The same as https://github.com/HIPERFIT/futhark/pull/197 except squashed.


Athas:
Looks good to me if Travis agrees.

PR #205, opened 2016-06-17
Switch to new array syntax

### LINKED ISSUE #179: Improve array syntax
At present, the type for an NxM array of integers is `[[int,M],N]`.  It is kind of annoying that you have to read it right-to-left to get the natural shape.  One solution is to instead use an `[N,[M,int]]` notation, but maybe someone can come up with something better.

Athas:
Fixes #179.


Athas:
Note that this breaks just about every Futhark program ever written.  See #179 for a program that can convert the old syntax to the new.

PR #231, opened 2016-09-04
Simplify the syntax for 'write'.

nqpz:
This is now the syntax in the source language:

If n > 1:
```
write (is_1, ..., is_n) (vs_1, ..., vs_n) (as_1, ..., as_n)
```

If n = 1:
```
write is vs as
```

I have also changed the write tests, and added a new one.  All tests pass.  All write-using programs in the futhark-benchmarks repo pass.

I have also documented the change in the docs.

I have also simplifed the write-using programs in the futhark-benchmarks repo, but I won't commit those until this has been merged.  The current versions should still work with this new syntax, since all of them use write expressions where n = 1 (I think).


src/Futhark/Internalise.hs
```diff
@@ -642,11 +642,14 @@ internaliseExp desc (E.Copy e _) = do
...
-internaliseExp desc (E.Write i v as loc) = do
-  sis <- internaliseExpToVars "write_arg_i" i
-  svs <- internaliseExpToVars "write_arg_v" v
-  sas0 <- forM as $ internaliseExpToVars "write_arg_a"
-  sas <- mapM (ensureSingleElement "Futhark.Internalise.internaliseExp: Every I/O array in 'write' must be a non-tuple.") sas0
+internaliseExp desc (E.Write is vs as loc) = do
+  let internaliseWriteArrays name xs =
+        mapM (ensureSingleElement
+              "Every I/O array in 'write' must be a non-tuple.")
+        =<< mapM (internaliseExpToVars name) xs
```
Athas: Use a single mapM and `<=<` (doesn't `hlint` complain about this?).

Athas:
It does not seem right that

```
 write (indexes, indexes) (values0, values1) (array1, array2)
```

works but

```
let is = (indexes, indexes)
let vs = (values0, values1)
let as = (array1, array2)
in write is vs as
```

is a type error.

Athas:
Why does the source language `write` support multiple arrays in the first place?  I can't think of why, and neither could Cosmin.  (The core language `write` is a different matter, of course.)

nqpz:
I suppose it is not necessary, since they will be fused anyway.  Should I just remove that feature?

Athas:
Yes please.  Always simplify!  This is what we should aim for: http://www.chaosmonger.com/blog/simplify.jpg

Athas:
Much better now!

PR #274, opened 2017-01-11
Add git info to --version

RasmusWL:
Output of `futhark --version` will now be like the following. I think the format for the timestamp of the commit is horrible, but I do think it is important to have, as it tells you how recent a version was used which is usually what you want to know for non debugging purposes. 

```
Futhark 0.1
git: 8791fd4 (Wed Jan 11 18:26:23 2017 +0100)
(C) HIPERFIT research centre
Department of Computer Science, University of Copenhagen (DIKU)
```

If the branch is not "master", the branchname will be inserted before the hash, and if any files have been modified this will be displayed at the end of the line

# Caveats 

Due to the way the `gitrev` package works (using TemplateHaskell),
`Futhark.Version` and `Futhark.Util.Options` will be recompiled on *every*
compilation. This kinda sucks, and is why I didn't just merge it right away.

### DIFF
--- a/futhark.cabal
+++ b/futhark.cabal
@@ -51,7 +51,8 @@ Library
                  neat-interpolation >= 0.3,
                  file-embed >= 0.0.9,
                  directory,
-                 directory-tree
+                 directory-tree,
+                 gitrev >= 1.2.0
 
   Exposed-modules: Futhark.Actions
                    Futhark.Analysis.AlgSimplify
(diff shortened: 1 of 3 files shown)


Athas:
Those rebuilds seem pretty harmless.  It's a small module, and a drop in the bucket compared to the ridiculous link times we already suffer.

People can use intero-mode or GHCi for snappier feedback.

PR #295, opened 2017-02-05
Binary input

RasmusWL:
Finally it is here, a binary input format for Futhark.

Besides implementing reading the binary input format in the C runtime system, I have also updated `futhark-dataset` so it can generate data in the binary format. 

I've added documentation which should explain the format in details.

I have not added any form of automatic testing, which would probably be nice.


Athas:
Looks good!  Don't worry about the tests; I'll eventually convert most of our benchmarks to use this format, so testing will be implicit.

The only serious flaw I can see is the use of magic enum values for indicating primitive types.  This is a bad idea, as adding new primitive types will result in invalidation of existing data files.  I much prefer a textual representation, e.g. `i32` or `f32`.  If you wish, make them all four bytes long (e.g. "\0i32") for ease of parsing (it's not about being readable for humans, but about being robust in the face of changes).  Since these only appear in the array header, the impact on performance should be neglible.

Athas:
Why not encode scalars as 0-dimensional arrays?  That's conceptually what they are, and it seems like it would even save some space.

RasmusWL:
Encoding scalars as 0-dimensional arrays would make sense. I think I didn't realise that empty arrays would still contain size information for all dimensions (which would all be 0).

Encoding the primitive type as a string instead of magic enum values does make sense.

I'll try to fix both cases soon!

Athas:
Please do the type encoding as a *fixed-width* string, i.e. four bytes and zero-padded.  This makes parsing much easier, as the header becomes constant-size (ignoring the array shape).

RasmusWL:
I did not see your comment before now, but I did indeed make it 4 characters long. I used space for padding, as it made alignment in the code nicer than NULL bytes.

Athas:
Spaces are fine too.  Doing a review now.

docs/binary-data-format.rst
```diff
@@ -0,0 +1,63 @@
...
+characters makes it easy to use binary input for some arguments, and text input
+for others.
+
+The general format has this header::
+
+  b <version> <num_dims> <type>
+
+Where ``version`` is the version of the binary format used for encoding
+(currently 1), ``num_dims`` is the number of dimensions in the array (0 for
+scalar), and ``type`` is a 4 characters describing the type of the values(s) --
```
Athas: Typo here ("a 4 characters").

docs/binary-data-format.rst
```diff
@@ -0,0 +1,63 @@
+.. _binary-data-format:
+
+Binary Data Format
+==================
+
+Futhark supports a binary format for input files, which this page
```
Athas: I would prefer an introduction more like:

Futhark programs compiled to an executable support both textual and binary input.  Both are read via standard input, and can be mixed, such that one argument to an entry point may be binary, and another may be textual.  The binary input format takes up significantly less space on disk, and can be read much faster than the textual format.  This chapter describes the binary input format and its current limitations.  The input formats (whether textual or binary) are not used for Futhark programs compiled to libraries, which instead use whichever format is supported by their host language.

(Then the "currently reading..." stuff about limitations.)

docs/binary-data-format.rst
```diff
@@ -0,0 +1,63 @@
...
+scalar), and ``type`` is a 4 characters describing the type of the values(s) --
+see below for more details.
+
+Encoding a scalar value is done by appending the binary little endian
+representation of it::
+
+  b <version> 0 <type> <value>
+
+To encode an array we must encode the number of dimensions ``n`` as a single
+byte, each dimension ``dim_i`` as an ``int64``, and finally all the values in
```
Athas: Instead of ``int64``, clarify "a signed 64-bit two's complement little-endian integer" (or make it unsigned, which probably makes more sense).

rts/c/reader-bin.h
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    void* tmp = realloc(*data, elem_count * elem_size);
+    if (tmp == NULL) {
+        panic(1, "binary-input: Failed to allocate array of size %i.\n",
+              elem_count * elem_size);
+    }
+    *data = tmp;
+
+    if (IS_BIG_ENDIAN && elem_size != 1) {
```
Athas: Instead of this, why not just always fread() and then flip the bytes if on a big-endian platform?  I think your control flow is too complicated this way.

src/Futhark/Representation/Primitive.hs
```diff
@@ -121,6 +121,8 @@ data PrimType = IntType IntType
               | Cert
               deriving (Eq, Ord, Show)
 
+-- If you add a value here or change the meaning, you will inflict great damage
+-- on the parsing of binary input files. You have been warned!
```
Athas: This is hopefully not true anymore.

docs/binary-data-format.rst
RasmusWL: Yes, your introduction is better! :+1:

docs/binary-data-format.rst
RasmusWL: Fixed.

docs/binary-data-format.rst
RasmusWL: Fixed. 

Right now the C runtime system uses an `int64_t` to store the size for each dimension. Shouldn't we change that to an `uint64_t` as well?

rts/c/reader-bin.h
RasmusWL: I did not agree at first, but tried it out. I think it's nicer this way.

src/Futhark/Representation/Primitive.hs
RasmusWL: Nope, fixed.

PR #320, opened 2017-03-12
Make Python programs able to read binary input

### LINKED ISSUE #296: Add support for binary input to Python
This is necessary before we can migrate the benchmarks to binary data sets.

RasmusWL:
Fixes #296

### DIFF
--- a/rts/python/panic.py
+++ b/rts/python/panic.py
@@ -1,0 +1,5 @@
+def panic(exitcode, fmt, *args):
+    # FIXME: Insert the name of the Futhark program, but I don't know where it can be found
+    sys.stderr.write('<Futhark program>: ')
+    sys.stderr.write(fmt % args)
+    sys.exit(exitcode)
--- a/rts/python/panic.py
+++ b/rts/python/panic.py
@@ -1,5 +1,4 @@
 def panic(exitcode, fmt, *args):
-    # FIXME: Insert the name of the Futhark program, but I don't know where it can be found
-    sys.stderr.write('<Futhark program>: ')
+    sys.stderr.write('%s: ' % sys.argv[0])
     sys.stderr.write(fmt % args)
     sys.exit(exitcode)
--- a/rts/python/reader.py
+++ b/rts/python/reader.py
@@ -22,7 +22,7 @@ def get_char(f):
         return c
 
 def get_chars(f, n):
-    s = ''
+    s = b''
     for _ in range(n):
         s += get_char(f)
     return s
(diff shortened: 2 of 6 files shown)


rts/python/panic.py
```diff
@@ -0,0 +1,5 @@
+def panic(exitcode, fmt, *args):
+    # FIXME: Insert the name of the Futhark program, but I don't know where it can be found
```
Athas: sys.argv[0]

rts/python/reader.py
```diff
@@ -257,8 +264,263 @@ def verify_array_dims(l, dims):
...
+    for i in range(rank):
+        bin_size = read_bin_u64(f)
+        elem_count *= bin_size
+        shape.append(bin_size)
+
+    bin_reader = FUTHARK_PRIMTYPES[bin_type_enum]['bin_reader']
+
+    arr = np.empty(elem_count, dtype=ctype)
+    for i in range(elem_count):
+        arr[i] = bin_reader(f)
```
Athas: Isn't this extremely inefficient?

RasmusWL: it should be much more efficient than using textual input data :) but yes, it should be possible to do something better, but I am not sure how.

Athas: Fine; nothing can be worse than the old parser.

RasmusWL:
The old changes did *not* work with Python 3. It has been a bit of a mess to make it work, but now it does. At least for the few cases that I tried.

The problem arises from the fact that Python 3 will (by default), try to read characters as utf8, and will thus complain if binary data is used. So, for Python 3 we can use `sys.stdin.buffer` to get access to the raw bytes. This means that all string comparisons must now be with `b'<string>'` instead of just `'<string>'`. Another point of trouble is that for Python 3 if `type(b) == bytes` then `type(b[0]) == int` -- so when looping over all the characters of a string, I do this mess ` [b[i:i+1] for i in range(len(b))]`

You might be thinking: Couldn't this be avoided by *only* using bytes when we're dealing with binary input? but alas, `peek(1)` can return *any* number of characters, so we will get into trouble if we tried to mix using both textual and binary input to a program, which seems like a handy feature to have.

defining `input_stream` as the last thing in `reader.py` does seem like a hack, but I didn't know where to place the following lines otherwise (handling that python 2 does not have the `buffer` property on `sys.stdin`):

```
if sys.version_info >= (3,0):
     input_stream = sys.stdin.buffer
 else:
     input_stream = sys.stdin
```

Athas:
As I understand it, Python 2/3 compatibility frequently involves some hacks.  This one doesn't look so bad, all things considered.  So, it works now?

rts/python/reader.py
svip: @Athas Watch me, poor man!

titouanc: I think it could be faster by unpacking the whole array in one pass (either `struct.unpack(fmt*array_size, input)` or using `np.fromfile(f, dtype, array_size)`

RasmusWL:
at least it seems to work on my machine :)

PR #321, opened 2017-03-13
[add] fucolor utility: colorize vars in futhark output

titouanc:
As a human, I find it easier to read Futhark output with (generated) variable names in colorz.

Works well in 256-colors terminals, with all kind of futhark outputs (C/py OpenCL, futhark IL, ...)

2017-03-13-115733_911x1365_scrot

### DIFF
--- a/tools/futcolor
+++ b/tools/futcolor
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+
+"""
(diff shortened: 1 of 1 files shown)


Athas:
That's a fun tool.  Maybe it can be useful for debugging.

PR #334, opened 2017-05-03
Extend the experimental memory block merging optimisation.

nqpz:
Now set the environment variable `MEMORY_BLOCK_MERGING` to `1` to enable the optimisation in all tools.

Also:

  + Add more tests, and move the existing ones a little around.
  + Make the debug printing more readable.

Running `MEMORY_BLOCK_MERGING=1 futhark-test tests/memory-block-merging` shows that not all the 27 tests in the directory pass with the optimisation enabled.  Some of the bugs seem to have easy fixes, but some of them maybe don't.  (They all pass without the optimisation enabled.)

This is very much work in progress.


coancea:
I have pulled, but I found only 20 (instead of 27) tests and all are 
passing (?)

What is going on?

Cosmin

coancea:
now you're talking. Thank you!

coancea:
The invalid tests have two causes:

1. One you are aware of: the existential context of if/loop statements, 
which needs to be filtered.
     (I have seen your FIXME comment in `transformStmt`)

2. Most tests fail because the code generation does not use the third 
(FreeVarSubsts) field of Coalesced.
     See DataStructs.hs:

    `type FreeVarSubsts = M.Map VName (ExpMem.PrimExp VName)`

    `data Coalesced = Coalesced CoalescedKind -- the kind of coalescing
                                ArrayMemBound -- destination mem_block 
info @f_m_x[i]@ (must be ArrayMem)
                                FreeVarSubsts -- substitutions for free 
vars in index function`

     For every pattern element (and FParam) we would need to:
         a) generate a new integer identifier for every key in its 
FreeVarSubsts (`newIdent`)
         b) substitute the new identifier name for the key in the new 
index function (`substituteNames`)
         c) generate statements that binds the new identifier to their 
corresponding PrimExp
             (`primExpToExp`, but this requires a MonadBinder monad)

    Am I correct Troels?

    This would solve most of the cases except for one like in 
blk-chain2.fut:
     `let b       = map (+i1) a
      let ind1    = z[k,i1,i2] - i1 -- i1 + 1
      let u[ind1] = b`

    because the `PrimExp` datatype cannot represent the array-indexing 
expression `z[k,i1,i2]`,
    and as such translating ind1 before the definition of `b` will fail...

So, what do you (Niels and Troels) think?
Should we use the current support or should we get rid entirely of the 
`FreeVarSubsts` field and
rely on an aggressive hoisting of math computations instead?

Cosmin

Athas:
Right, you'll need to create them as bindings.

I'm not sure what you mean by "code generation" in this context.

We could also extend the PrimExp representation to permit array indexing.  As long as the result is a scalar, it shouldn't be a problem.  Will require a little extra care in code generation (*actual* code generation, as in the GenericC module), but nothing we can't do if it saves work elsewhere.

coancea:
Ok, so I inlined the substitutions directly in the index function such that all frevarsubst maps are now empty.
(Next step would be to remove them completely from representation and simplify the translation part since no new statements need to be generated).

There is now (only) one test that does not validate however: misc/if-ip-concat-1.fut, which passes type-checking but produces an incorrect result. We need to take a deeper look at it!

nqpz:
On Fri, 05 May 2017 16:45:49 -0700, Cosmin wrote:
> Ok, so I inlined the substitutions directly in the index function such that all frevarsubst maps are now empty.
> (Next step would be to remove them completely from representation and simplify the translation part since no new statements need to be generated).

OK.  I just removed the use of FreeVarSubsts in the MemoryBlockMerging
module, so it's fine if you remove it from the representation.

> There is now (only) one test that does not validate however: misc/if-ip-concat-1.fut, which passes type-checking but produces an incorrect result. We need to take a deeper look at it!

Right!

PR #341, opened 2017-05-16
Allow underscores in numeric literals

samrat:
Changes to lexer and runtime for this issue: https://github.com/HIPERFIT/futhark/issues/292

- Change Lexer.x to allow underscores in both integers and floats
- Update runtime systems(reader.h and reader.py)
- Update language reference
- Add tests


Athas:
Looks good to me, thanks!

Issue #344, opened 2017-06-08
Add ceil, floor, trunc, and round to futlib/math.real module type


melsman:
Here is a possible implementation of ceil:

````futhark
let ceil (x:f64) : f64 =
  let i = i64(x)
  let ix = f64(i)
  in if x >= 0.0 then
        if ix < x then f64(i+1i64) else x
     else if ix > x then f64(i+1i64) else x
````

melsman:
Fixed by #349

PR #349, opened 2017-06-15
floor, ceil, trunc, and round operations for f64 and f32 modules

melsman:
fixes issue #344

### DIFF
--- a/futlib/math.fut
+++ b/futlib/math.fut
@@ -70,6 +70,11 @@ module type real = {
 
   val log: t -> t
 
+  val ceil : t -> t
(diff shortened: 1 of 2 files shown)


melsman:
An alternative design could be to have the operations return integer values instead. Such a design would require us to include an associated int sub-module within the f64 and f32 modules...

Also, currently, the functions do not operate reasonably on values that cannot be represented as integers of the same size as the float...

Athas:
I find this design to be sound.  The best way to convert from floating-point to an integer is domain-sensitive, and should be left up to the application programmer.

PR #350, opened 2017-06-16
Futlib Sobol library

melsman:
Let's discuss the design before merging...


Athas:
There is no reason to have a copyright header for the .fut files - the one in the LICENSE file should suffice (or else we should add it to the other files as well).

It's not clear to me which restrictions are on the sizes of the arrays of a module implementing `sobol_dir`.  If they have to agree in their outer size, maybe something like this would be better:

```
module type sobol_dir = {
  val asm = [](u32,i32,[]u32)
}
```

Of course, this is the kind of hack you normally see in languages without shape annotations.  So maybe this is better:

```
module type sobol_dir = {
  val n: i32
  val a: [n]u32
  val s: [n]i32
  val m: [n][]u32
}
```

PR #376, opened 2017-07-14
Change ranges to produce empty arrays when second element is invalid

Qata:
This PR brings our ranges into line with Haskell's (i.e. `[9,9..20]` produces `[]` in Haskell).
Please point out if I've done something completely wrong, this is my first time touching a compiler.


Qata:
This PR brings our ranges into line with Haskell's (i.e. `[9,9..20]` produces `[]` in Haskell).

src/Language/Futhark/TypeChecker/Terms.hs
```diff
@@ -536,7 +536,12 @@ checkExp (Range start maybe_step end loc) = do
   let start_t = toStructural $ typeOf start'
   maybe_step' <- case maybe_step of
     Nothing -> return Nothing
-    Just step -> Just <$> (require [start_t] =<< checkExp step)
+    Just step -> do
+      case (start, step) of
+        ((Literal (SignedValue (Int32Value x)) _), (Literal (SignedValue (Int32Value y)) _)) -> do
+          when (x == y) $ warn loc "First and second element of range are identical, this will produce an empty array."
```
Athas: Why only warn if they are the same 32-bit integer?  For any expression `e`, `[e..e...]` will produce an empty array (unless `e` itself fails first).  You could warn whenever `start==step`.

src/Futhark/Internalise.hs
```diff
@@ -419,46 +419,59 @@ internaliseExp desc (E.Range start maybe_second end _) = do
...
-  distance <- letSubExp "distance" $
-              I.If downwards (resultBody [down_distance]) (resultBody [up_distance])
-              [I.Prim $ IntType it]
-  num_elems_maybe_neg <- letSubExp "num_elems_maybe_neg" $ I.BasicOp $ I.BinOp (I.SDiv it) distance step
-  num_elems <- letSubExp "num_elems" $ I.BasicOp $ I.BinOp (I.SMax it) zero num_elems_maybe_neg
-
-  pure <$> letSubExp desc (I.BasicOp $ I.Iota num_elems start' step it)
+              subtracted_step <- letSubExp "subtracted_step" $ I.BasicOp $ I.BinOp (I.Sub it) second' start'
+              case (start', second') of
+                (Constant (IntValue (Int32Value x)), Constant (IntValue (Int32Value y)))
```
Athas: Why this special case for when `start'` and `second'` are integers?  Your computation of `same_step` below should work in all cases, and the simplifier will constant-fold the branch away if they turn out to be integers at some point.

In general, since Futhark is such an aggressively optimising compiler, the philosophy is to keep passes such as the internaliser as simple as possible, and let the specialised optimisation passes deal with cleaning up the code.

Athas:
Also, the style police on Travis found some minor issues: https://travis-ci.org/HIPERFIT/futhark/jobs/253493546#L1536

Athas:
Another thing, could you also update the language reference (docs/language-reference.rst) to mention this behaviour?

As an aside, I don't think this is how Haskell behaves: `[9,9..20]` produces an infinite list, not an empty one.  That's not really an option for us, though, and perhaps empty lists are better than error.

Qata:
Oh I misremembered how Haskell does it, but yeah, since it's not an option it'd be good to have this defined behaviour instead.

src/Futhark/Internalise.hs
Qata: The speciality is mistakenly for integers, it's intended to be for constants to eek out the tiniest bit more speed in compiled code by leaving out the `If` branch. Should I just remove it and let the optimiser take care of it?

Athas: Yes, the optimiser will deal with this.  The generated code should be identical.  If it is not, that is a bug in the optimiser.

Athas:
Could you also add a test case to `tests/range0.fut` to verify your intended behaviour?

src/Language/Futhark/TypeChecker/Terms.hs
Qata: Since `start` and `step` don't conform to `Eq`, should I warn when `(Literal x _, Literal y _) && x == y` and also `(Var x_name _ _, Var y_name _ _) where x_name == y_name`?

Athas: Oh, I forgot that we don't have an `Eq` instance for ASTs.  Should probably fix that one of these days.  In that case, just do as best you can.

Qata:
Okay so I'm finally completely stumped.
For the final line `pure <$> letSubExp "range_step" (I.If step_invalid (resultBody [empty_array]) (resultBody [array]) [...])` I have no idea what to put in the `[ExtType]` array.

Athas:
You have to put the return type of the branch.  If that branch is computing the step integer (rather than an array), then you put `[Prim int32]`.

Qata:
It's computing the array, since we want an empty array when the value is invalid.

Athas:
In that case, you want something like ``staticShapes [IntType it `arrayOfRow` array_length]``.  However, I'm not sure that's the best way to do it, as it is slightly harder to do array optimisations across branches.  It is better to compute the desired size with a branch (that checks whether the step is zero), but leave the actual `iota` unbranched.

Athas:
If you wish to avoid explicit branches entirely, this can probably be done with the use of `SMin`, `SMax`, and `Abs` operations.

Qata:
I've added tests, one of them is failing due to #378 .

Athas:
I fixed #378, but you will have to re-base your patch because some things changed.  Also turned out that I forgot to commit my `tests/range0.fut` in the first place.  Oops.

Qata:
All fixed up!

PR #384, opened 2017-07-30
Fix integer division in the Python backend.

nqpz:
Previously, division of two integers would result in a float.

The `//` integer division operator used in the fix works with both Python 3 and Python 2 (at least Python 2.7).

### DIFF
--- a/rts/python/scalar.py
+++ b/rts/python/scalar.py
@@ -22,6 +22,13 @@ def unsigned(x):
   else:
     return np.uint64(x)
 
+def divide_dwim(x, y):
+  t = type(x)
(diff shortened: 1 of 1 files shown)


Athas:
Don't we have the property that all scalars are of numpy types, and can thus just use numpy.floor_divide?

nqpz:
We cannot use numpy.floor_divide for numpy floats.

nqpz:
For example, `numpy.floor_divide(3.0, 2.0)` returns 1.0, which is not what we want.

Athas:
We never call sdivN and udivN with floats - these are signed and unsigned integer division.

nqpz:
Right.

PR #385, opened 2017-07-30
Fix integer division in the Python backend.

nqpz:
Previously, division of two integers could result in a float.

### DIFF
--- a/rts/python/scalar.py
+++ b/rts/python/scalar.py
@@ -29,13 +29,13 @@ def ashrN(x,y):
   return x >> y
 
 def sdivN(x,y):
-  return x / y
+  return x // y
 
 def smodN(x,y):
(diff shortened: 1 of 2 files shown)


Athas:
Do you have a test case for this?  I'm disturbed that our existing tests did not catch this.

nqpz:
Test committed.  If you compile it with futhark-py or futhark-pyopencl, and then run it with python3 (important), you will get a `TypeError: 'numpy.float64' object cannot be interpreted as an integer`.

Since such programs default to being run with Python 2, this is only relevant when the `--library` flag is used together with Python 3 modules.

Athas:
This kind of thing is why I hope I'll never have to generate code in a unityped language again.

PR #387, opened 2017-08-07
Allow multiple structure tests.

nqpz:
I would like to use this feature.  Is it okay?

### DIFF
--- a/src/Futhark/Test.hs
+++ b/src/Futhark/Test.hs
@@ -63,7 +63,7 @@ data ProgramTest =
               , testAction ::
                    TestAction
               , testExpectedStructure ::
-                   Maybe StructureTest
+                   [StructureTest]
               }
   deriving (Show)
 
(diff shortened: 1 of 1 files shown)


Athas:
When is this useful?

nqpz:
On Mon, 07 Aug 2017 07:43:07 -0700, Troels Henriksen wrote:
> When is this useful?

I want to check that a program has a specific number of alloc statements
both after its sequential CPU compilation pipeline and after its full
GPU compilation pipeline.

The memory block merging algorithms need to treat
first-order-transformed programs differently from kernel-extracted
programs, so having e.g. a `structure cpu` test that passes does not
mean that a `structure gpu` will pass.  It would be useful to be able to
write both structure tests in the same program.

Typically it should be the same number of allocs for both pipelines, so
another approach might be to add a `structure agnostic` field (with a
better name) that would always run it for both pipelines, but that seems
messy.

Issue #388, opened 2017-08-08
Writing to unique input parameter arrays in entry points causes errors in the generated C code


nqpz:
This is related to the work-in-progress memory block optimisations.

Consider this `main` function:
```
let main (xs: *[#n]i32): [n]i32 =
  map (+ 1) xs
```
In both the CPU and GPU pipelines, this will result in a program with a single memory allocation, namely one for the map.  I have an optimisation that currently reasons that since `xs` is unique, the map can try reusing the memory of `xs` instead of requiring a new allocation.  After size and index analysis, the compiler will accept the reuse, and a subsequent simplifier will then remove the allocation statement.

Reusing unique parameter memory seems like a trick that is okay to do (assuming proper interference analysis etc. is also performed), since the caller will not use the memory again.  There are at least two cases:

  + In stand-alone generated programs, entry point parameter memory can always said to be "unique" even when the programmer has not manually declared them to be so, in that its input memory contains the parsed values and is not directly controllable by the user.  The optimisation will always be useful in this case, and it cannot confuse any user.
  + In generated libraries, reusing unique entry point parameter memory means that the programmer using the library must make sure not to reuse the memory sent to library -- but only if the iibrary programmer chooses to manually make input parameters unique.  I think this is a slippery slope.

So, the concept of unique parameters in *entry point* functions seems a bit murky.  I think the best approach is to only perform the optimisation for stand-alone programs, not libraries.  I have a few programs that benefit from this feature.  In any case, my problem is this:

When a generated program is run with `-r N` -- used by e.g. `futhark-bench` -- it performs both a warmup run, and, if N > 1, runs the entry point function more than once, which means that reusing input memory in one run can cause the next run to give a wrong result.  This is easy to fix by copying all unique memory blocks before each run.  This will give an input copying overhead, but it only relates to benchmarking, and the effect is a possibly reduced memory footprint.

Opinions?  Downside: Slightly more copying when benchmarking.  Upside: Reduced memory footprint in some standalone programs.

Athas:
Yes, I have known about this for a while.  You don't actually need memory block merging to show this; the following program will produce spurious results when benchmarked:

```
let main (xs: []i32) =
  xs with [0] <- xs[0] + 1
```

I did not bother fixing it, since it did not occur in practise.  But now it does.

I don't think the situation is murky at all.  The right solution is to copy unique parameters when running standalone executables.  In libraries, it is fine to make destructive updates to the input - that's what the uniqueness types mean!

[...]

PR #389, opened 2017-08-09
Copy entry function memory blocks when benchmarking.

nqpz:
In the generated code, C or Python: If performing a warmup run or
running more than one time, first copy any parameter memory blocks
before each run.  This ensures that programs that write to unique
input arrays do not cause a problem in future runs.

The copying occurs before the timing begins.

Currently (when in benchmarking mode) the compiler copies every
parameter memory block.  It would be more efficient to copy only those
associated with a unique array.  That would seemingly mean some
additions to the ImpCode representation.

Related to issue #388.

### DIFF
--- a/src/Futhark/CodeGen/Backends/GenericC.hs
+++ b/src/Futhark/CodeGen/Backends/GenericC.hs
@@ -587,11 +587,11 @@ mainCall pre_timing fname (Function _ outputs inputs _ results args) = do
   ret <- newVName "main_ret"
   refcount <- asks envFatMemory
   let readstms = readInputs refcount args
-  (argexps, prepare) <- collect' $ mapM prepareArg inputs
-  (argnames_mem_copies, prepare_mem_copies) <- collect' $ mapM prepareMemCopy inputs
-  (argexps', copy_mem_before_run) <-
-    collect' $ sequence $ zipWith3 copyMemBeforeRun inputs argexps argnames_mem_copies
-  free_mem_copies <- collect $ zipWithM_ freeMemCopy inputs argnames_mem_copies
+  (argexps_mem_copies, prepare) <- collect' $ mapM prepareArg inputs
+  let argexps = zipWith fromMaybe
+                (map (\p -> [C.cexp|$id:(paramName p)|]) inputs)
+                argexps_mem_copies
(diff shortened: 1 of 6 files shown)


src/Futhark/CodeGen/Backends/GenericC.hs
```diff
@@ -677,6 +693,40 @@ prepareArg (MemParam name (Space sid)) = do
...
+copyMemBeforeRun (MemParam name space) arg (Just arg_mem_copy) = do
+  copy <- asks envCopy
+  let size = [C.cexp|$id:name.size|]
+      dest = rawMem' True arg_mem_copy
+      src = rawMem' True arg
+  case space of
+    DefaultSpace ->
+      stm [C.cstm|memmove($exp:dest, $exp:src, $exp:size);|]
+    _ ->
+      copy dest [C.cexp|0|] space src [C.cexp|0|] space size
```
Athas: Can't you use `copy` no matter the destination space?

Athas:
I think a simpler way to do this in the C backend is to move `$items:prepare` to just before the main function is called, and modify `prepareArg` to always copy, even if the destination is in `DefaultSpace`.  This reduces code duplication significantly.

Probably something similar is possible in the Python backend.

src/Futhark/CodeGen/Backends/GenericC.hs
nqpz: `SequentialC` does not have a custom `GenericC.opsCopy` function, and `COpenCL` does not handle copying to `DefaultSpace`, but I suppose I could extend the two modules to have `copy` work with all spaces, and also for the Python backend.

Athas: You should add a universal copy function that uses opsCopy if necessary. It's amazing we don't already have that.

nqpz:
I have simplified the copying.  I am not sure what you mean by moving `$items:prepare` and have `prepareArg` always copy.  Shouldn't we only copy when absolutely necessary, i.e. if more than one run will be performed?

Athas:
Does it matter?  I'd rather have simpler code.  We don't measure this copy, and the parsing is significantly slower anyway.  Such a unification would allow you to get rid of `copyMemBeforeRun` (and its siblings).

Issue #382, opened 2017-07-22
Add hexadecimal float literals


Athas:
Read the justification for a similar proposal for GHC: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0004-hexFloats.rst

This should be doable as a mere lexer change.

samrat:
My WIP commit for this issue: https://github.com/samrat/futhark/commit/dfd5e9e206124b589dbe10aaaffcb3d8f14256e5. Comments welcome, since I'm quite new to Haskell.

Questions: 
- Will this syntax be supported: `0xff.ffp2f32` ? My current commit will always create a `f64`. There is a tradeoff here: if we support specifying `f32`, the exponent will always have to be present(in order to disambiguate the "f"-- see https://gcc.gnu.org/onlinedocs/gcc/Hex-Floats.html)
- Will hexadecimal floats also be supported as input? ie. will the Python & C readers also need to support hexadecimal floats?

Athas:
I think we should definitely support the numeric suffixes.  If that requires that the exponent is required, so be it.

I would definitely prefer that the Python and C readers are also extended, but it is not critical.  It can be made a separate issue if you'd rather not spend your time on that.

PR #394, opened 2017-09-07
Support hexadecimal float literals in lexer and parser

samrat:
See: https://github.com/diku-dk/futhark/issues/382

This does not include changes for Python and C readers.

### DIFF
--- a/docs/language-reference.rst
+++ b/docs/language-reference.rst
@@ -56,6 +56,10 @@ and decimal literals are of type ``f64``.  Hexadecimal literals are
 supported by prefixing with ``0x``, and binary literals by prefixing
 with ``0b``.
 
+Floats can also be written in hexadecimal format such as ``0x1.fp3``,
+instead of the usual decimal notation. Here, ``0x1.f`` evaluates to
+``1 15/16`` and the ``p3`` multiplies it by ``2^3 = 8``.
+
 .. productionlist::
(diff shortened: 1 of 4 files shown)


Athas:
How come using hexadecimal literals in the test input works when you didn't implement it in the C reader?  Does it also work with the Python backend (which is not automatically run by Travis for sad reasons)?  Try `futhark-test --compiler=futhark-py`.

samrat:
Erm, turns out the test runner doesn't quite behave the way I thought it did. The test input goes through the Alex lexer(`Futhark.Test.Values` calls `scanTokens`). So, the values the compiled Futhark program sees are the "normal", decimal floats(converted from the hexadecimal literals in `fromHexRealLit`).

The test I added do test the changes to the lexer and parser-- it's just that the values are scanned by `futhark-test` and not the reader. If the inputs were scanned directly by the reader, the test I wrote would have failed. 

I could change my test so that `main` just returns a tuple with two arrays:

```
let main (): ([]f64, []f32) =
    ([0xf.fp2f64], [...])
```

Let me know what you think.

Athas:
Oh yeah, I forgot that `futhark-test` applies its own lexing.  However, I would prefer if the new syntax is actually present in the test program (but having it in the input is also good).

samrat:
Alright, updated the commit with the test.

PR #395, opened 2017-09-08
Support hexadecimal floats in Python and C reader.

samrat:
This finishes up the work for https://github.com/diku-dk/futhark/issues/382


Athas:
I'm surprised you were able to work with my faux-parser combinators in the Python reader.

PR #423, opened 2017-10-25
Count remaining test cases instead of using a set and print the correct test name.

ahovgaard:


src/futhark-test.hs
```diff
@@ -337,20 +336,18 @@ runTests config paths = do
...
-              Failure s -> do clear
-                              T.putStrLn (T.pack (testCaseProgram test) <> ":\n" <> s)
-                              next (failed+1) passed
-
-  (failed, passed) <- getResults (S.fromList included) 0 0
+
+      getResults 0         failed passed = clear >> return (failed, passed)
+      getResults remaining failed passed = do
+        (test, res) <- takeMVar resmvar
+        report (testCaseProgram test) failed passed remaining
```
Athas: I'm not sure this is correct now.  The `report` functions prints text indicating which test we are *waiting* for, but you are now printing the name of the test that was most recently *completed*.  The original motivation for this piece of information was to report tests that were hanging (because of infinite loops or whatnot).

ahovgaard: I see. It's my misunderstanding then. I thought it was a bug because it was stuck on one test most of the time when I ran the test and I thought it was supposed to print the test it completes. I think it still prints "Waiting for BabyBearFun.fut" (in my case) for a long time though, even if that test may not have started yet.

Athas: Yes, it's a little confusing.  It would be better if it printed the next test to be run, instead of some arbitrary test.  But I'm not sure that this information is easily available.  You could add another kind of message that is put in the MVar whenever a test is chosen for testing, and then use that to print the message.

ahovgaard: I think it'll print whatever test is first in the list regardless of whether it has been started by a thread or not.

Athas: Indeed, there is definitely room for improvement.  But I think it's important to maintain the property that if some test is hanging (or just significantly slower than everything else), it will eventually hang while having its name displayed here.  That's an important feature.

ahovgaard: You are right. Please just ignore this pull request then.

Issue #419, opened 2017-10-17
Cosmin says this program fails


Athas:
```
import "/futlib/math"
import "/futlib/array"

let sgmPrefSum [n] (flags: [n]i32) (data: [n]i32) : [n]i32 =
    #2 (unzip (scan (\(x_flag,x) (y_flag,y) ->
                     let flag = x_flag | y_flag in
                     if y_flag != 0
                     then (flag, y)
                     else (flag, x + y))
                    (0, 0)
                    (zip flags data)))

let bin_packing_ffh [q] 
            (w: i32)              -- the bin size
            (all_perm  : *[q]i32) -- the current permutation of elems [0..q-1]
            (all_data0 :  [q]i32) -- the data (weights/lengths) corresponding to elements in perm
          : ([]i32, [q]i32)       -- resulting shape + resulting permutation

=
    let all_data = scatter (replicate q 0) all_perm all_data0
    let len   = q
    let cur_shape = replicate 0 0
    let goOn  = true
    let count = 0
    let (_,all_perm,_,cur_shape, _,_) = 
    loop ((len,all_perm,all_data,cur_shape, goOn,count)) while goOn && count < 100 do
        let (data, _) = split (len) all_data
        let (perm, _) = split (len) all_perm

        -- 1. initial attempt by first fit heuristic
        let scan_data = scan (+) 0 data
        let ini_sgms  = map (/w) scan_data

        let num_sgms  = (last ini_sgms) + 1  -- OK
        let flags = map (\i -> if i == 0 then 1
                               else if unsafe ini_sgms[i-1] == ini_sgms[i]
                                    then 0 else 1
                        ) (iota len)
        let ones  = replicate len 1
        let tmp   = sgmPrefSum flags ones
        let (inds1,inds2,vals) = unzip (
            map (\ i -> if (i == len-1) || (unsafe flags[i+1] == 1) 
                             -- end of segment
                             then (i+1-tmp[i], ini_sgms[i], tmp[i])
                             else (-1,-1,0)
                ) (iota len)
          )
        let flags = scatter (replicate len 0) inds1 vals
        let shapes= scatter (replicate num_sgms 0) inds2 vals

        -- 2. try validate: whatever does not fit move it as a first segment
        let scan_data = sgmPrefSum flags data
        let ini_sgms  = scan (+) 0 (map (\x -> if x > 0 then 1 else 0) flags) 
                        -- map (/w) scan_data
        let moves =
            map (\ i ->
[...]

PR #428, opened 2017-11-05
Make context-returns explicit for functions and branches.

Athas:
This does not fix anything by itself, but opens up a fix for #419.  It
is also a conceptual simplification.  I suspect a little more surgery
could be performed to simplify further.


Athas:
This definitely breaks memory block merging, so @nqpz will have to repair it.

nqpz:
I will do a general cleanup of the memory block merging code after my defense.  Thank you for the heads up wrt. this part.

PR #476, opened 2018-02-14
Fix set_bit.

ahovgaard:


ahovgaard:
The test of `i64.set_bit` actually fails where I expect that it should succeed. I'm not sure why exactly, but it looks like `i64.set_bit 31 (2^32-1)` returns `-1i64`.

Athas:
Something about sign extension?

Athas:
Might be a bug in the interpreter. Does it work if you pass -c to futhark-test?

ahovgaard:
It gives the same result when compiled and interpreted.

futlib/math.fut
```diff
@@ -394,7 +394,7 @@ module i64: (size with t = i64) = {
   let num_bits = 64
   let get_bit (bit: i32) (x: t) = to_i32 ((x >> i32 bit) & i32 1)
   let set_bit (bit: i32) (x: t) (b: i32) =
-    ((x & i32 (intrinsics.~(1 intrinsics.<< b))) | x intrinsics.<< i32 b)
+    ((x & i32 (intrinsics.~(1 intrinsics.<< bit))) | i32 (b intrinsics.<< bit))
```
Athas: Try making the RHS a conversion from `u32` instead.

Athas: The shadowing inside the module makes it really obnoxious to write.  You want something like:

    ((x & i32 (intrinsics.~(1 intrinsics.<< bit))) | intrinsics.zext_i32_i64 (b intrinsics.<< bit))

Athas:
I assume it works now, but could you squash the commits?  While unlikely in practice, having defect commits in there could cause bisection problems in the future.

Issue #461, opened 2018-01-17
Integrate memory block merging


nqpz:
## Recap

+ In-place lowering: The existing optimisation that we would like to be more general than.
+ Memory block coalescing: The optimisation that looks at `copy`, `concat`, and in-place update statements and tries to merge the left-hand side memory block with the right-hand side memory block.
+ Memory block reuse: The optimisation that tries to let statements reuse memory blocks whose last uses have occured.

## Short status

I have updated the code to work with the current compiler infrastructure and AST; see the https://github.com/diku-dk/futhark/tree/niels-memory-block-merging-final branch.  It passes all our 90+ memory block-specific tests, but it does not handle all our general test programs and benchmarks yet (which means that there are some more things we need to test).  The memory footprint reduction is always non-negative and ranges from 0% to 70%.

I think I can fix the problems mentioned below, but my original timeframe for doing so slipped, so I'm not sure.

## Current compilation and transformation problems

+ [x] Type errors after memory block reuse.  This happens in `tests/futlib_tests/random.fut` and a couple of benchmarks and is related to how the memory block merging modifies a program.  Something goes wrong, some problems seemingly related to IfAttr and constant-size memory blocks, and some seemingly related to the "max trick" employed by memory block reuse.  All of the program transformation happens in the MemoryBlockMerging.MemoryUpdater module, so this is the only place that needs fixing.
+ [x] In the GPU pipeline, the modification of InterestCalib manages to create a program with a memory block size that is not lambda-invariant.  We should probably just restrict memory block merging inside lambdas.
+ [x] In the CPU pipeline, GHC runs out of memory on compilation of InterestCalib.  This is likely due to some silly bug.
+ [x] heston32 compiles but gives a segmentation fault when it is run.  heston64 does not, so this is probably related to how the code calculates the byte size of a memory block.
+ [x] nn-opt-1 and bfs_filt_padded_fused give the wrong results when memory reuse is enabled.  Maybe the interference analysis is lacking in some edge cases.

## Runtime regressions

+ [x] LocVolCalib and nn-opt-4 (at least) run a bit slower when memory reuse is *enabled*.  Even the most basic setting of "only reuse memory if the blocks do not interfere and the block sizes are exactly the same" results in a slower program.  I'm at a bit of a loss here.
[...]

PR #477, opened 2018-02-18
Make memory block merging work with the current compiler

nqpz:
Also see #461.

Extend futhark-bench.hs to also save stderr in a JSON file.  This is useful when
used with '--pass-option=-debugging'.

Works with all benchmarks and test programs (that also work without it).  There
are still a few cases where the resulting code is poorer in terms of speed or
peak memory usage.  The optimisations are still disabled by default in the the
Futhark.Passes module.

Remove all memory block merging debugging code.

Move all memory block tests to the general tests directory.

Rename the *Hoisting modules.

Be a bit more conservative in the last use analysis.

Disable memory reuse on constant-sized allocations.  Makes the heston32
benchmark not segfault, but I'm not entirely sure why.

Create a function offsetIndexDWIM that handles edge cases.

Fix misc. edge cases in MemoryUpdater.

Fix the PrimExp Eq instance wrt. to literals.


Athas:
Could you factor out your changes to the test suite in a separate patch, so we can delay committing those until we enable this stuff by default?

nqpz:
Right, done.

Issue #323, opened 2017-03-23
Add higher-order functions


Athas:
Add function types to the grammar:

```
type        ::= ... | function_type
function_type  ::=  type "->" type
```

Add lambda forms as expressions:

```
exp ::= ... | "\" params "->" exp 
```

We'll also have to modify the grammar in a few more places to permit SOAC arguments to be expressions and the like.

The trick will be in the type rules, to permit static interpretation of function values:

1. A branch may not return a function.
2. A merge parameter in a `do`-loop may not be a function.
3. An array cannot contain functions.
4. An abstract type may not be a function (or contain one, if a tuple or record).

Athas:
Implemented in #486.  Many bugs and omissions likely still remain, which will have to be fixed, but the foundation is strong.

PR #486, opened 2018-03-06
Add support for higher-order functions.

ahovgaard:
See issue #323.

### DIFF
--- a/src/Futhark/Doc/Generator.hs
+++ b/src/Futhark/Doc/Generator.hs
@@ -178,6 +178,7 @@ prettyType t = case t of
             toHtml (nameToString name) <> ":" <> prettyType tp
   TypeVar et targs ->
     prettyTypeName et <> foldMap ((<> " ") . prettyTypeArg) targs
+  LiftedTypeVar et -> prettyTypeName et
   Array et shape u -> prettyU u <> prettyShapeDecl shape <> prettyElem et
   Arrow _ (Just v) t1 t2 ->
     parens (vnameHtml v <> ": " <> prettyType t1) <> " -> " <> prettyType t2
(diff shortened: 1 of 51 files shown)


Athas:
Cool!  I'll take a closer look tomorrow.

Athas:
Travis reports a problem with [issue481.fut](https://github.com/diku-dk/futhark/blob/master/tests/issue481.fut).  Looks like a type substitution is missing somewhere.  A similar error occurs for [lud.fut](https://github.com/diku-dk/futhark-benchmarks/blob/master/rodinia/lud/lud.fut), so I think it's the same issue.

A more interesting failure is with [kmeans.fut](https://github.com/diku-dk/futhark-benchmarks/blob/master/accelerate/kmeans/kmeans.fut), which fails with:

```
Internal compiler error.
Please report this at https://github.com/diku-dk/futhark/issues.
After internalisation:
In function recentre_2601
A unique tuple element of return value of function recentre_2601 is aliased to some other tuple component.
```

The problem seems to be that the type of the local function `recentre` is incorrectly inferred to be unique (or maybe only the lifted version is?), although it's not yet clear to me where, how, or why.

Athas:
The return type of `recentre` is changed from Nonunique to Unique after the defunctionaliser.  Specifically, the return looks like this after the monomorphiser:

```
Array (ArrayRecordElem
(fromList [(Name "1",RecordArrayElem (ArrayPrimElem (FloatType
Float32) ())),(Name "2",RecordArrayElem (ArrayPrimElem (FloatType
Float32) ()))])) (ShapeDecl {shapeDims = [NamedDim (QualName
{qualQuals = [], qualLeaf = VName (Name "k") 2462})]})
Nonunique
```

And this after defunctionalisation:

```
Array (ArrayRecordElem
(fromList [(Name "1",RecordArrayElem (ArrayPrimElem (FloatType
Float32) ())),(Name "2",RecordArrayElem (ArrayPrimElem (FloatType
Float32) ()))])) (ShapeDecl {shapeDims = [AnyDim]}) Unique
```

ahovgaard:
The k-means benchmark works now. I just needed to combine the shape and uniqueness information from the original return type in a local function with the type of the defunctionalized body, just as for top-level value bindings. By the way, I don't think the return type on `recentre` actually has the named size `k`.

Athas:
Right, it doesn't.  (But that's fine.). I have run my benchmarks, and there is no performance regression.  As soon as you squash your commits, I'll merge this.

Athas:
Futhark is now a higher-order programming language!

PR #505, opened 2018-03-15
Defunctionalization of higher-order infix operators.

### LINKED ISSUE #495: Support higher-order infix operators
The defunctionaliser does not yet handle infix operators.  I think I added the necessary annotations to the infix operator and operator section AST nodes, and the type checker also deals with it, so it should mostly be a matter of wiring it up to the existing machinery for function applications.

ahovgaard:
Fixes #495.


Athas:
Great work!  Now I can finally define the `>>=` operator.  (I see you already defined some Arrow combinators in your tests.)

PR #519, opened 2018-04-04
Turn 'map' into a function.

### LINKED ISSUE #518: Turn map into a function
`map` is currently a language construct, not a function.  This means, among other things, that you can't do `map (map f) xss`.  There is a `map1` function in the basis library that can be used instead, but it feels ugly.  The main reason `map` remains a language construct is that it can accept multiple input arrays, as in `map f x y z`.  However, the compiler will currently warn if you do so.  There are also some quirks around aliasing, but I'd rather get rid of those anyway - none of our benchmarks make use of it.

There is no doubt that `map` will *eventually* become a normal function, as part of my general crusade against language constructs.  Deus vult.  The only question is whether it should happen *before* the next release (in a week), or after.  Opinions?

Athas:
Closes #518.

This bans some older uses of 'map'; specifically those where we
consumed the inputs.  I am considering adding an array comprehension
construct to bring back that capability.


Athas:
Note that some programs, like Heston calibration, now compile extremely slowly because they produce a huge number of tiny lifted functions.

PR #533, opened 2018-04-15
Use higher-order functions instead of modules for segmented_scan.

ahovgaard:
The `accelerate/pagerank` benchmark needs to be modified as well.


Athas:
I will take care of updating Pagerank.  It will be done in a somewhat hacky way, because I made the mistake of making the implementation of segmented reduction an assignment for my Chalmers lecture, and I'd rather not give away the full solution!

PR #551, opened 2018-05-16
Omit empty closure environments from defunctionalized programs.

ahovgaard:
This optimization to the defunctionalizer will omit (nested) empty closure environments from the records constructed for lambda-abstractions and it will omit either of `e1'` and `e2'` in an application if they only represent empty closure environments. It also avoids lifting functions that only capture empty environments.

This makes the defunctionalized programs significantly more readable and it could possibly improve compilation times. Do we have any good benchmark programs to evaluate compilation times?

### DIFF
--- a/src/Futhark/Internalise/Defunctionalise.hs
+++ b/src/Futhark/Internalise/Defunctionalise.hs
@@ -233,20 +233,22 @@ defuncExp e@(Lambda tparams pats e0 decl tp loc) = do
 
   -- Construct a record literal that closes over the environment of
   -- the lambda.  Closed-over 'DynamicFun's are converted to their
-  -- closure representation.
+  -- closure representation.  Empty closure environments are omitted.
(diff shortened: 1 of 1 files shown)


Athas:
Actually, leave off `--defunctionalise`.  The interesting thing is how long the generated code takes to pass through the internaliser (which is the default if no other options are passed to `futhark`).

ahovgaard:
By the way, I think the inlining in `defuncApply` may be less effective now because of the added checks that neither of `sv`, `sv1`, and `sv2` are "empty". For example, we could still inline `e0'` even when `sv1` and `sv2` are empty.

Athas:
I had to revert this commit because it would remove functions that we would still need to call.  Specifically, `apply_e` (which references `fname'`) is always present in the residual, so it is never correct to not generate the `fname'` function.

PR #554, opened 2018-05-25
Alternative sobol reducer using higher-order functions

melsman:
Maybe we should rename the function to something else than `redomap`?

### DIFF
--- a/futlib/sobol.fut
+++ b/futlib/sobol.fut
@@ -23,10 +23,16 @@ module type sobol = {
   -- | `chunk i n` returns the array `[v(i),...,v(i+n-1)]` of sobol
   -- vectors where `v(j)` is the `j`'th D-dimensional sobol vector
   val chunk : i32 -> (n:i32) -> [n][D]f64
+  -- | `simulate f g ne n` runs the function `f` on `n`
+  -- `D`-dimensional point-vice normalised sobol vectors. The results
+  -- are reduced using the monoid defined by the (assumed) associate
+  -- function `g` with its neutral element `ne`.
+  val simulate 't : ([D]f64 -> t) -> (t -> t -> t) -> t -> i32 -> t
+  -- | The `Reduce` module implements a modularised version of
+  -- `simulate`.
   module Reduce :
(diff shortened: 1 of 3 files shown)


melsman:
Sobol-sequences are quite nice:

image

Athas:
But `pi` converges immediately!  Why not just use that?

melsman:
Good question... Maybe sometimes you don't know what pi is! ;)

fre. 25. maj 2018 kl. 16.39 skrev Troels Henriksen <notifications@github.com

futlib/sobol.fut
```diff
@@ -26,6 +26,7 @@ module type sobol = {
   module Reduce :
       (X : { include monoid
              val f : [D]f64 -> t }) -> { val run : i32 -> X.t }
+  val redomap 't : ([D]f64 -> t) -> (t -> t -> t) -> t -> i32 -> t
```
Athas: Why is the neutral element required?  Could we not simply require that if the chunk function is applied to an empty array of `f64`s, then the result must be the neutral element?

Athas:
Why do we use `f64` in the interface?  As far as I can see, we construct them from 32-bit integers anyway.

melsman:
... it's not a chunk... it's an element in the D-dimensional space...

fre. 25. maj 2018 kl. 16.42 skrev Troels Henriksen <notifications@github.com

melsman:
Good point. I guess there are two alternatives:

1. Use u32 as the type. In this case the use will need to do some manual
conversions; I don't think I like that...

2. Somehow parameterise the module over the type...

fre. 25. maj 2018 kl. 16.42 skrev Troels Henriksen <notifications@github.com

Athas:
Instead of `redomap`, why not `run` or `simulate`?

Also, the test looks like it's failing.

melsman:
don't merge yet...

melsman:
yes - if tiu think it's fine ;)

søn. 27. maj 2018 kl. 12.22 skrev Troels Henriksen <notifications@github.com

PR #568, opened 2018-06-15
Implemented a C# backend for Futhark

MikkelStorgaard:


src/Futhark/CodeGen/Backends/CSOpenCL/Boilerplate.hs
```diff
@@ -0,0 +1,302 @@
...
+                           Exp $ CS.simpleCall "CL10.GetDeviceInfo"
+                             [ Var "ctx.opencl.device", Var $ clooString s
+                             , CS.simpleCall "new IntPtr" [CS.simpleCall "Marshal.SizeOf" [which']]
+                             , CS.toIntPtr $ Var "ptr", Out ctxNULL ]
+                         ]
+                       ]
+
+        clooString s = case s of
+                         "MAX_COMPUTE_UNITS" -> "ComputeDeviceInfo.MaxComputeUnits"
+                         _ -> undefined
```
Athas: dude

MikkelStorgaard: I don't recall this part. Will fix

Athas:
Apart from smaller code issues, there are two main things we'll need before merging this:

  * Reading/writing of the [binary data format](http://futhark.readthedocs.io/en/latest/binary-data-format.html).  It's a trivial format, so I don't foresee many challenges here.

  * A methodology for testing/benchmarking the C# code generator on the [buildbot](http://buildbot.futhark-lang.org/#/).  This involves figuring out what packages to install, and what environment variables to set up, to make everything work nicely with `futhark-test` and `futhark-bench`.  This one may be a little trickier, because both these tools currently assume that the generated executables can be invoked directly, which I'm not sure Mono is OK with that.

Issue #416, opened 2017-10-13
Support compressed (gzipped) data files for futhark-test and futhark-bench


Athas:
Even using the binary data format, some data files for our benchmarks are very large.  However, they are also often very *sparse*.  For example, `kdd_cup.in` for the Rodinia K-means benchmark is 66MiB in the binary format, but only 2.1MiB after being compressed with `gzip`.

The test tooling provided by `futhark-test` and `futhark-bench` should support compressed input- and output-files, and automatically decompress them before passing their contents to the test program.  This will not be an invasive change, as the tooling is already passing the input on the stdin stream - decompressing along the way would be straightforward to add.

The motivation for this change is that Github is not really supposed to be used for giant data files, but many of the benchmarks we port do indeed have such giant files.  In the future, we might even be able to unify the `futhark-benchmarks` and `futhark` repositories.  Currently they are split, and mostly because `futhark-benchmarks` contains a bunch of inconveniently large data files.

There is [zlib package](https://hackage.haskell.org/package/zlib) for Haskell which looks quite easy to use.

Athas:
Closed in 2a38574af2c50f675807199ecfa8d840b1d9ead1.

PR #574, opened 2018-07-01
Support for gzipped files for futhark-bench/futhark-test  

zfnmxt:
A first attempt at implementing #416 . I wasn't sure if I should add any tests for the issue (esp. since futhark-benchmarks is a separate repo) so I've left it as-is. 

The change simply decompresses any input files that end in ".gz".

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -49,6 +49,7 @@ dependencies:
   - extra
   - process
   - markdown
+  - zlib
 
 when:
 - condition: "!impl(ghc >= 8.0)"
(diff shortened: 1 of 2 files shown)


Athas:
I think by the way at least `futhark-benchmark` works, it will always be obvious which data file is suspect.

zfnmxt:
This happens:
```
Codec.Compression.Zlib: compressed data stream format error (incorrect header check)
```
I'll look into adding the filename.

zfnmxt:
Right, with `futhark-benchmark` it's essentially not an issue.

Issue #421, opened 2017-10-22
futhark-test should report the actual number of tests run


lolkat2k:
When running `futhark-test` on the following program:
```
-- Test the add function.
-- ==
-- entry: add
-- input { 1 } output { 2 }
-- input { 2 } output { 3 }

entry add (x : i32) : i32 = x + 1
```
it outputs `(0 failed, 1 passed)`.

This is due to `futhark-test` counting the number of files and not the number of tests run. If we take an input-output pair as being one test, then the output should be `(0 failed, 2 passed)`.

lolkat2k:
Note that, if `futhark-test` is given a directory with x `.fut`-files, `futhark-test` will output `(0 failed, x)` passed.

Athas:
This would be relatively straightforward to fix, I think.  `futhark-test` uses files as the basic unit of labour (which should not change because that's also the unit of compilation), but each file could easily count as several tests.  Or maybe a separate statistic could count the number of test cases run.  All files have their test block parsed before anything starts executing, so we have access to all the information we need.

zfnmxt:
Fixed by 8d88228.

Issue #209, opened 2016-06-30
Add some colour to futhark-test error output


athas:
It would be easier to read the output with some pretty formatting, spacing and the like.  Easy to do.

Athas:
The most important thing is to highlight errors somehow, so they are easier to find when scrolling.  I'd rather not make it a total colour party, because that makes *everything* hard to find.

PR #580, opened 2018-07-07
Adds case reporting to futhark-test. (#421)

zfnmxt:
At the moment, the case reporting is super rudimentary and only appears at the end of output. I also started working on #209 in parallel so I figured it'd be best to not embellish too much here and leave improving the output for that issue.

I also refactored `runTests` a good bit by adding a `TestStatus` record. @Athas, if you're not a fan of these changes, I can also simply tack on a "cases run" argument to `getResults`.

Finally, I thought it would also potentially be nice to report the number of cases failed/passed, but this would be more involved to add. It's also (somewhat) at odds with how `futhark-test` currently reports errors; namely, it only reports a single error per file (even if there are multiple).

### DIFF
--- a/src/futhark-test.hs
+++ b/src/futhark-test.hs
@@ -260,11 +260,11 @@ compareResult _ (RunTimeFailure f) (SuccessResult _) =
 --- Test manager
 ---
 
-data TestStatus = TestStatus { _testStatusRemain :: [TestCase]
-                             , _testStatusRun :: [TestCase]
-                             , _testStatusFail :: Int
-                             , _testStatusPass :: Int
-                             , _testStatusCases :: Int
+data TestStatus = TestStatus { testStatusRemain :: [TestCase]
+                             , testStatusRun :: [TestCase]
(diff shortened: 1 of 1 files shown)


src/futhark-test.hs
```diff
@@ -260,6 +260,13 @@ compareResult _ (RunTimeFailure f) (SuccessResult _) =
 --- Test manager
 ---
 
+data TestStatus = TestStatus { _testStatusRemain :: [TestCase]
+                             , _testStatusRun :: [TestCase]
+                             , _testStatusFail :: Int
+                             , _testStatusPass :: Int
+                             , _testStatusCases :: Int
+                             }
+
```
Athas: Don't underscore-prefix field names that are actually used.  It's just a hack to shut up GHC from warning about unused names, in those cases where the field names are only used to improve the readability of the record definition.

Athas:
I'm OK with the refactoring into a record (it was getting to be a bit messy), but does this patch have any functionality changes?

Issue #630, opened 2018-09-08
Found futhark-test output confusing


einar-io:
At least two students found

    futhark-test lssp.fut
    0 failed, 1 passed, 3 cases run.

confusing, because the numbers do not add up.

Suggestion: Add clarifying units

    futhark-test lssp.fut
    0 test file(s) failed, 1 test file(s) passed, 3 test cases run.

athas:
@zfnmxt I think you were the last one to go near this.  Fancy having another go?

zfnmxt:
Sure. I've actually already fixed it; just never got around to polishing it.

zfnmxt:
Fixed by 8d88228.

PR #634, opened 2018-09-16
Adds error accumulation to futhark-test, addresses issues #630/#421, improves futhark-test output.

zfnmxt:
Changes:

- Adds error accumulation. `futhark-test` will now report all errors for a program, with some limitations (which can be easily changed as desired): Structure tests, warning tests, compile tests, compiled tests, and interpreted tests do not mutually accumulate. For example, if there is an error when running a compiled test, futhark-test will not run any interpreted tests for that program. This can be modified by placing an `accErrors_` call in the relevant place but I thought this was the most reasonable implementation.

- Adds (basic) table building and uses it to format the output of `futhark-test` more clearly.

- Adds a (tiny) bit of color to the output of `futhark-test` :) 

- Because `futhark-test` now allows for multiple failures when testing a program, `<test>.fut.actual` and `<test>.fut.expected` test output files were changed to `<test>.fut.<entry>.<index>.actual` and `<test>.fut.<entry>.<index>.expected`, respectively.

- Changes "files" to "programs" and calls the individual tested components of programs a "run" (e.g., a single input-output case or warning test). 

- Now reports the total number of runs run, the failed number of runs, and the passed number of runs.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -59,6 +59,7 @@ dependencies:
   - free >= 4.12.4
   - zip-archive >= 0.3.1.1
   - time >= 1.6.0.1
+  - ansi-terminal >= 0.6.3.1
 
 when:
 - condition: "!impl(ghc >= 8.0)"
--- a/src/Futhark/Test.hs
+++ b/src/Futhark/Test.hs
@@ -110,6 +110,7 @@ data TestRun = TestRun
                { runTags :: [String]
                , runInput :: Values
                , runExpectedResult :: ExpectedResult Values
+               , runIndex :: Int
                , runDescription :: String
                }
              deriving (Show)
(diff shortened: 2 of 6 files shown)


athas:
Are all the intermediate commits valid?  If not, could you squash them?  Having broken commits in the history makes bisecting difficult.

Also, could you perhaps use a slightly older version of ansi-terminal so we can keep working with older Stack snapshots?  (This is not an unimportant feature; currently older snapshots are needed for correct functioning on Windows due to GHC bugs.)

athas:
Just one style issue to go!

athas:
Yes, follow the suggestion.  It's more important to not have to fiddle with tools than to agree with every suggestion they make.

PR #660, opened 2018-10-25
Fixed tabulate_3d.

tpolakovic:
Fix to the typo in tabulate_3d, which caused the function to fail. Array now has correct dimensions.


athas:
Thanks, looks good!

PR #662, opened 2018-10-31
Adds enums to Futhark.

zfnmxt:
Here's the (rather late--sorry, my dad was visiting on short notice) promised PR for enums.

Some things are definitely a bit clunky right now (`Parser.y` with its boilerplate for matches and the pattern match warning code in `Terms.hs` come to mind), but I expect most of this will be addressed with the increasing complexity of the next stage of the project.


athas:
A glorious commit!  I'll do a quick review, but I expect that we'll polish most things post-merge.

src/Futhark/Internalise.hs
```diff
@@ -880,6 +904,57 @@ internaliseExp _ e@E.ProjectSection{} =
 internaliseExp _ e@E.IndexSection{} =
   fail $ "internaliseExp: Unexpected index section at " ++ locStr (srclocOf e)
 
+boolOp :: VName -> E.Exp -> E.Exp -> E.Exp
+boolOp op l r = E.BinOp (qualName op) (Info $ vacuousShapeAnnotations ft)
+                 (l, sType l) (r, sType r) (Info (E.Prim E.Bool)) noLoc
+  where sType e = Info $ toStruct $ vacuousShapeAnnotations $ E.typeOf e
+        arrow   = Arrow S.empty Nothing
+        ft      = E.typeOf l `arrow` E.typeOf r `arrow` E.Prim E.Bool
```
athas: This entire thing is pretty odd.  You construct a pretty baroque mechanism (also `vnM` below) to construct `&&` and `==` expressions.  Why not just generate `If` expressions for the former?  They will be internalised the same way.  I'd suggest just defining two functions `andExp, eqExp :: E.Exp -> E.Exp -> E.Exp`.

zfnmxt: `andExp` was easy enough to do, but with `eqExp` I'm still not really sure how to handle the `VName` stuff (in regards to `vnM`). The tag doesn't actually matter here (as long as it's less than `maxIntrinsicTag`), so I could just put `(-1)` and call it a day, but that seems a bit jank to me. What do you suggest?

athas: The nicest way would be to generate core code, but I guess that's a bit awkward.  Using `-1` will suffice for now.

src/Language/Futhark/Attributes.hs
```diff
@@ -1035,3 +1081,15 @@ type UncheckedDec = DecBase NoInfo Name
 
 -- | A Futhark program with no type annotations.
 type UncheckedProg = ProgBase NoInfo Name
+
+-- | A case (of a match expression) with no type annotations.
+type UncheckedCase = CaseBase NoInfo Name
+
+-- | An unmatched pattern. Used in in the generation of
+-- unmatched pattern warnings by the type checker.
+data Unmatched p = UnmatchedNum p [ExpBase Info VName]
```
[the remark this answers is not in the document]
zfnmxt: Not sure if there's really a better place. `Terms.hs` would be nice, but results in a circular import because `Unmatched` has a `Pretty` instance.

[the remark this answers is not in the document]
athas: Just move the Pretty instance as well.

athas:
Could you re-squash your commits?

zfnmxt:
All squashed.

zfnmxt:
Sorry, got lazy with hlint since spacemac's haskell layer does it (but only if you pay attention!). 

Issues should be fixed now, let's see what Travis says.

athas:
Hlint remains upset.

zfnmxt:
Messed up merging the last commit from `enums` into `enums_merge`. Not sure what happened, but it's there now.

Also, to be sure, I installed hlint from the lts-12.11 snapshot as Travis seems to use. The lts-12.11 hlint still generates some warnings (for code outside of the PR) that Travis doesn't seem to generate. Why is that?

athas:
The Travis job runs `tools/style-check.sh`, which disables a few `hlint` rules.

PR #684, opened 2018-12-17
Use ImpCode formulation of transpose kernels.

athas:
Previously we used hand-written OpenCL kernels.  This caused minor
problems for the coming CUDA backend and major issues for the coming
Vulkan backend.  With the kernels written in the ImpCode.Kernels
representation, these backends can now simply use their general code
generators.  Transposes are no longer magical (after ImpGen.Kernels).

We still have the same four versions that we used to.  In the short
term, the transpose block size is now hardcoded, but that will be
fixed eventually.

Performance is supposed to be unaffected, but transposing byte arrays
is inexplicably a bit slower.


athas:
@jstokh and @Zyntax9, you may want to be aware of this.

PR #704, opened 2019-01-25
CUDA backend

jstokh:


athas:
[Travis complains](https://travis-ci.org/diku-dk/futhark/jobs/484455405).  Also, the module Futhark.CodeGen.ImpGen.Cuda should be renamed. @jstokh

athas:
Still a style issue left.

athas:
The tests are all green, and I've run the benchmarks.  Results are generally on par with OpenCL, sometimes faster and sometimes slower.  Some benchmarks, like nbody and mandelbrot are still slow for some reason.

There are also some benchmarks that fail.  Primarily [transpose.fut](https://github.com/diku-dk/futhark-benchmarks/blob/master/micro/transpose.fut), because we exceed the CUDA grid limitations.  @jstokh had originally developed a virtualisation technique to deal with it, but I thought that was unnecessary with chunking in place.  Guess not.  I suggest it is re-introduced in a general manner for all multidimensional kernels.

The more troubling failure is [bfast](https://github.com/diku-dk/futhark-benchmarks/blob/master/misc/bfast/bfast-cloudy.fut), which simply produces a completely garbage result.  Something is very wrong here.

I think I will still merge this, but I won't be hooking up the automatic benchmarking until the above has been resolved.

Issue #718, opened 2019-02-28
Add header guards to generated .h files


athas:
When compiling a Futhark program with `--library`, the generated `.h` files do not have header guards.  We should fix this.

athas:
Closed in 7877e8d70037db3af0a7ca49ca756ee8a160a401.

PR #744, opened 2019-04-29
Header guards

mxxo:
An attempt to add header guards to generated C headers, resolving #718. @athas, may I ask your advice on where to draw the filename from? I thought I could use one of the generated function filepaths but this ended up evaluating to GenericC.hs instead of the C file like I intended. 

Thank you very much for your time. 
Max

### DIFF
--- a/src/Futhark/CodeGen/Backends/GenericC.hs
+++ b/src/Futhark/CodeGen/Backends/GenericC.hs
@@ -426,7 +426,7 @@ decl x = item [C.citem|$decl:x;|]
 addrOf :: C.Exp -> C.Exp
 addrOf e = [C.cexp|&$exp:e|]
 
--- | Public names must have a consitent prefix.
+-- | Public names must have a consistent prefix.
 publicName :: String -> CompilerM op s String
 publicName s = return $ "futhark_" ++ s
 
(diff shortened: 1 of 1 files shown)


athas:
As a first approximation, I think hard-coded header guard names are fine.  After that (or alternatively), we'll need to extend `GenericC.compileProg` to take as argument a `String` that indicates the source file name.  This is not particularly difficult, but it requires plumbing it all the way through from the [c](https://github.com/diku-dk/futhark/blob/master/src/Futhark/CLI/C.hs) and [opencl](https://github.com/diku-dk/futhark/blob/master/src/Futhark/CLI/OpenCL.hs) subcommands.

Also, be careful with your commits.  This pull request also has some commits related to fixing typos.

mxxo:
Right, thank you very much. I'll separate out the unrelated files from this branch.

mxxo:
Hi there, sorry about the force push! I'll make sure to keep things well separated in the future.

kfl:
You could also just use `#pragma once` as guard. It should be supported by all modern C compilers.

athas:
I'd be OK with `#pragma once`.

Your current patch is inconsistent in that the `#ifdef` is added explicitly to `headerdefs`, while the `#endif` is added with `headerDecl`.  The latter is only for dynamically constructing the contents, so I'd prefer the `#endif` to also be put in `headerdefs` (if you don't decide to go with `#pragma once`).

mxxo:
Thank you both for your help, this seems much cleaner and doesn't change the compileProg interface

PR #777, opened 2019-07-04
Move to a segmap-style kernel representation based on PPoPP paper.

### LINKED ISSUE #666: A new kernel representation
It is time to redo the kernel representation.  The current kernel
representation is based on a generic `Kernel` that contains
`KernelExp`s.  Specific operations like `scan` and `reduce` is
implemented by putting these constructs together with `If` and
`DoLoop`.  The idea was that a single unified representation would
simplify things.  To a degree, it did.  However, some things are
really hard to express this way in Futhark's purely functional IR, so
we end up with nasty hacks like `InPlaceReturn`.

Cosmin and I have discussed moving the kernel representation closer to
the notation used in the PPoPP paper (with `segmap`/`segred`/`segscan`).
The purpose is twofold:

  * Support specialised implementations of e.g. scan kernels.

  * Create a cleaner separation between thread-specific and
    workgroup-wide statements, which will enable us to do more
    sophisticated intra-group parallelism.

This will be heroic project.  Pretty much *everything* downstream of
kernel extraction and until imperative code generation will have to be
modified.  Realistically, this means I am the only one who can do it,
and I will likely have to remove e.g. memory block merging, because it
will be too onerous to update.

The new representation will be based on having multiple kernel
constructs: certainly `segmap`/`segred`/`segscan`, and probably also
`seggenred`, and possibly also `segscatter` (although I suspect it can
be rolled into `segmap`).  The following important properties must
hold:

  * Kernel babysitting (fixing coalescing) must be doable.  This is
    probably easy, since our current approach does not actually mutate
    the kernels themselves.  The only requirement is that we have a
    notion of which iterations of a `segmap` are spatially adjacent
    (but this follows in a natural way).

  * We must be able to generate straightforward
    sequential-code-per-thread in the kernel extractor, which can then
    be *tiled* to improve locality.  This need only work for `segmap`,
    as the current tiler also only works for kernels that were
    originally `map` nests.

  * Memory expansion must still be possible.

  * ImpCode must remain unchanged.  This will allow us to retain all
    our code generators (including the new CUDA and Vulkan ones)
    without modification.

Cosmin and I envision that we'll keep the `segmap` etc. constructs
late in the compiler, and only lower them during imperative code
generation.  This means that things like the versioned implementation
of `segred` will be added implicitly *including its memory
allocations*.

It'll be fun!  As a first proof of concept, I think I will add a
`SegRedKernel` construct.  This will not require me to *modify*
anything, so it should be simple.  If that goes well, I'll add a
`SegScanKernel`, which can hopefully quickly be improved to use the
efficient scan implementations developed by the PMPH students.  Once
those are done, I'll start considering how to handle the more general
case and support better intra-group parallelism.

The current kernel representation has lasted us for three years, and
was grown in an ad-hoc manner, and based on constraints
(e.g. extremely barebones runtime support) that are no longer true
(e.g. now we have an excellent allocator).  Still, it was enough to
make Futhark probably the fastest functional language available.  I
hope that with our experience, and some care, the new representation
will last longer, and permit even better performance.

athas:
One of the primary intents is to improve intra-group parallelism, and
in particular implement flattening in there as well.  This work has
been begun, but not yet finished.

Closes #666.


athas:
When this work was outlined in #666, I said it would be a heroic project.  It sure was!  This huge pull request is merely the last in a long chain of changes.  But now we're almost there!

athas:
At least this part is nice:

```
 74 files changed, 5901 insertions(+), 6827 deletions(-)
```

Unfortunately, I think the balance is because **this commit removes register tiling**, at least until @coancea finds time to re-implement it.

PR #779, opened 2019-07-06
Add sum types to Futhark

### LINKED ISSUE #530: Adding sum types to Futhark
It is time to consider a design for sum types in Futhark.  It will be
a while before we will have time to implement it, but the design
itself is fairly subtle, so we might as well get started on a design
document.  There are two parts to the problem: source language design
and efficient implementation.  The latter is the most difficult, but
the former has its challenges as well.  I encourage comments and
feedback - while I have put some thought into the following, I am by
no means certain that my ideas are the best possible solution.

## Language Design

I would like the sum types to be structurally typed, just like
everything else in the Futhark type system.  This means they cannot be
recursive, but that's fine.  For example, this would be the definition
of an option type for integers:

```
none | some i32
```

And the usual type abbreviation mechanism can be used to give it a
name (and make it polymorphic):

```
type option 'a = none | some a
```

This has some challenges for type inference.  For example, this
top-level binding would be ambiguous:

```
let x = some 0
```

What is the type of `x`?  It could be `option i32` (which is the same
as `none | some i32`), or it could be the unary type `some i32`, or it
could be any type that contains a constructor `some i32`.  A similar
issue already exists for records, and is solved with a type
annotation:

```
let x: option i32 = some 0
```

We could also add some form of row polymorphism, but don't think it's
worth it.

We also have to come up with a syntax for pattern matching.  The two
superficial complications are that Futhark does not have significant
whitespace, and that the `|` token is a valid operator.  Still, we
could just use `when` or something:

```
match x
  when none    -> 0
  when some x' -> x'
```

A bigger question is how we determine, in a pattern, whether a name is
in binding position or refers to a constructor.  I do not want to
repeat the mess we have in Standard ML.  OCaml and Haskell solves this
very elegantly with casing, but that's not an option in Futhark.
Perhaps we could require a special sigil (prefix colon?) when matching
a constructor:

```
match x
  when :none    -> 0
  when :some x' -> x'
```

Or maybe binding names must be prefixed with a question mark:

```
match x
  when none     -> 0
       some ?x' -> x'
```

We need to find a solution.  This design flaw in Standard ML must not
be allowed to propagate.

## Implementation

In contrast to most other of our language enhancements (modules,
higher-order functions, type inference), adding sum types will likely
require extensions to the core language to support unions.  Or rather,
we can make do without, if we are willing to accept (significant)
inefficiency.  For example, an array of type `[n](option i32)` could be
represented in the core language as two arrays:

```
flags: [n]i8
vals: [n]i32
```

The `flags` indicate for each particular value which constructor is
used, and `vals` contain the constructor payload.  We still maintain
an integer for those that are `none`, but its value is undefined.  But
what happens if we have an array of a more complicated type, say one
with four constructors: `[n](c1 i32 | c2 f32 | c3 i64 | c4 f64)`.
Then the core representation could be something like:

```
flags: [n]i8
c1_vals: [n]i32
c2_vals: [n]f32
c3_vals: [n]i64
c4_vals: [n]f64
```

Essentially, an array of sum types is represented as an array of every
possible constructor for each element.  This works, but is expensive
for types with many constructors.  Yet it is simple to implement (no
core language changes required), so maybe we should just do it this
way at first, and later move (transparently!) to a more efficient
representation.

One such more efficient representation would require adding *arrays of
unions* to the core language.  Then the cost of an array of sum types
would be proportional to the size of the array times the *largest*
possible constructor.  The example above would be:

```
flags: [n]i8
vals: [n](i32|f32|i64|f64)
```

The code generator would treat the `vals` array as an array of C-style
`union`s.  Multiple arrays are still required for non-unary
constructors, such as the array `[](c1 i32 | c2 (i32,i32))`:

```
flags: [n]i8
c1_vals: [n]i32
c2_vals_a: [n]i32
c2_vals_b: [n]i32
```

Although in this case the `c1_vals` and one of the `c2_vals_*` arrays
could presumably be merged.

It would also be nice if we could efficiently handle cases where
constructors have the same size in bytes, but different number of
elements, such as in the type `[](c1 i32 | c2 (i16,i16))`.
Unfortunately, it's not clear how this should interact with the
array-of-tuples transformation.

athas:
Closes #530.

Co-authored-by: zfnmxt <37446532+zfnmxt@users.noreply.github.com>


athas:
This will need a little more testing, but it is in surprisingly good shape.

I am considering merging it, but not documenting it just yet, to give us some time to play with it first before inflicting it on an unsuspecting public.

The biggest problem is that things like `[#none, #some [1,2,3]]` do not work.  This will unfortunately require a more elaborate solution (much better size inference, probably dependent types at the source level), so we'll document that as a bug for now.

PR #783, opened 2019-07-16
Merge master into mgpu-husks

athas:
@Zyntax9 I have not tested it very carefully, but it seems to work.


Zyntax9:
Running tests I get a number of "Cannot allocate memory in SOAC", a select few examples being:

- redomapNew.fut
- reg-tiling/reg3d-test2.fut
- segredomap/add-then-reduce.fut

There are also a handful of programs that cause access violations on the device when running with a single node (and I would expect more nodes to be no different), where a select few are:

- replicate1.fut
- scatter/write3.fut
- shapes/known-shape.fut

Additionally, stack complains about missing pattern for "(Inner (Husk _ _ _ _ _))" in unAllocOp of ExpandAllocations.hs.

athas:
Don't worry about ExpandAllocations.hs.  Do the other programs work on your branch?

Zyntax9:
Yes. Are you unable to reproduce it?

athas:
I have fixed some of it, but I have not yet actually run it on a CUDA device.

athas:
I will not be able to debug it myself, since I cannot compile your code on the old CUDA machines we have:

```
tests/replicate1.c:1574:3: error: unknown type name 'pthread_barrier_t'; did you mean
      'pthread_attr_t'?
  pthread_barrier_t node_sync_point;
  ^~~~~~~~~~~~~~~~~
  pthread_attr_t
```

Zyntax9:
Only memory access violations are left. I can take a look at it.
Do you have a suggestion for an alternative to the pthread barrier that will make it work on the machines at your disposal?

athas:
Barriers can be implemented manually with locks and condition variables, but let's not worry about that for now.

Zyntax9:
The latest commit does not fix the memory access problem, but I expect it would fail in the multi-node case. The search continues!

Zyntax9:
The problem can be narrowed down to the program
```
let main(xs: []i32): i32 =
  reduce (+) 0 xs
```
which runs into an illegal memory access inside the CUDA kernel with any number of nodes.
Note that the OpenCL backend works and the single-node execution of the generated CUDA code shouldn't be that different.

Zyntax9:
I tested it with the master branch and it is also a problem on that.

athas:
How are you running this?  I cannot reproduce on `master`.  Even `cuda-memcheck` does not find anything wrong.

Zyntax9:
I just compile and run tests, i.e.
```
> futhark cuda .\test.fut
> echo "[1,2,3]" | .\redomap_test.exe
C:\Users\Zyntax\Downloads\test.exe: .\redomap_test.c:3389: CUDA call
  cuMemcpyDtoH(&read_res_3643, mem_3604.mem + 0 * sizeof(int32_t), sizeof(int32_t))
failed with error code 700 (an illegal memory access was encountered)
```
Running CUDA memcheck with the same input I get:
```
> cuda-memcheck .\test.exe
========= CUDA-MEMCHECK
[1,2,3]
.\test.exe: .\test.c:3389: CUDA call
  cuMemcpyDtoH(&read_res_3643, mem_3604.mem + 0 * sizeof(int32_t), sizeof(int32_t))
failed with error code 719 (unspecified launch failure)
========= Invalid __shared__ write of size 4
=========     at 0x00000f80 in segred_nonseg_3598
=========     by thread (159,0,0) in block (0,0,0)
=========     Address 0x00000280 is out of bounds
=========     Device Frame:segred_nonseg_3598 (segred_nonseg_3598 : 0xf80)
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame:C:\WINDOWS\SYSTEM32\nvcuda.dll (cuMemcpyPeer_ptds + 0x1c435c) [0x1d3ba5]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x4e1d]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x58be]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x4912]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x13c7]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x14fb]
=========     Host Frame:C:\WINDOWS\System32\KERNEL32.DLL (BaseThreadInitThunk + 0x14) [0x17bd4]
=========     Host Frame:C:\WINDOWS\SYSTEM32\ntdll.dll (RtlUserThreadStart + 0x21) [0x6ce71]
...
========= Invalid __shared__ write of size 4
=========     at 0x00000f80 in segred_nonseg_3598
=========     by thread (128,0,0) in block (0,0,0)
=========     Address 0x00000204 is out of bounds
=========     Device Frame:segred_nonseg_3598 (segred_nonseg_3598 : 0xf80)
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame:C:\WINDOWS\SYSTEM32\nvcuda.dll (cuMemcpyPeer_ptds + 0x1c435c) [0x1d3ba5]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x4e1d]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x58be]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x4912]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x13c7]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x14fb]
=========     Host Frame:C:\WINDOWS\System32\KERNEL32.DLL (BaseThreadInitThunk + 0x14) [0x17bd4]
=========     Host Frame:C:\WINDOWS\SYSTEM32\ntdll.dll (RtlUserThreadStart + 0x21) [0x6ce71]
=========
========= Program hit CUDA_ERROR_LAUNCH_FAILED (error 719) due to "unspecified launch failure" on CUDA API call to cuMemcpyDtoH_v2.
=========     Saved host backtrace up to driver entry point at error
=========     Host Frame:C:\WINDOWS\SYSTEM32\nvcuda.dll (cuMemcpyPeer_ptds + 0x1cde23) [0x1dd66c]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x4e4c]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x58be]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x4912]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x13c7]
=========     Host Frame:C:\Users\Zyntax\Downloads\test.exe [0x14fb]
=========     Host Frame:C:\WINDOWS\System32\KERNEL32.DLL (BaseThreadInitThunk + 0x14) [0x17bd4]
=========     Host Frame:C:\WINDOWS\SYSTEM32\ntdll.dll (RtlUserThreadStart + 0x21) [0x6ce71]
=========
========= ERROR SUMMARY: 33 errors
```
Where the omitted errors appear similar, but with the thread id being ranging from 128 to 159.

Zyntax9:
I think I may have found the problem! It seems that the definition of int64_t in the CUDA prelude is not 64 bit. At line 298 of ToOpenCL.hs it is defined as
```
typedef long int64_t;
```
but I suggest it should be
```
typedef long long int64_t;
```
Since the size of long should be compiler specific, I am not sure whether or not this is just a problem for the GCC supplied with mingw-w64. However, the typedef of uint64_t seems to be unsigned long long, so it just seems to be a mistake.
Given that this is a problem on master as well I'll let you decide where this should be changed.

athas:
`long long` is not guaranteed to exist in OpenCL, but `long` is guaranteed to be 64 bit.  Apparently CUDA does not properly fix the size of integer types, even in kernels.  What a shitshow.  I'll see if I can figure out a fix (maybe `long long` exists and is 64 bit everywhere we care about).

athas:
Why would that have an effect on this program, though?  It does not use any device-side 64-bit integers.  It does on the host, but there we use `<stdint.h>`.

athas:
Ah, it is because we do use int64_t as a standin for an eight-byte type when allocating shared memory.

Zyntax9:
Yes! It also explains why tid 128 and above fails to access the shared memory, as memcheck suggests.

Zyntax9:
I'm running tests with the change I suggested in the CUDA prelude on husks-merge. It seems to work, except for .\soacs\stream1.fut that fails to compile due to double binding of x_5578 and a number of metric-mismatches. The latter is likely just because of the expected kernels now having an outer husk.

athas:
Alright, give it another go.

Zyntax9:
The problem with .\soacs\stream1.fut should be fixed now. Tests are running and thus far only seem to have the metric mismatches left.

Zyntax9:
The tests look fine. From what I can see the only new test errors are the metric mismatches of new test files, so I think we are ready to merge it into mgpu-husks.
Do I just press the "Merge pull request" button?

athas:
Yes, just press the button.

PR #784, opened 2019-07-19
Work-in-progress on statically checking size-dependent types for Futhark.

athas:
Not documented.  We don't even have a clear spec for this yet!

I need to talk to @melsman about what precisely the type rules should be, and @coancea about whether this is even a good idea, or if it makes Futhark too hard to use.


athas:
We will need to change some of our coding habits.  Expressions like `iota (length xs)` and `replicate (length xs) v` are bad, and should either use size names (in the case of `iota`), or something like `map (const v) xs` instead of the `replicate`.

athas:
Closed in favour of #823.

Issue #747, opened 2019-05-01
Change header guards to not use #pragma once


athas:
#744 added header guards using `#pragma once`, which was great.  A small snag is that when we construct command line executables, we manually insert the generated header in the .c file, `#pragma once` and all.  Some C compilers warn about `#pragma once` outside of headers.  There are plenty of other spurious warnings issued for the code we generate, so fixing this is not urgent, but it would be nice to do so.

A good fix would involve plumbing through the name of the original `.fut` file from the CLI frontends to the code generator, and use that to generate a normal `#ifdef`/`#define` header guard.  We might eventually also consider using this to generate the public names, rather than just prefixing them all with `futhark_`, but that requires more thought.

mxxo:
I think this bad boy is ready 😎

PR #799, opened 2019-10-09
Fix #pragma once compiler warnings 

mxxo:
Work towards proper header guards (#747). 

The header guard name is hardcoded right now, coming full circle from my `pragma once` patch :sweat_smile:. 

I started plumbing the `.fut` name from the backends outwards and felt like I was vandalizing all the function signatures for a small benefit (at one point having to tweak C# backend code for a C-related change). This is probably mostly an inexperience thing.

The change seems more justifiable to support more relevant public names, but perhaps it's better to defer that decision.

### DIFF
--- a/src/Futhark/CodeGen/Backends/GenericC.hs
+++ b/src/Futhark/CodeGen/Backends/GenericC.hs
@@ -1280,7 +1280,7 @@ data CParts = CParts { cHeader :: String
 
 -- | Produce header and implementation files.
 asLibrary :: CParts -> (String, String)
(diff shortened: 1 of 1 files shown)


athas:
Perhaps a better way of doing this is to keep using `#pragma once`, but add it in the [`asLibrary` function](https://github.com/diku-dk/futhark/blob/ce11059400bc78d36c51b9ec66e190d377ccdbba/src/Futhark/CodeGen/Backends/GenericC.hs#L1282).  This will avoid adding it for `.c` files (which is really what #747 is about)/

athas:
Thanks!

PR #802, opened 2019-10-28
Adds link to Parallel Programming in Futhark book

DhruvDh:
I found this book on /r/futhark, and saw that it isn't linked to from anywhere I could see (the README or the website), thought to add it here.


athas:
It's mentioned [on the Docs page](https://futhark-lang.org/docs.html), but the more links, the merrier!  (And our README could do with some improvement anyway.)

PR #817, opened 2019-12-12
Remove unnecessary for-loop in kernelLoop

Munksgaard:
If the number of elements to process is exactly equal to the number of threads
available, we don't need to add a for-loop and can instead execute the code
within directly.

### DIFF
--- a/src/Futhark/CodeGen/ImpGen/Kernels/Base.hs
+++ b/src/Futhark/CodeGen/ImpGen/Kernels/Base.hs
@@ -134,13 +134,16 @@ compileThreadExp dest e =
 -- multidimensional loops, use 'groupCoverSpace'.
 kernelLoop :: Imp.Exp -> Imp.Exp -> Imp.Exp
            -> (Imp.Exp -> InKernelGen ()) -> InKernelGen ()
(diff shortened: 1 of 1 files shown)


Munksgaard:
A question: What if `n < numThreads`? Some of the threads shouldn't execute anything, but perhaps the loop is fine for achieving that?

athas:
Yes, the loop will handle that just fine.

Munksgaard:
> Yes, the loop will handle that just fine.

Although my previous comment didn't make that clear, I was curious if there was a better way to handle that case, or if the loop is the preferred method?

Zyntax9:
Assuming that the compiler can determine that each thread will only have at most a single element, a simplification rule may be more general. One could also be introduce for loops with no iterations. I don't remember whether or not there are such rules, but @athas will know.

athas:
For things like `segmap`s nested inside group `segmap`s I think there is information available about whether the logical iteration count (the size of the SegSpace) is guaranteed to be at most the physical group size.  In that case, you could use a branch instead of a loop.  I have no idea whether there is any performance advantage to that.

This information is encoded in the `SegVirt` field of a `SegLevel`.

PR #818, opened 2019-12-13
Add png2data tool

### LINKED ISSUE #794: More data conversion tools
I wrote [data2png.py](https://github.com/diku-dk/futhark/blob/master/tools/data2png.py) for converting Futhark data dumps to PNG files.  It has already proven quite useful.  I have come to to realise that I would also like a `png2data`, and probably converters for all kinds of other file formats.  Audio?  Why not!  Writing these in Python is probably easiest, but I'm not too picky.

Munksgaard:
Fixes #794, at least partly.

### DIFF
--- a/tools/png2data.py
+++ b/tools/png2data.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python3
+#
+# Turn a PNG image into a Futhark value of type [height][width]i32 encoded in
+# the binary data format.
+# Usage: ./png2data image.png output.data
+#
+# Absolutely no error checking is done.
+#
+# png2data currently supports 24-bit RGB PNGs. Transparency, grayscale, multiple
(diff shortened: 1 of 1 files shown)


athas:
Can you clarify which PNG formats this will work with?  E.g. what happens if there is or is not an alpha channel?

Munksgaard:
Right now it only handles RGB-type PNGs. Transparency and grayscale cause a crash. I'll see if I can make it a bit more robust.

athas:
It's not necessarily important that it can handle anything but RGB.  Just put a comment indicating its current restrictions .

athas:
Also, you might want to document whether the pixels are ARGB or RGBA or what.  I think it's the former based on a quick inspection.

Munksgaard:
I've tried to clarify what png2data supports.

PR #823, opened 2019-12-31
Statically checking size-dependent types for Futhark.

athas:
This pull request replaces #784, because I want to emphasize that it's not so much about dependent types, as about checking sizes.

After half a year of work, this implementation can now handle all our tests and all our benchmarks.  We still need to decide whether we even *want* this (@coancea has been very silent on the topic).  The following remains to be done:

1. Documenting what the rules actually are.  Currently it's just in code. **Done** (sort of).
2. Formalisation with @melsman.  Probably has to go hand-in-hand with (1).
3. Size inference is probably still a bit fishy.  I don't expect we can get full type inference in all cases (nor is it particularly important), but I suspect there may be places where the compiler will infer something that makes no sense.  **Pretty sure it now does a good job.**
4. Size inference for `loop` is particularly fishy, because for various reasons it uses a bespoke mechanism rather than the one based on `let`-generalisation that we use for functions.
5. Interaction with the module system is still unclear. **Non-lifted abstract types may not be instantiated with types that contain anonymous sizes.**
6. For convenience, expressions like `0..<n` and `xs[0:n]` are given sizes of `n`.  However, strictly this is only true if `n>=0`.  We should check explicitly for this at run-time (currently it does do a safety check, but on the size of the produced array, not `n` itself). **Done**
7. Error messages involving existential sizes are probably unacceptably bad.  We should provide some auxiliary information about where the existential sizes come from.  **Done**

Point (6) is an interesting situation, because it's an interaction with the old compilation scheme, which did not trust size annotations.  Most of the internaliser still treats sizes just as advice, and inserts safety checks everywhere.  We can probably do better now.


athas:
I improved the error messages and added some documentation.  We are ready to actually merge this.  We will need to discuss whether to go all-in immediately, or turn size errors into mere size *warnings* for a first release.

athas:
Talked to people at lunch.  Merging this soon.

Issue #827, opened 2020-01-15
What does `futhark bench` measure?


Munksgaard:
It would be nice to have an exact specification of what `futhark bench` measures somewhere, perhaps in the man pages. I'm assuming that kernel compilation is not included in the timing output, but what about creation of and copying data to the GPU buffers before running the actual kernels? What about reading data from and releasing the GPU buffers after running the kernel? What about copying to and from buffers in between running kernels (if that is even possible)? What about setting kernel arguments? I'm sure there could be more questions asked here.

Munksgaard:
This has been closed by #828

PR #828, opened 2020-01-15
Add section about what `futhark bench` measures

Munksgaard:
This is an attempt at addressing my questions in #827. 

The answers have been inferred by inspecting this patch to the rodinia benchmarks created by @athas: https://github.com/diku-dk/futhark-pldi17/blob/master/rodinia_3.1-some-instrumentation.patch. By inspecting the changes to nearestNeighbor.cpp, it looks like writing to buffers is not included, but reading from buffers is.

### DIFF
--- a/docs/man/futhark-bench.rst
+++ b/docs/man/futhark-bench.rst
@@ -108,10 +108,8 @@ WHAT FUTHARK BENCH MEASURES
 ===========================
 
 ``futhark bench`` measures the time it takes to run the given Futhark
-program. It starts measuring after all kernels have been compiled and
-data has been copied to the device buffers but before setting any kernel
-arguments. It stops measuring after the kernels are done running and
-data has been read off the buffer but before releasing the buffers.
+program by passing the ``-t FILE`` option to the generated program. See
+the man page for the specific compiler to see exactly what is measured.
 
 EXAMPLES
 ========
--- a/docs/man/futhark-bench.rst
+++ b/docs/man/futhark-bench.rst
@@ -104,6 +104,15 @@ OPTIONS
   ``foo.fut`` will be passed the tuning file ``foo.fut.tuning`` if it
   exists.
 
+WHAT FUTHARK BENCH MEASURES
+===========================
+
+``futhark bench`` measures the time it takes to run the given Futhark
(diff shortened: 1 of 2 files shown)


docs/man/futhark-bench.rst
```diff
@@ -104,6 +104,15 @@ OPTIONS
   ``foo.fut`` will be passed the tuning file ``foo.fut.tuning`` if it
   exists.
 
+WHAT FUTHARK BENCH MEASURES
+===========================
+
+``futhark bench`` measures the time it takes to run the given futhark
```
athas: Capitalise *Futhark*.

athas:
This is already sort of documented [here](https://futhark.readthedocs.io/en/latest/usage.html#executable-options), but in an oblique form.  I agree it is better to have it in the `futhark-bench` manpage.

athas:
Actually, maybe the `futhark-bench` manpage should mention that it will use the `-t` option to perform the timing, and then we should actually document what `-t` means for each of the compilers.  (It means what you wrote here, so we can keep that, too.)

Munksgaard:
> This is already sort of documented [here](https://futhark.readthedocs.io/en/latest/usage.html#executable-options), but in an oblique form. I agree it is better to have it in the `futhark-bench` manpage.

The page you link says that "writing the result" is not included in the timing, oes that mean that the final calls to `clEnqueueReadBuffer` is not included? My text says that it is.

athas:
It is not.

Munksgaard:
In de1e50b I've tried to move some of the explanation to the futhark-opencl man page. That way we can be more specific about how the particular backend handles timing. What do you think?

Also, how do I generate the man pages locally? It doesn't seem like they're getting installed when I run `stack build && stack install`.

Munksgaard:
Hm, or perhaps it should just be added to https://futhark.readthedocs.io/en/latest/usage.html#opencl-specific-options ? What would you prefer?

athas:
You generate manpages by going into `docs/` and running `make html` (you'll need Sphinx and I think something called sphinxcontrib-bibtex).

I prefer the manpages to have a quick summary of all options taken by the generated binaries, and this is a step on that path.  Your approach is better than just stuffing it in the User's Guide.

PR #835, opened 2020-01-17
Support multiple entry points when auto tuning

### LINKED ISSUE #830: futhark autotune doesn't allow files without main
It seems like `futhark autotune` doesn't allow working on files that don't have a `main` function:

```futhark
-- ==
-- entry: foo
-- compiled random input { [10]f32 }
-- auto output
entry foo (xs: []f32): []f32 = xs
```

```
$ futhark autotune foo.fut
Compiling foo.fut...
user error (Program does not have a 'main' entry point with datasets.)
```

Given that we can run the program just fine (with `--entry-point`) and that both `futhark bench` and `futhark test` work, this seems like a strange restriction.

Munksgaard:
Also supports having no `main` function.

Fixes #830

### DIFF
--- a/src/Futhark/CLI/Autotune.hs
+++ b/src/Futhark/CLI/Autotune.hs
@@ -73,16 +73,18 @@ type RunDataset = Path -> RunPurpose -> IO (Either String ([(String, Int)], Doub
 
 type DatasetName = String
 
-prepare :: AutotuneOptions -> FilePath -> IO [(DatasetName, RunDataset)]
+prepare :: AutotuneOptions -> FilePath -> IO [(DatasetName, RunDataset, T.Text)]
 prepare opts prog = do
   spec <- testSpecFromFileOrDie prog
   copts <- compileOptions opts
 
   truns <-
     case testAction spec of
-      RunCases ios _ _
-        | Just runs <-
-            iosTestRuns <$> find ((=="main") . iosEntryPoint) ios -> do
+      RunCases ios _ _ | not $ null ios -> do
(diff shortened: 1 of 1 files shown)


Munksgaard:
The implementation currently doesn't try to guess anything about which tuning parameters belong to which function.

Initially I thought it would be easy to do based on the prefix of the tuning parameters: For each entry point, only attempt to tune the parameters that have a matching prefix. However, I don't think that'll work if we have functions calling other functions:

```futhark
let bar(...) =
  ...

let main(...) =
  ... bar is not used here ...

entry foo() =
  ... bar is used here ...
```

How will the autotuner know that tuning parameters related to `bar` should only be tuned when running the `foo` entry point?

athas:
The threshold parameters are named for each *entry point*.  In practice, `bar` is always inlined.

Munksgaard:
> The threshold parameters are named for each _entry point_. In practice, `bar` is always inlined.

I see! In that case I'll see what I can do to include that change.

Munksgaard:
Do we have any testing for the autotune feature?

athas:
Nope, not yet!

athas:
Does this approach maintain already existing threshold parameters for entry points that are not being tuned?  It's not a blocker if it doesn't; we can add that later.

Munksgaard:
I don't think it's currently possible to selectively choose which entry points to tune or not. All entry points are tuned each time the autotuner is run, as far as I can tell, so that shouldn't be an issue.

PR #855, opened 2020-02-19
Index function generalization/antiunification across if-then-else

Munksgaard:
The purpose of this commit is to decrease copies in if-then-else branches.

Before, if two branches of an if-then-else returned arrays with different index
functions (LMADs), the compiler would normalize the arrays in one or both of the
branches, which could result in unnecessary copying.

With this commit, the compiler will try to generalize the index functions and
return the existentials in addition to the array itself. Consider the following
expression:

```futhark
  if (arr[0] < 0.0)
  then arr[2+n/2:n]
  else arr[2+n/4:n]
```

Before, the results would be copied into fresh arrays in both branches, but
really, we could return the same array as before in addition to the computed
offsets to the index function. This way, we can avoid some unnecessary copies.

Along the way, we've had to relax the type checker for the ExplicitMemory
representation a bit. To illustrate why, consider the expression above. In some
cases the simplifier will hoist the computation of the offset to outside the
generated branches, meaning that the type checker cannot know that the offset
computed and the offset of the returned array match. We therefore introduce the
function `closeEnough` which we hope is good enough to catch future errors in
the code generation, but which is relaxed enough to let this particular case
through.

We also ran into a case in ImpGen where the assumption that all
allocations (outside of the return of lambda functions) are explicit in the IR
wasn't being respected.

The changes pass all unit tests and seem to have no regression in benchmark
performance. There doesn't seem to be any speedup either, but hopefully this
change will pave the way for future changes that have more impact on efficiency
of the generated code.

### DIFF
--- a/src/Futhark/Analysis/PrimExp.hs
+++ b/src/Futhark/Analysis/PrimExp.hs
@@ -7,6 +7,7 @@ module Futhark.Analysis.PrimExp
   , primExpType
   , primExpSizeAtLeast
   , coerceIntPrimExp
+  , leafExpTypes
   , true
   , false
   , constFoldPrimExp
@@ -18,6 +19,7 @@ module Futhark.Analysis.PrimExp
 import           Control.Monad
 import           Data.Traversable
 import qualified Data.Map as M
+import qualified Data.Set as S
 
 import           Futhark.Representation.AST.Attributes.Names
 import           Futhark.Representation.Primitive
@@ -345,3 +347,15 @@ instance Pretty v => Pretty (PrimExp v) where
   ppr (ConvOpExp op x)  = ppr op <+> parens (ppr x)
   ppr (UnOpExp op x)    = ppr op <+> parens (ppr x)
   ppr (FunExp h args _) = text h <+> parens (commasep $ map ppr args)
+
+leafExpTypes :: Ord a => PrimExp a -> S.Set (a, PrimType)
+leafExpTypes (LeafExp x ptp) = S.singleton (x, ptp)
+leafExpTypes (ValueExp _) = S.empty
+leafExpTypes (UnOpExp _ e) = leafExpTypes e
+leafExpTypes (ConvOpExp _ e) = leafExpTypes e
+leafExpTypes (BinOpExp _ e1 e2) =
+  S.union (leafExpTypes e1) (leafExpTypes e2)
+leafExpTypes (CmpOpExp _ e1 e2) =
+  S.union (leafExpTypes e1) (leafExpTypes e2)
+leafExpTypes (FunExp _ pes _) =
+  S.unions $ map leafExpTypes pes
--- a/src/Futhark/Pass/ExplicitAllocations.hs
+++ b/src/Futhark/Pass/ExplicitAllocations.hs
@@ -20,6 +20,7 @@ import qualified Data.Map.Strict as M
 import qualified Data.Set as S
 import qualified Control.Monad.Fail as Fail
 import Data.Maybe
+import Data.List (zip4, partition)
 
 import Futhark.Representation.Kernels
 import Futhark.Optimise.Simplify.Lore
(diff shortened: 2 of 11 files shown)


athas:
What does this actually *do*?  What are the performance implications, if any, on the benchmark suite?  Use `futhark bench` and [`cmp-bench-json.py`](https://github.com/diku-dk/futhark/blob/master/tools/cmp-bench-json.py) to summarise.

src/Futhark/Pass/ExplicitAllocations.hs
```diff
@@ -611,25 +643,68 @@ allocInExp (DoLoop ctx val form (Body () bodybnds bodyres)) =
...
+        res_else = bodyResult fbranch'
+        size_ext = length res_then - length trets
+        (ind_ses0, r_then_else) =
+          foldl (\(acc_ise, acc_ext) (r_then, r_else, i) ->
+                   if r_then == r_else then ((i, r_then) : acc_ise, acc_ext)
+                   else (acc_ise, (r_then, r_else) : acc_ext)
+                ) ([], []) $ reverse $ zip3 res_then res_else [0 .. size_ext - 1]
+        (r_then_ext, r_else_ext) = unzip r_then_else
+        ind_ses = zipWith (\(i, se) k -> (i-k, se)) ind_ses0 [0 .. length ind_ses0 - 1]
+        rets'' = foldl(\acc (i, se) -> fixExt i se acc) trets ind_ses
```
athas: Missing space.

src/Futhark/Pass/ExplicitAllocations.hs
```diff
@@ -611,25 +643,68 @@ allocInExp (DoLoop ctx val form (Body () bodybnds bodyres)) =
...
+      fsubs = map (selectSub snd) subs
+  (tbranch', trets) <- addResCtxInIfBody rets tbranch spaces tsubs
+  (fbranch', frets) <- addResCtxInIfBody rets fbranch spaces fsubs
+  if frets /= trets then error "In allocInExp, IF case: antiunification of then/else produce different ExtInFn!"
+    else do -- above is a sanity check; implementation continues on else branch
+    let res_then = bodyResult tbranch'
+        res_else = bodyResult fbranch'
+        size_ext = length res_then - length trets
+        (ind_ses0, r_then_else) =
+          foldl (\(acc_ise, acc_ext) (r_then, r_else, i) ->
```
athas: I don't like multi-line lambdas, except in `do` context.  Could you give this lambda a local name instead?  Also, isn't this just `partition`?

src/Futhark/Pass/ExplicitAllocations.hs
```diff
@@ -641,6 +716,64 @@ allocInExp e = mapExpM alloc e
...
+                     [Maybe (Int, ExtIxFun, M.Map Int (PrimExp VName))] ->
+                     AllocM fromlore tolore (Body tolore, [BodyReturns])
+addResCtxInIfBody ifrets (Body _ bnds res) spaces substs = do
+  let num_vals = length ifrets
+      (ctx_res, val_res) = splitFromEnd num_vals res
+      ext_ini = S.size $ shapeContext ifrets
+  ((res', bodyrets'), all_body_stms) <- collectStms $ do
+    mapM_ addStm bnds
+    (val_res', mem_ctx_res, bodyrets, _) <-
+      foldM (\(res_acc, ctx_acc, br_acc, k) (ifr, r, mbixfsub, sp) ->
```
athas: Give this giant fold function a local binding.

src/Futhark/Analysis/PrimExp.hs
```diff
@@ -345,3 +349,74 @@ instance Pretty v => Pretty (PrimExp v) where
   ppr (ConvOpExp op x)  = ppr op <+> parens (ppr x)
   ppr (UnOpExp op x)    = ppr op <+> parens (ppr x)
   ppr (FunExp h args _) = text h <+> parens (commasep $ map ppr args)
+
+getPrimExpExtsWithPTp :: PrimExp (Ext v) -> S.Set (Int,PrimType)
```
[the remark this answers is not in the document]
Munksgaard: The updated commit addresses this.

src/Futhark/Analysis/PrimExp.hs
```diff
@@ -345,3 +349,74 @@ instance Pretty v => Pretty (PrimExp v) where
...
+getPrimExpExtsWithPTp (CmpOpExp _ e1 e2) =
+  S.union (getPrimExpExtsWithPTp e1) (getPrimExpExtsWithPTp e2)
+getPrimExpExtsWithPTp (FunExp _ pes _) =
+  S.unions $ map getPrimExpExtsWithPTp pes
+
+type MapPExp v = M.Map Int (PrimExp v, PrimExp v)
+
+-- | Generalization (anti-unification)
+-- We assume that the two expressions have the same type.
+leastGeneralGeneralization :: (Eq v) => Int -> PrimExp v -> PrimExp v ->
```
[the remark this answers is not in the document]
Munksgaard: In the updated commit I've added src/Futhark/Analysis/PrimExp/Generalize.hs for this function.

src/Futhark/Pass/ExplicitAllocations.hs
```diff
@@ -265,24 +266,33 @@ allocsForPattern sizeidents validents rts hints = do
           return $ PatElem (identName ident) $
           MemMem space
 
-        MemArray bt _ u (Just (ReturnsInBlock mem ixfun)) ->
-          PatElem (identName ident) . MemArray bt shape u .
-          ArrayIn mem <$> instantiateIxFun ixfun
+        MemArray bt _ u (Just (ReturnsInBlock mem extixfun)) -> do
+          (ixfn_idptps, ixfn) <- instantiateExtIxFn ident extixfun
+          let patels = map (\(nm, ptp) -> PatElem nm $ MemPrim ptp) ixfn_idptps
```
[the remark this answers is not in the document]
Munksgaard: Good catch, I've done so.

src/Futhark/Pass/ExplicitAllocations.hs
```diff
@@ -291,6 +301,35 @@ allocsForPattern sizeidents validents rts hints = do
         known (Free v) = Just v
         known Ext{} = Nothing
 
+        mkMemIdent :: (MonadFreshNames m) => Ident -> Space -> m Ident
+        mkMemIdent ident space = do
+          let memname = baseString (identName ident) <> "_mem"
+          newIdent memname $ Mem space
+
+        instantiateExtIxFn :: (MonadFreshNames m) =>
```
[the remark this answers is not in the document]
Munksgaard: Done!

src/Futhark/Pass/ExplicitAllocations.hs
```diff
@@ -291,6 +301,35 @@ allocsForPattern sizeidents validents rts hints = do
...
+                              Ident -> ExtIxFun -> m ([(VName, PrimType)], IxFun)
+        instantiateExtIxFn idd ext_ixfn = do
+          let idnm = baseString (identName idd) <> "_ixfn"
+              (is0, ptps) = unzip $ S.toList $
+                            foldMap getPrimExpExtsWithPTp ext_ixfn
+              is = S.toList $ S.fromList is0
```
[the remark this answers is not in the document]
Munksgaard: You're right, fixed.

src/Futhark/Pass/ExplicitAllocations.hs
Munksgaard: :+1:

src/Futhark/Pass/ExplicitAllocations.hs
Munksgaard: I've rewritten this bit using partition.

src/Futhark/Pass/ExplicitAllocations.hs
Munksgaard: Done!

athas:
By an amazing coincidence, we suddenly need this functionality in order to implement register tiling.

PR #857, opened 2020-02-26
Reuse generalizations and remove extraneous pattern values exps

Munksgaard:
This commit attempts to fix some errors in the recent existentialization across
if branches, as exhibited here:
https://travis-ci.org/diku-dk/sorts/builds/655258023

Unfortuntately, it still doesn't seem to work...

### DIFF
--- a/src/Futhark/Pass/ExplicitAllocations.hs
+++ b/src/Futhark/Pass/ExplicitAllocations.hs
@@ -20,7 +20,7 @@ import qualified Data.Map.Strict as M
 import qualified Data.Set as S
 import qualified Control.Monad.Fail as Fail
 import Data.Maybe
-import Data.List (zip4, partition)
+import Data.List (zip4, partition, union)
 
 import Futhark.Representation.Kernels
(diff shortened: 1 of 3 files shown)


athas:
What about the test programs?

src/Futhark/Pass/ExplicitAllocations.hs
```diff
@@ -733,7 +733,7 @@ addResCtxInIfBody ifrets (Body _ bnds res) spaces substs = do
     mapM_ addStm bnds
     (val_res', mem_ctx_res, bodyrets, _) <-
       foldM helper ([], [], [], ext_ini) (zip4 ifrets val_res substs spaces)
-    return (ctx_res <> mem_ctx_res <> val_res', bodyrets)
+    return (ctx_res `union` mem_ctx_res <> val_res', bodyrets)
```
athas: This seems risky to me, as the values returned from the context must match exactly the order in which they occur in the pattern (and the existentials in whatever type is matched to this body).  I have a hard time seeing that treating this as a set can be correct.

PR #865, opened 2020-03-10
Drop hpack.

athas:


flip111:
what's the motivation to drop hpack support ?

athas:
Mostly Windows.  Also to eliminate an intermediate step when using `cabal`-based builds, which is what e.g. Homebrew prefers.

flip111:
What's the problem of hpack on windows and mac? Maybe hpack can be improved.

athas:
Nothing deep.  On Windows the main problem is that `hpack` isn't in Chocolatey.  On Homebrew the problem isn't so great, because `hpack` is available (after I added it), but I guess it's simpler for users that they don't have to run it.

flip111:
Wouldn't including the generated cabal file into the repository be enough to avoid running hpack? Then hpack can be used by contributors of futhark.

athas:
Sure, but it's yet another moving part that can break, and I can no longer justify the cost.

flip111:
This is an even stronger motivation why to drop hpack support https://gist.github.com/merijn/8152d561fb8b011f9313c48d876ceb07#so-which-tool-should-i-use

PR #866, opened 2020-03-10
Add support for mad_hi and mul_hi.

porcuquine:
This PR adds the `mul_hi` and `mad_hi` primitives for the OpenCL backend. Emulation in the interpreter and C backend works as expected. The C Sharp and Python backends are not implemented here.


athas:
Not bad!  And you even made the tests a lot better than the ones for the other primitives.  I have a few style remarks (and maybe CI will have a few more), but it looks good.

src/Futhark/Representation/Primitive.hs
```diff
@@ -1134,3 +1184,39 @@ convOp s from to = text s <> text "_" <> ppr from <> text "_" <> ppr to
 prettySigned :: Bool -> PrimType -> String
 prettySigned True (IntType it) = 'u' : drop 1 (pretty it)
 prettySigned _ t = pretty t
+
+mul_hi8 :: IntValue -> IntValue -> Int8
+mul_hi8 a b =
+  let a' = intToWord64 a in
+    let b' = intToWord64 b in
+      fromIntegral (shiftR (a' * b') 8)
```
athas: More idiomatic Haskell:

```haskell
  let a' = intToWord64 a
      b' = intToWord64 b
  in fromIntegral (shiftR (a' * b') 8)
```

porcuquine: Thanks, fixed.

tests/primitive/mad_hi.fut
```diff
@@ -0,0 +1,88 @@
+-- Test u8.mad_hi
+-- tags { no_csharp no_python}
```
athas: The `tags` field need to be after the `==` line.

porcuquine: Fixed and fixed some test failures introduced when previously fixing parsing errors.

PR #867, opened 2020-03-13
Implement binary searching in Autotuner

Munksgaard:
The purpose of this PR is to improve the autotuner by taking into account
programs with changing levels of parallelism. Previous to this commit, if a
program had different levels of parallelism, called `ePar`, the autotuner would
choose the first one it encountered. In the LUD benchmark, that would always be
the highest `ePar`, since the level of parallelism is decreasing. But how do we
know that this is the best ePar to compare against?

Assuming that program performance is monotonic in the threshold value chosen, we
can instead try to find the best `ePar` by binary search over all the possible
values. Of course, autotuning will take longer if there are more `ePar`s, but
this is not the case for most of our benchmarks, and for those that it is, we
hope that we can get a better `ePar` as a result.

The runtime for the autotuner should only increase by `log(n)`, where `n` is the
number of `ePar`s in a given program. Tuning LUD goes from 163s to 300s on
gpu04.

~~Unfortunately, for LUD, this commit doesn't seem to improve the runtime.
Choosing the first `ePar` happens to be the correct choice.~~

Edit: I now seem to be consistently getting better results with the new autotuner:

```
lud.fut
  data/512.in:                                                          2.16x
  data/64.in:                                                           0.99x
  data/256.in:                                                          1.35x
  data/16by16.in:                                                       1.00x
  data/2048.in:                                                         1.47x
```

### DIFF
--- a/src/Futhark/Bench.hs
+++ b/src/Futhark/Bench.hs
@@ -156,6 +156,7 @@ data RunOptions =
   , runRuns :: Int
   , runExtraOptions :: [String]
   , runTimeout :: Int
+  , runVerbose :: Int
   }
 
 -- | Run the benchmark program on the indicated dataset.
@@ -185,6 +186,10 @@ benchmarkDataset opts program entry input_spec expected_spec ref_out =
   let (to_run, to_run_args)
         | null $ runRunner opts = ("." </> binaryName program, options)
         | otherwise = (runRunner opts, binaryName program : options)
+
+  when (runVerbose opts > 0) $
+    putStrLn $ unwords ["Running executable", show to_run,
+                        "with arguments", show to_run_args]
 
   run_res <-
     timeout (runTimeout opts * 1000000) $
--- a/src/Futhark/CLI/Autotune.hs
+++ b/src/Futhark/CLI/Autotune.hs
@@ -30,11 +30,12 @@ data AutotuneOptions = AutotuneOptions
                     , optTuning :: Maybe String
                     , optExtraOptions :: [String]
                     , optVerbose :: Int
+                    , optTimeout :: Int
                     }
 
 initialAutotuneOptions :: AutotuneOptions
 initialAutotuneOptions =
-  AutotuneOptions "opencl" Nothing 10 (Just "tuning") [] 0
+  AutotuneOptions "opencl" Nothing 10 (Just "tuning") [] 0 60
 
 compileOptions :: AutotuneOptions -> IO CompileOptions
 compileOptions opts = do
@@ -112,7 +113,7 @@ prepare opts prog = do
     forM (mapMaybe (runnableDataset $ iosEntryPoint ios)
                    (iosTestRuns ios)) $
       \(dataset, do_run) -> do
-      res <- do_run 60000 [] RunBenchmark
+      res <- do_run (optTimeout opts) [] RunBenchmark
       case res of Left err -> do
                     putStrLn $ "Error when running " ++ prog ++ ":"
                     putStrLn err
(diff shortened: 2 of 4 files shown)


melsman:
Is there an example of a case where the extra work pays of?

flip111:
Hi @Munksgaard how do you figure that binary search can be applied here? Perhaps it's also worth to try every `ePar`

Munksgaard:
> Hi @Munksgaard how do you figure that binary search can be applied here? Perhaps it's also worth to try every `ePar`

The advantage over trying every `ePar` is the reduced run-time. Imagine a for-loop with decreasing (or increasing) `ePar` for every iteration. If the loop has thoussands of iterations, tuning will take a _long_ time. The hope is that we can be "good enough" using the binary search, while only taking slightly longer than the old implementation.

Munksgaard:
> Is there an example of a case where the extra work pays of?

Well, I may have botched my earlier results, because now I'm getting a consistent improvement in the performance of the LUD benchmark with my new binary tuner:

```
$ ../../../tools/cmp-bench-json.py old.json  binary.json 

lud.fut
  data/512.in:                                                          2.16x
  data/64.in:                                                           0.99x
  data/256.in:                                                          1.35x
  data/16by16.in:                                                       1.00x
  data/2048.in:                                                         1.47x
```

where old.json is benchmark data with the old autotuner, and binary.json is benchmark data with the new binary autotuner. I'd say that's pretty good, but I wish I knew why I was getting other results before. My guess is that I messed something up when comparing the results.

flip111:
> The hope is that we can be "good enough" using the binary search

Can you test every `ePar` separately to see if significant more optimal parallelization points can be found? If this is the case maybe linear search can be put behind a compiler flag.

src/Futhark/Bench.hs
```diff
@@ -157,6 +157,7 @@ data RunOptions =
   , runExtraOptions :: [String]
   , runTimeout :: Int
   }
+  deriving Show
```
athas: Why?

src/Futhark/CLI/Autotune.hs
```diff
@@ -30,6 +29,7 @@ data AutotuneOptions = AutotuneOptions
                     , optExtraOptions :: [String]
                     , optVerbose :: Int
                     }
+  deriving Show
```
athas: Why?

src/Futhark/CLI/Autotune.hs
```diff
@@ -207,80 +197,107 @@ thresholdForest prog = do
...
-          when (optVerbose opts > 0) $ putStrLn $
-            "Sampling run failed:\n" ++ err
-          return (thresholdMin, thresholdMax)
-        Right (cmps, _) ->
-          case lookup v cmps of
+tuneThreshold opts datasets already_tuned (v, _v_path) = do
+  putStrLn $ unwords ["Tuning threshold", v]
+
+  (e_pars', _) <- getAllEPars already_tuned v
+  let e_pars = sort $ nub $ thresholdMin : thresholdMax : e_pars'
```
athas: `nub` is going to destroy us eventually; go through a Set to remove the duplicates.

src/Futhark/Bench.hs
Munksgaard: Leftover from debugging...

src/Futhark/CLI/Autotune.hs
Munksgaard: Leftover from debugging...

Munksgaard:
> > The hope is that we can be "good enough" using the binary search
> 
> Can you test every `ePar` separately to see if significant more optimal parallelization points can be found? If this is the case maybe linear search can be put behind a compiler flag.

That is certainly an option, but I don't think it's relevant for this particular feature. What do you think, @athas ?

athas:
If it's easy to add as an option, why not, but let's leave it for a future PR.

Munksgaard:
Wait a bit with merging this, we want to handle some more cases.

athas:
Is this ready soon?

Munksgaard:
> Is this ready soon?

There are some weaknesses in the tuner that I want to try to address as soon as I'm done writing this paper. Specifically:

 - It doesn't handle multiple datasets very nicely. It just sums their run times.
 - It should use the the best performing test run instead of the average when determining the run time in a single benchmark run.

Munksgaard:
If you want, we can close the pull request, and I can resubmit it when I'm ready?

athas:
No need, let's leave it open as motivation for both of us!

src/Futhark/CLI/Autotune.hs
```diff
@@ -130,12 +130,17 @@ prepare opts prog = do
                  fromIntegral (sum (map runMicroseconds runres)) /
                  fromIntegral (optRuns opts))
 
+              bestRuntime :: ([RunResult], T.Text) -> ([(String, Int)], Int)
+              bestRuntime (runres, errout) =
+                (comparisons (T.unpack errout),
+                 maximum $ map runMicroseconds runres)
```
Munksgaard, line 132: This should be minimum...

PR #888, opened 2020-03-25
Fix "being" typo

NPN:


athas:
This should actually be "being", not "behind".

NPN:
Oops, sorry. Should be fixed now.

athas:
Thanks!

PR #892, opened 2020-03-27
This is fine to stack-allocate.

athas:


athas:
Don't forget about [the other one](https://github.com/diku-dk/futhark/pull/893)!

PR #903, opened 2020-04-07
Add timeout parameter to `futhark autotune`

Munksgaard:


docs/man/futhark-autotune.rst
```diff
@@ -59,6 +59,14 @@ OPTIONS
 
   Change the extension used for tuning files (``.tuning`` by default).
 
+--timeout=seconds
+
+  Initial tuning timeout for each dataset in seconds. After running the intitial
```
athas, line 64: Are you sure this is in seconds?  I think it should be (for commonality with `futhark bench`), but I don't think that's what the code does.

The default timeout should certainly be 60 seconds, not 60000 seconds.

Munksgaard: I'll verify and update the PR.

Munksgaard: Turns out that it actually was set to 60000 seconds by default.... It's now set to 60.

PR #904, opened 2020-04-07
Use stock GenericC.hs for multicore backend

athas:
Ideally, we should not need to modify GenericC for the multicore backend to work.  This moves us closer to that dream.


HnimNart:
Seems like this messes up the order of declaring static arrays after usage, for example with curry1.fut 
Which section declares static arrays?

athas:
Oh shit, I thought I tested that.  I'll fix it.

PR #906, opened 2020-04-07
Fix multicore backend for programs that contain safety checks.

athas:
With this change, the Futhark ray tracer works with the multicore backend.  And it's actually the second-fastest CPU implementation, by a significant margin!


athas:
What I'm doing here is starting a new "compiler instance" (for lack of a better word) to generate the task function body.  This is because GenericC always assumes it is compiling a "top level" function, where everything declared so far is still in scope.

PR #907, opened 2020-04-09
 Implement binary searching in Autotuner, take 2

Munksgaard:
The purpose of this PR is to improve the autotuner by taking into account programs with changing levels of parallelism. Previous to this commit, if a program had different levels of parallelism, we call them `ePar`, the autotuner would choose the first one it encountered. In the LUD benchmark, that would always be the highest `ePar`, since the level of parallelism is decreasing. But how do we know that this is the best ePar to compare against?

With these changes, the autotuner will be equipped to handle such cases. It employs a binary search combined with continual refinement of the search space to efficiently find the threshold value that guarantees optimal performance across all test datasets.

The changes build upon the following realization, which also motivated not using ranges for size-invariant parallelism tuning: If, for a given program, the optimal threshold parameter value for one dataset is `x`, the optimal threshold parameter value for another dataset is `y`, and `y` < `x`, it can never hurt the performance of the program on the first dataset to use `y` as the threshold parameter. If the optimal threshold parameter value found for the first dataset is `x` and not `y` or some value in between, it's just because that amount of parallelism isn't exposed when running the program on the first dataset.

After coming to this realization, the only thing we need to do in order to tune a particular threshold in a program with size-variant parallelism, is to find the optimal threshold parameter value for each dataset and then take the minimum of those. Using a binary search is a straight-forward to find the optimal threshold parameter value for each dataset. However, because of the realization above, we can improve the heuristics of the binary search quite a bit: Every time we have tuned the threshold on a given dataset and found the optimal threshold parameter value `x`, when we tune on the next dataset, there's no need to try any possible values of `ePar` above `x`. In addition, if there were any values of `ePar` below `x` in the first dataset, there's no need to try any of those again on the second dataset. This means that after tuning the first dataset, the only candidates we have for further improving the threshold on the next dataset, are those `ePar`s that are between `y` and `x`, where `x` is the optimal threshold parameter value found for the first dataset, and `y` is the maximum `ePar` less than `x` in the first dataset. This heuristic greatly improves the time taken to tune a program with size-invariant parallelism over the naive binary tuner.

The result is a tuner that gives us optimal threshold parameter values both for size-variant and -invariant programs. It is faster when tuning size-invariant programs than the old tuner because of the heuristics described above. It is also much faster than a previous attempt at rewriting the autotuner when tuning size-variant parallelism.

### DIFF
--- a/src/Futhark/Bench.hs
+++ b/src/Futhark/Bench.hs
@@ -187,7 +187,7 @@ benchmarkDataset opts program entry input_spec expected_spec ref_out =
         | null $ runRunner opts = ("." </> binaryName program, options)
         | otherwise = (runRunner opts, binaryName program : options)
 
-  when (runVerbose opts > 0) $
+  when (runVerbose opts > 1) $
     putStrLn $ unwords ["Running executable", show to_run,
                         "with arguments", show to_run_args]
 
--- a/src/Futhark/CLI/Autotune.hs
+++ b/src/Futhark/CLI/Autotune.hs
@@ -4,13 +4,14 @@ module Futhark.CLI.Autotune (main) where
 
 import Control.Monad
 import qualified Data.ByteString.Char8 as SBS
-import Data.Time.Clock.POSIX
+import Data.Function (on)
 import Data.Tree
-import Data.List (intersect, isPrefixOf, foldl', sort, sortOn)
+import Data.List (intersect, isPrefixOf, sort, sortOn, elemIndex, minimumBy)
 import Data.Maybe
 import Text.Read (readMaybe)
 import Text.Regex.TDFA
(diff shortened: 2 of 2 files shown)


Munksgaard:
All in all, while the binary search code is a little convoluted and cumbersome, I think this is a much nicer solution to autotuning size-variant parallelism. It actually gives optimal results, both for size-variant and -invariant parallelism and it is reasonably fast.

Munksgaard:
We may have to rewrite both chapters 3 and 4 in our paper though...

Munksgaard:
The remaining CI errors are just azure acting up, as far as I can tell.

athas:
If @Microsoft-TeamLeader says this is good, who am I to argue?

PR #908, opened 2020-04-10
Fix typo in prelude/soacs docs

tomsmeding:
The docs referred to a neutral element for "`op`", but this particular function (`reduce_by_index`) has the relevant parameter named `f`. This PR fixes the typo.


athas:
Thanks!

PR #911, opened 2020-04-11
Implement constants in the core language.

athas:
With this commit, the core AST now has a distinct notion of constants,
as something separate from function definitions.  All constants are in
scope in all functions.  At run-time, the constants are only computed
once.

This is a rather major change, that required modifications to
essentially all compiler passes, and all code generation backends.
Compile times seem mostly unaffected on large programs, with small
regressions on small programs.  The advantage is that programs that
make heavy use of complex constants will now run much faster in
library mode.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -340,6 +340,11 @@ following function::
 
   struct futhark_context *futhark_context_new(struct futhark_context_config *cfg);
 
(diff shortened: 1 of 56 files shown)


flip111:
Does this help in dead code removal ?

athas:
No.

Issue #912, opened 2020-04-12
Error formatting improvements


mxxo:
I want to try improving Futhark's command-line errors. I've been going through the nice work done in #634 and would like to try adding colour output and caret diagnostics to typechecking errors. 

For example, the duplicate parameters (https://github.com/diku-dk/futhark/blob/master/tests/errors/duplicate-params.fut) error is 

```text
Error at .\tests\errors\duplicate-params.fut:6:20-20:
Name x also bound at .\tests\errors\duplicate-params.fut:6:11-11
If you find this error message confusing, uninformative, or wrong, please open an issue at https://github.com/diku-dk/futhark/issues.
```

which is understandable, but somewhat rudimentary. 

Compared to the same error in Rust: 

```text
error[E0415]: identifier `x` is bound more than once in this parameter list
 --> src\main.rs:5:22
  |
5 | fn dup_param(x: i32, x: i32) {
  |                      ^ used as parameter more than once

error: aborting due to previous error

For more information about this error, try `rustc --explain E0415`.
error: Could not compile `errs`.
```

which is a bit of a unfair comparison (general-purpose language vs. desert hedgehog), but perhaps there's ideas we can take advantage from to make debugging easier. 

Thank you for your time and all your work on Futhark.

athas:
I certainly wouldn't mind something like this.  I don't know exactly how other languages implement this, but I think Futhark already carries around most of the necessary parts.  In particular, we do maintain pretty complete source location information (filename, line, and column span).  I think if we also carry around a full in-memory copy of all the input files during type checking (which isn't all that much data), we could just do a lookup in those based on the source location.

flip111:
Great idea @mxxo !

mxxo:
What kind of error format would you like to see? `clang`, `rustc` and recent `gcc`s come to mind for good error messages, and there's some nice error formatting documents here: 
- https://github.com/rust-lang/rfcs/blob/master/text/1644-default-and-expanded-rustc-errors.md
- https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html
- http://www.jonathanturner.org/2016/08/helping-out-with-rust-errors.html

I think the things we could improve are: 
1. Use the `SrcLoc` information (and a copy of the input file like you mentioned) to print out where the error happens (with some caret helpers or --- underlining etc). Some of the printing logic is a little hairy, I don't know how far we want to go here. 
[...]

PR #913, opened 2020-04-12
Print TypeError "Error" in red

mxxo:
Print the "Error" part of a TypeError in red with the error message right after it. 
(c.f. #912)

For [duplicate parameters](https://github.com/diku-dk/futhark/blob/master/tests/errors/duplicate-params.fut), the error message is: 
```text
Error: Name "x" also bound at .\tests\errors\duplicate-params.fut:6:11-11.                                
at: .\tests\errors\duplicate-params.fut:6:20-20                          
```
image

(paths are weird because I'm on windows, double "at" is a little clunky)

instead of 

```text
Error at .\tests\errors\duplicate-params.fut:6:20-20:
Name x also bound at .\tests\errors\duplicate-params.fut:6:11-11
If you find this error message confusing, uninformative, or wrong, please open an issue at https://github.com/diku-dk/futhark/issues.
```

image

### DIFF
--- a/src/Language/Futhark/TypeChecker/Monad.hs
+++ b/src/Language/Futhark/TypeChecker/Monad.hs
@@ -97,7 +97,7 @@ data TypeError = TypeError SrcLoc Notes Doc
 
 instance Pretty TypeError where
   ppr (TypeError loc notes msg) =
-    "Error at" <+> text (locStr loc) <> ":" </>
+    text (inRed "Error") <+> "at" <+> text (locStr loc) <> ":" </>
     msg <> ppr notes
 
 unexpectedType :: MonadTypeChecker m => SrcLoc -> StructType -> [StructType] -> m a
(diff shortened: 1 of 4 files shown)


athas:
Don't worry about the Windows CI job failing.  That's GHC being flaky (it segfaults sometimes for reasons beyond my comprehension).

src/Language/Futhark/TypeChecker/Monad.hs
```diff
@@ -97,8 +97,8 @@ data TypeError = TypeError SrcLoc Notes Doc
 
 instance Pretty TypeError where
   ppr (TypeError loc notes msg) =
-    "Error at" <+> text (locStr loc) <> ":" </>
-    msg <> ppr notes
+    text (inRed "Error") <> ":" <+> msg </>
+    "  at:" <+> text (locStr loc) <+> ppr notes
```
athas, line 101: I don't think this works grammatically for many of the existing error messages.  Could you maintain the current phrasing and just put `Error` in red?  I'm OK with discussing whether the overall format for errors should change, but it seems like a distinct thing from introducing colour.

mxxo: Ok, sounds good! Better to rephrase later.

PR #915, opened 2020-04-13
Multicore merge

athas:
I broke some things, and here is the fix.

I also removed a call to `futhark_context_get_error()` from the scheduler.  The scheduler should ideally not use Futhark API functions, and I also don't think it's necessary.


HnimNart:
Yes I agree,  it was just a hack for getting the error message, such that the safety tests passed.

athas:
But the error message is in any case kept inside the `ctx`, which is available everywhere.  As long as you propagate the erroneous *return value*, and that something, somewhere, checks `ctx->error` eventually, then it's fine.  (But propagating that return value is also an interesting problem in itself!  But we can solve it.)

HnimNart:
Yes I just thought of that as well, I'll try to fix it now.

PR #918, opened 2020-04-14
Enable GCC output colorization by default.

emilmasoumi:
This makes debugging GCC output more readable.

### DIFF
--- a/src/Futhark/CLI/C.hs
+++ b/src/Futhark/CLI/C.hs
@@ -27,7 +27,8 @@ main = compilerMain () []
            ToExecutable -> do
              liftIO $ writeFile cpath $ SequentialC.asExecutable cprog
              ret <- liftIO $ runProgramWithExitCode "gcc"
-                    [cpath, "-O3", "-std=c99", "-lm", "-o", outpath] mempty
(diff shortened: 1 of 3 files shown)


athas:
There is an error in the option ordering.

src/Futhark/CLI/C.hs
```diff
@@ -27,7 +27,8 @@ main = compilerMain () []
            ToExecutable -> do
              liftIO $ writeFile cpath $ SequentialC.asExecutable cprog
              ret <- liftIO $ runProgramWithExitCode "gcc"
-                    [cpath, "-O3", "-std=c99", "-lm", "-o", outpath] mempty
+                    [cpath, "-O3", "-std=c99", "-lm", "-o",
+                     "-fdiagnostics-color=always", outpath] mempty
              case ret of
                Left err ->
                  externalErrorS $ "Failed to run gcc: " ++ show err
```
Microsoft-TeamLeader, line 34: Some error is coming, everytime it's different

athas:
Looks like `-fdiagnostics-color=always` is unsupported on the version of GCC that comes bundled with some of the distributions we still care about (like the Ubuntu running on Travis).  Maybe in a few years...

PR #933, opened 2020-04-22
Use invariants provided by assertions

### LINKED ISSUE #26: Use invariants provided by assertions
If you have an expression like `assert(x > 10)`, the compiler is currently not able to take advantage of this fact to optimise away the branch in e.g. `if x > 10 then ...`.

Modify the simplifier to refine the ranges in the symbol table whenever it encounters assertions.  Do it the naive way.

Munksgaard:
This is an attempt to fix #26.

The implementation is very simplistic right now, and I'm not even sure if I'm going about it in the right way, but it seems to pass my test and not mess up any of the other tests.

To be clear, it handles something like this:

```
let main(x: i32): i32 =
  assert(x > 32) (if x > 32 then 42 else 1337)
```

But not something like this:

```
let main(x: i32): i32 =
  assert(x <= 32) (if x > 32 then 42 else 1337)
```

Any pointer on how to improve the PR would be greatly appreciated.

### DIFF
--- a/src/Futhark/Optimise/Simplify/Rules.hs
+++ b/src/Futhark/Optimise/Simplify/Rules.hs
@@ -836,6 +836,15 @@ simplifyConcat _ _ _  _ = Skip
 
 ruleIf :: BinderOps lore => TopDownRuleIf lore
 
+ruleIf vtable pat (StmAux (Certificates uncs) _) (Var v, tb, _, _)
+  | Just _ <- find (\c -> case ST.lookupBasicOp c vtable of
+                       Just (Assert (Var v2) _ _, _) -> v == v2
+                       _ -> False) uncs = Simplify $ do
+  let ses = bodyResult tb
+  addStms $ bodyStms tb
+  sequence_ [ letBindNames_ [patElemName p] $ BasicOp $ SubExp se
+            | (p, se) <- zip (patternElements pat) ses ]
+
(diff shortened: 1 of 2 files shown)


athas:
I don't think this is the right way to go about it.  The right way is to do something like the [ResolveAssertions](https://github.com/diku-dk/futhark/blob/master/src/Futhark/Pass/ResolveAssertions.hs) pass, which feeds information into the algebraic simplifier.  It's too fragile to only look exactly at the assertions.

I think a good starting point is to extend [SymbolTable](https://github.com/diku-dk/futhark/blob/master/src/Futhark/Analysis/SymbolTable.hs) (which is used by the simplifier) to keep track of which things have been "asserted as true" in the current context, and then use that to construct range information that we pass into the simplifier (but not in *every* simplification pass, because this stuff is way too expensive - just extend ResolveAssertions).

athas:
Note that despite its name, the ResolveAssertions pass technically does not know about assertions.  It could be more properly named the "pass boolean expressions to the algebraic simplifier" pass, but the most important of those are definitely assertions.

Munksgaard:
Gotcha, I'll give it another try over the weekend.

athas:
Do we have any programs that we think might benefit from this?  The issue is five years old and I'm not sure it was motivated by an actual use case.

Munksgaard:
Not that I know of, I was just looking through the old issues trying to find a "simple" one to solve :-)

PR #935, opened 2020-04-24
Experiment to see if we can fix failing windows build on azure pipeline

Munksgaard:


athas:
Futhark requires at least GHC 8.8, so this will not fly.

athas:
Looks like the boulder stayed up this time.  Ready to merge?

Munksgaard:
I think so!

flip111:
Does this mean futhark works on windows again ?

athas:
Yes.

PR #939, opened 2020-04-26
Refactor memory representation.

athas:
ExplicitMemory is gone.  Instead we have two successor representations,
SeqMem and KernelsMem, built on the same foundation.

### DIFF
--- a/futhark.cabal
+++ b/futhark.cabal
@@ -157,6 +157,8 @@ library
       Futhark.Pass
       Futhark.Pass.ExpandAllocations
       Futhark.Pass.ExplicitAllocations
+      Futhark.Pass.ExplicitAllocations.Kernels
+      Futhark.Pass.ExplicitAllocations.Seq
       Futhark.Pass.ExtractKernels
       Futhark.Pass.ExtractKernels.BlockedKernel
       Futhark.Pass.ExtractKernels.DistributeNests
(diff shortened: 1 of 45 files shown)


Munksgaard:
What's the purpose of the two new memory representations?

athas:
Solely so that we can have different representations that each have a notion of memory.  Currently, memory implies GPU-style kernel operations.  With this change, we can also have a dedicated representation with memory annotations for the multicore backend.  The actual memory information is identical; the only thing that differs is the Op.

Munksgaard:
> Solely so that we can have different representations that each have a notion of memory. Currently, memory implies GPU-style kernel operations. With this change, we can also have a dedicated representation with memory annotations for the multicore backend. The actual memory information is identical; the only thing that differs is the Op.

Ah, so `SeqMem` is used whenever we we're trying to compile something for the multicore backend, while `KernelsMem` is used for the GPU backend?

athas:
Well, presumably the multicore backend will have an entirely third representation (MulticoreMem? Not sure).  SeqMem is just the proof-of-concept that it's possible and not actually all that hard.

Munksgaard:
> Well, presumably the multicore backend will have an entirely third representation (MulticoreMem? Not sure). SeqMem is just the proof-of-concept that it's possible and not actually all that hard.

Got it.

PR #943, opened 2020-04-28
Assign a dedicated IR to multicore

athas:
**You probably don't want to merge this yet.**

This PR introduces a distinct intermediate representation for multicore parallelism.  The change in code generation is not great yet, but it will allow us to add various extra information, and do different transformations, compared to the GPU backend.  The main visible change from your point of view is that parallelism may now be *nested*.  That is, the body of a SegOp may contain other SegOps.  This doesn't work with the current scheduler, which is why I think we should probably wait a bit before merging this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+  * Fixed some errors regarding constants (#941).
+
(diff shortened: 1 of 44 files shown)


athas:
If you're brave, you can merge this now @HnimNart.  Most things still will not work, especially because of the missing scheduler, but also the still-primitive MulticoreExtractor pass.  I will fix the latter eventually (meaning within a few days, probably).

PR #975, opened 2020-05-16
Add github actions workflow to run tests and builds

Munksgaard:
This commit sets up Github Actions to build and test pull requests and pushes to
the master branch.

Eventually, we can expand the workflow to include pushing new releases and
nightlies, but let's stick with this for now.

I believe the current functionality matches that of our travis setup, plus it build on Mac, Windows and using the different build systems (cabal, stack, nix).

 - [x] Do style checks, integration tests, unit tests, lib tests using nix
 - [ ] ~~Produce statically linked binary tarballs on every commit and release~~ Let's merge this PR. We can always expand the functionality later
 - [x] Test that we can build with cabal
 - [x] Test that we can build with stack
 - [x] Test that we can build on windows
 - [x] Test that we can build on mac
 - [ ] ~~Run benchmarks (and integration tests?) on a GPU machine~~ We need to think a little bit more carefully about how to approach this. We don't want to potentially allow arbitrary code execution on our GPU servers.


athas:
Really cool!  As I see it, there are several distinct things we need our CI to do:

1) Do style checks, run the integration tests, unit tests, lib tests, etc.

2) Produce statically linked binary tarballs on every commit (or every night; doesn't matter much) and release.

3) Test that we can build with `cabal`.

4) Test that we can build with `stack`.

5) Test that we can build on Windows.

Currently (1) is split between Travis, Buildbot, and Azure (although I think style checks are all that Azure doesn't do), and (4) is only done on Travis.

Munksgaard:
Got it. I've also tried to add a build for mac, since I don't see us doing that anywhere currently.

athas:
I would prefer if the (1) step, which is the most tooling-intensive, is built with Nix (which is the direction you seem to be going in).  I prefer the macOS build to be with `cabal` (as you've done it), because Homebrew also builds with `cabal`.

athas:
I suggest that for the "big" test (where we run hlint, the pkgtests, etc), you just use `nix-shell` to run some shell script.  Then all the setup and dependencies can be done in `shell.nix` (which will then also be able to reproduce a development environment nicely).

Munksgaard:
> I suggest that for the "big" test (where we run hlint, the pkgtests, etc), you just use `nix-shell` to run some shell script. Then all the setup and dependencies can be done in `shell.nix` (which will then also be able to reproduce a development environment nicely).

Good idea

Munksgaard:
Should I remove the travis CI files as well?

athas:
Yes.

PR #978, opened 2020-05-18
Fix typo in the language reference

mb64:
The reference includes `loop pat = initial = while cond do loopbody`.

I believe the second `=` should not be there.


athas:
Indeed!  It's been many a year since that was changed.

PR #980, opened 2020-05-19
Warn on out-of-bounds literals

### LINKED ISSUE #979: Warn on out-of-bounds literals
Currently, the Futhark compiler accepts, without any messages, code like

```futhark
let positive_number : i32 = 6789012345
let negative_number : u32 = -12 
let finite_number : f32 = 1e40

entry main : (i32, u32, f32) = (positive_number, negative_number, finite_number)

-- Output:
-- -1800922247i32
-- 4294967284u32
-- f32.inf
```

I tried implementing warnings myself, and ended up with 
```haskell
literalOverflowCheck :: ASTMappable x => x -> TypeM ()
```
which I added to `checkFunDef` and `checkOneExp`, but it didn't seem to actually run for my simple test-case.

If I could get some guidance as to the best way to add, I would love to make a PR.

mb64:
Resolves #979 

Areas where this PR falls short:
 * It adds an extra traversal.
 * I would have liked to include a test that it didn't warn for e.g. `-128 : i8` (interesting since the sub-expression of +128 is out of bounds).  I don't think this is possible with the current test setup.
 * For the program `let main : f64 = 1e400`, it warns `Literal Infinity out of bounds`.  This isn't fixable without changes to the AST/parser.
 * It doesn't warn for type-postfixed literals like `100000u8`, since these are truncated by the parser.  This isn't fixable without changes to the AST/parser.

### DIFF
--- a/src/Language/Futhark/TypeChecker/Terms.hs
+++ b/src/Language/Futhark/TypeChecker/Terms.hs
@@ -41,6 +41,7 @@ import Language.Futhark.TypeChecker.Types hiding (checkTypeDecl)
 import Language.Futhark.TypeChecker.Unify hiding (Usage)
 import qualified Language.Futhark.TypeChecker.Types as Types
 import qualified Language.Futhark.TypeChecker.Monad as TypeM
+import Futhark.Representation.Primitive (intByteSize)
 import Futhark.Util.Pretty hiding (space, bool, group)
 
 --- Uniqueness
@@ -2442,7 +2443,7 @@ causalityCheck binding_body = do
 --
 -- Note: currently unable to detect float underflow (such as 1e-400 -> 0)
 literalOverflowCheck :: Exp -> TermTypeM ()
-literalOverflowCheck = (() <$) . check
+literalOverflowCheck = void . check
   where check e@(IntLit x ty loc) = e <$ case ty of
           Info (Scalar (Prim t)) -> warnBounds (inBoundsI x t) x t loc
           _ -> error "Inferred type of int literal is not a number"
@@ -2453,11 +2454,7 @@ literalOverflowCheck = (() <$) . check
           Info (Scalar (Prim t)) -> warnBounds (inBoundsI (-x) t) (-x) t (loc1 <> loc2)
           _ -> error "Inferred type of int literal is not a number"
         check e = astMap identityMapper{mapOnExp = check} e
-        bitWidth :: IntType -> Int
-        bitWidth Int8 = 8
-        bitWidth Int16 = 16
-        bitWidth Int32 = 32
-        bitWidth Int64 = 64
+        bitWidth ty = 8 * intByteSize ty :: Int
         inBoundsI x (Signed t) = x >= -2^(bitWidth t - 1) && x < 2^(bitWidth t - 1)
         inBoundsI x (Unsigned t) = x >= 0 && x < 2^bitWidth t
         inBoundsI x (FloatType Float32) = not $ isInfinite (fromIntegral x :: Float)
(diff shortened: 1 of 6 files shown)


src/Language/Futhark/TypeChecker/Terms.hs
```diff
@@ -2436,6 +2437,38 @@ causalityCheck binding_body = do
...
+          Info (Scalar (Prim t)) -> warnBounds (inBoundsI x t) x t loc
+          _ -> error "Inferred type of int literal is not a number"
+        check e@(FloatLit x ty loc) = e <$ case ty of
+          Info (Scalar (Prim (FloatType t))) -> warnBounds (inBoundsF x t) x t loc
+          _ -> error "Inferred type of float literal is not a float"
+        check e@(Negate (IntLit x ty loc1) loc2) = e <$ case ty of
+          Info (Scalar (Prim t)) -> warnBounds (inBoundsI (-x) t) (-x) t (loc1 <> loc2)
+          _ -> error "Inferred type of int literal is not a number"
+        check e = astMap identityMapper{mapOnExp = check} e
+        bitWidth :: IntType -> Int
```
athas, line 2456: Import `intByteSize` from `Futhark.Representation.Primitive` and multiply by eight.

src/Language/Futhark/TypeChecker/Terms.hs
```diff
@@ -2436,6 +2437,38 @@ causalityCheck binding_body = do
...
+        bitWidth Int32 = 32
+        bitWidth Int64 = 64
+        inBoundsI x (Signed t) = x >= -2^(bitWidth t - 1) && x < 2^(bitWidth t - 1)
+        inBoundsI x (Unsigned t) = x >= 0 && x < 2^bitWidth t
+        inBoundsI x (FloatType Float32) = not $ isInfinite (fromIntegral x :: Float)
+        inBoundsI x (FloatType Float64) = not $ isInfinite (fromIntegral x :: Double)
+        inBoundsI _ Bool = error "Inferred type of int literal is not a number"
+        inBoundsF x Float32 = not $ isInfinite (realToFrac x :: Float)
+        inBoundsF x Float64 = not $ isInfinite x
+        warnBounds inBounds x ty loc = if inBounds
```
athas, line 2468: Use `unless`.

src/Language/Futhark/TypeChecker/Terms.hs
```diff
@@ -2436,6 +2437,38 @@ causalityCheck binding_body = do
...
+          then pure ()
+          else warn loc $ "Literal " <> show x <> " out of bounds for inferred type " <> pretty ty
```
athas, line 2470: End error messages with a dot.

src/Language/Futhark/TypeChecker/Terms.hs
```diff
@@ -2436,6 +2437,38 @@ causalityCheck binding_body = do
           align (textwrap "Bind the expression producing" <+> pquote (pprName d) <+>
                  "with 'let' beforehand.")
 
+-- | Traverse the expression, emitting warnings if any of the literals overflow
+-- their inferred types
+--
+-- Note: currently unable to detect float underflow (such as 1e-400 -> 0)
+literalOverflowCheck :: Exp -> TermTypeM ()
+literalOverflowCheck = (() <$) . check
```
athas, line 2445: Use `void`.

athas:
I left some comments.  I'm OK with the extra pass for now - there are already lots of other much slower things going on.  Your warning test is correct, although I don't remember if the regex library we use supports brace notation.  I do think you should also have tests that check the boundaries (i.e. that for `i8` we do not warn about -128 or 127, but do warn about -129 and 128).

athas:
Also, don't mind the Windows build failures.  We're having some problems with GHC getting wedged for some reason.

mb64:
I did your requested changes, and expanded the tests to cover edge cases where it shouldn't warn.  It seems that the regex library does indeed support brace notation.

Is there a concrete code format policy for furthark?  I did my best to imitate what seemed to be the convention of the other code, but Haskell is particularly hard to format consistently.

athas:
There is the [style guide](https://github.com/diku-dk/futhark/blob/master/STYLE.md), but it leaves a lot of leeway.  Until we have fully automated formatting, I'm OK as long as the code looks fairly similar to its surroundings.

athas:
This looks good to me.  Could you also add an entry to `CHANGELOG.md` in the "Added" section?  Then I'll merge once the CI is done.

athas:
Great, thanks!

athas:
Thinking about it, is there a good reason we don't make this an actual error?

athas:
The next release will break backwards compatibility in other minor ways anyway, and I like to roll multiple small-but-technically-breaking-changes into the same release.

PR #981, opened 2020-05-19
Add an attribute system to Futhark.

### LINKED ISSUE #801: Add attribute system
I'd like to soon turn on incremental flattening by default.  One problem is that compile times increase significantly - partly because multiple versions are necessary, but also partly because we generate versions that *we* know will never be useful, but the compiler does not know.  For example, LocVolCalib has this `main`:

```
let main (outer_loop_count: i32) (numX: i32) (numY: i32) (numT: i32)
         (s0: f32) (t: f32) (alpha: f32) (nu: f32) (beta: f32): []f32 =
  ...
```

*We* know that `outer_loop_count` will never be sufficiently parallel on its own, but the compiler does not, so it generates a version of the code that is a big `map` with a huge sequential computation inside.  It would be nice if we could hint this to the compiler, and so not have it waste time generating (and subsequently optimising) that version.  I think this should be done by extending the notation for entry points.  Here's a draft of what it could look like:

```
let main (outer_loop_count: i32) (numX: i32) (numY: i32) (numT: i32)
         (s0: f32) (t: f32) (alpha: f32) (nu: f32) (beta: f32): []f32
  with insufficient_parallelism outer_loop_count =
  ...
```

The `with insufficient_parallelism outer_loop_count` is not an ordinary expression, but an annotation using some grammar for hints.  Here we assert that the size `outer_loop_count` never provides enough parallelism on its own.  For a matrix multiplication where we know the outer sizes are always sufficient, we could put:

```
let main [n][m][l] (x: [n][l]i32) (y: [l][m]i32): [n][m]i32
  with sufficient_parallelism (n*m) =
  ...
```

While I don't intend to use this notation for anything except multiversioning hints in the short term, in the long term it might also be the place we put invariants for refinement types or Cosmin's old ideas about shape annotations.

I'm willing to bike shed the syntax.  In the draft above, I made these hints a first-class syntactical construct (that occurs before the `=` in an entry point definition).  Other languages might hide them in comments, but I've never liked that much.

athas:
Closes #801.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+
   * Warnings for overflowing literals, such as `1000 : u8`.
+
+  * Futhark now supports an attribute system, whereby expressions can
+    be tagged with attributes that provide hints or directions to the
+    compiler.  This is an expert-level feature, but it is sometimes
+    useful.
 
 ### Removed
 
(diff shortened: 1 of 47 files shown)


athas:
One thing that sucks is that now the `Futhark.Representation.AST.Attributes` module hierarchy is confusingly named, since it's not about this kind of attributes at all.  What should they be renamed to?  `Props`?

We should also shorten `Representation` to just `Rep`.

PR #991, opened 2020-05-27
Split tests

Munksgaard:


athas:
I've realised this doesn't scale when we want to make the tests more complete (e.g. run with all backends).  It'll just be a large number of shell scripts that are 90% boilerplate for one line of actual work.  Better to just make it all explicit in the CI YAML.

PR #993, opened 2020-05-28
Remove C# backends.

### LINKED ISSUE #984: Removing the C# backend
Nobody uses the C# backend and it is a drain on scarce maintenance resources.  Would anyone miss it were it to go?

Note that it is essentially impossible to use as a library in its current state, because it is so hard to get data in and out of it (#618 and #619 are examples of this).

athas:
Closes #984.


Munksgaard:
Perhaps there should be a link to https://github.com/diku-dk/futhark/issues/984 in the changelog?

PR #1001, opened 2020-06-04
Don't output tty control codes if the terminal is dumb

### LINKED ISSUE #1000: Executing `futhark test` in a dumb terminal produces a lot of unnecessary output
For instance, running `futhark test` through an eshell inside emacs will output something like this:

```
│ programs │ 1392   │ 0      │ 3/1395    │
│ runs     │ 2353   │ 0      │ 139/2492  │
...
│ programs │ 1393   │ 0      │ 2/1395    │
│ runs     │ 2392   │ 0      │ 100/2492  │
...
│ programs │ 1394   │ 0      │ 1/1395    │
│ runs     │ 2428   │ 0      │ 64/2492   │
...
│ programs │ 1395   │ 0      │ 0/1395    │
│ runs     │ 2492   │ 0      │ 0/2492    │
...
```

Which goes on and on for as many times as there are tests. Perhaps we should detect if the output terminal is dumb (edit: meaning `TERM=dumb`), in which case we should handle the output like with `--notty`.

Munksgaard:
Fixes #1000

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   * Fixed a typer checker bug for tracking the aliases of closures
     (#995).
 
+  * Fixed handling of dumb terminals in futhark test (#1000).
+
 ## [0.15.8]
 
 ### Added
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -15,7 +15,7 @@ export GIT_SSL_CAINFO=/etc/ssl/certs/ca-certificates.crt
 cabal install --install-method=copy --overwrite-policy=always --installdir=$HOME/.cabal/bin
 
 echo "# Integration tests"
-futhark test --no-terminal tests examples --no-tuning
+futhark test tests examples --no-tuning
 
 echo "# Library tests"
 make -C libtests/c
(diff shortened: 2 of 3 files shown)


athas:
Is `TERM=dumb` set by anything else than Emacs?

Munksgaard:
I'm pretty sure `dumb` is a valid predefined TERM variable, it's even mentioned in the TERM(7) manual: https://linux.die.net/man/7/term. It may be used in other contexts, though I haven't run in to any.

That said, if your question is "is this really worth supporting, if the only plausible use case is someone using eshell to run futhark test?", I don't know. If your question is "is there perhaps a better way to do this?", I'm sure there is. A more general solution would be to explicitly check the term capabilities to see if moving the cursor is actually supported.

athas:
This is worth supporting, it just doesn't address the issue of CI services claiming to be terminals, yet not supporting common control codes.

Munksgaard:
> This is worth supporting, it just doesn't address the issue of CI services claiming to be terminals, yet not supporting common control codes.

Does that happen? Where?

Munksgaard:
Ah, you're talking about having to explicitly pass `--no-terminal` in the CI?

Munksgaard:
It actually looks like Github Actions uses the "dumb" terminal as well: 
https://github.com/diku-dk/futhark/pull/1001/checks?check_run_id=737857033#step:6:8

Munksgaard:
Looks like it works without `--no-terminal` now: https://github.com/diku-dk/futhark/pull/1001/checks?check_run_id=737877331#step:8:10

PR #1003, opened 2020-06-05
Existentialize loop returns in order to eliminate some unnecessary copies

Munksgaard:
The approach here is to always existentialize or generalize as much of the array
type and index function as possible. Therefore, the technique is a bit different
from the technique used to avoid copies in `if` branches, where we explicitly
try to do the least amount of generalization possible. The approach taken here
is simpler, but leads to more existentials. We therefore rely on the simplifier
to remove some of the invariant loop returns.

For this to work, we need to relax the type checker a bit. We also fixed a bug
in the imperative code generator that didn't handle ScalarSpace correctly.

Here's the resulting benchmark differences from master:

```

futhark-benchmarks/accelerate/canny/canny.fut
  data/lena512.in:                                                      0.99x
  data/lena256.in:                                                      0.96x

futhark-benchmarks/accelerate/crystal/crystal.fut
  #0 ("200i32 30.0f32 5i32 1i32 1.0f32"):                               0.98x
  #4 ("2000i32 30.0f32 50i32 1i32 1.0f32"):                             1.00x
  #5 ("4000i32 30.0f32 50i32 1i32 1.0f32"):                             1.00x

futhark-benchmarks/accelerate/fft/fft.fut
  data/64x256.in:                                                       0.93x
  data/128x512.in:                                                      0.97x
  data/1024x1024.in:                                                    0.96x
  data/512x512.in:                                                      0.97x
  data/256x256.in:                                                      0.98x
  data/128x128.in:                                                      0.95x

futhark-benchmarks/accelerate/fluid/fluid.fut
  benchmarking/medium.in:                                               0.99x

futhark-benchmarks/accelerate/hashcat/hashcat.fut
  rockyou.dataset:                                                      1.00x

futhark-benchmarks/accelerate/kmeans/kmeans.fut
  data/k5_n50000.in:                                                    1.00x
  data/trivial.in:                                                      1.01x
  data/k5_n200000.in:                                                   1.00x

futhark-benchmarks/accelerate/mandelbrot/mandelbrot.fut
  #1 ("1000i32 1000i32 -0.7f32 0.0f32 3.067f32 100i32 16...."):         1.01x
  #3 ("4000i32 4000i32 -0.7f32 0.0f32 3.067f32 100i32 16...."):         1.00x
  #2 ("2000i32 2000i32 -0.7f32 0.0f32 3.067f32 100i32 16...."):         1.00x
[...]

PR #1006, opened 2020-06-06
Let the allocator use memory blocks with right size but wrong tag.

athas:
Worst case, this means we push a fresh allocation to later rather than
now, so it should not have any negative consequences.

That said, our free list logic is still fairly crude, and there is
likely room for further refinement.

With this change, the performance impact of #1003 drops to just the extra copies.


athas:
I've noticed that this change has significant impact on the performance of many other programs.  Mostly positive (some significantly so), but also some negative.  I will still merge this, but I think more work on the dynamic allocator is warranted.

PR #1010, opened 2020-06-08
Perform incremental flattening by default.

athas:
Moderate flattening is completely gone - instead, you can use
attributes to control parallelisation, if needed.

I expect this will have problems and require lots of tweaking, but it
is time we address it directly.

Compile times are harmed, of course.  I hope that is offset by the
significant improvements we've made to compile times recently (and
besides, most of the time is currently spent before kernel extraction,
during inlining, which this change does not affect).

This commit also makes minor changes to the heuristics we use to prune
the multi-versioning tree.  Nothing major, but I expect it will also
require more tuning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+  * Incremental flattening is now performed by default.  Use
+    attributes to constrain and direct the flattening if you have
+    exotic needs.  This will likely need further iteration and
+    refinement.
+
 ### Fixed
 
   * Fix bug in slice simplification (#992).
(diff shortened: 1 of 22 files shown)


athas:
The [slowest compiling Futhark program I know of](https://github.com/filecoin-project/neptune-triton) goes from 1m57s to 3m29s with this change.  That's a lot more than most of our benchmarks.  I still think it would be nice to shrink it down (and I have some concrete ideas), but I don't think it's a blocker for merging thus.

athas:
The versions generated are mostly baloney, and tend to involve an intra-group parallel version where the innermost parallelism is of size 9.  This constant is perfectly visible to the compiler, so maybe we should have an ad-hoc rule not to generate intra-group versions if the inner parallelism is known to be below some threshold (say, 32).

PR #1013, opened 2020-06-09
Add some allocation debugging information

Munksgaard:


rts/c/opencl.h
```diff
@@ -848,8 +848,14 @@ static int opencl_alloc(struct opencl_context *ctx, size_t min_size, const char
     // expose OpenCL pointer values directly to the application, but
     // instead rely on a level of indirection.
     if (size >= min_size) {
+      if (ctx->cfg.debugging)
+        fprintf(stderr, "No need to allocate: Found a block in the free list.\n");
```
athas, line 852: Braces are mandatory.  I guess I should add that to the [style guide](https://github.com/diku-dk/futhark/blob/master/rts/c/STYLE.md).

Munksgaard: I've pushed an update

PR #1027, opened 2020-06-19
Clean up lexer and fix bug

mb64:
This PR cleans up the lexer, making it shorter, more correct, and IMO more readable.

Previously, the lexer failed to lex hexadecimal float literals like `0xabc.def321p0f32` that were type suffixed, but also contain the hex digit sequence `f32` or `f64` beforehand.  This is fixed.

Edit: `happy` does not seem to be very happy in the CI... not sure why

### DIFF
--- a/src/Language/Futhark/Parser/Lexer.x
+++ b/src/Language/Futhark/Parser/Lexer.x
@@ -14,13 +14,14 @@ module Language.Futhark.Parser.Lexer
 import qualified Data.ByteString.Lazy as BS
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
-import Data.Char (ord, toLower)
+import Data.Char (ord, toLower, digitToInt)
 import Data.Int (Int8, Int16, Int32, Int64)
 import Data.Word (Word8)
 import Data.Bits
 import Data.Function (fix)
 import Data.List
 import Data.Monoid
+import Numeric
 
 import Language.Futhark.Core (Int8, Int16, Int32, Int64,
                               Word8, Word16, Word32, Word64,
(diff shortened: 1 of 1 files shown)


athas:
Nice catch!

Ignore the `test-unit-and-style` check (it seems to be flaky today), but something is now going wrong with the `tests/hexfloats.fut` test.

src/Language/Futhark/Parser/Lexer.x
```diff
@@ -51,12 +53,12 @@ import Futhark.Util.Loc hiding (L)
...
 @qualunop = (@identifier ".")+ @unop
 
-@opchar = ("+"|"-"|"*"|"/"|"%"|"="|"!"|">"|"<"|"|"|"&"|"^"|".")
-@binop = ("+"|"-"|"*"|"/"|"%"|"="|"!"|">"|"<"|"|"|"&"|"^") @opchar*
+$opchar = [\+\-\*\/\%\=\!\>\<\|\&\^\.]
+@binop = ($opchar # \.) $opchar*
 @qualbinop = (@identifier ".")+ @binop
 
 @space = [\ \t\f\v]
-@doc = "-- |"[^\n]*(\n@space*"--"[^\n]*)*
```
athas, line 59: Can you explain why this change is right?  Won't this match the following?

```
-- | foo bar
baz
```

Or does `.` not match newlines in Alex?

mb64: Yes – in Alex (like many regexes) `.` is everything but newlines.

[From the docs:](https://www.haskell.org/alex/doc/html/charsets.html)
"The built-in set ‘.’ matches all characters except newline (\n)."

athas:
CI says you have a syntax error in one of the lexer code fragments.

mb64:
Huh, odd.  I thought I had tested it on my machine beforehand, but it was clearly wrong. (`let ... in in ...`)

Should be fixed now

athas:
Thanks for the fix!

PR #1034, opened 2020-06-24
Add warn(safety_checks) attribute for warning on bounds checks.

### LINKED ISSUE #1023: An attribute for warning on safety checks
We need an attribute that you can put on an expression, which will make the compiler issue a warning if it is unable to remove all safety checks inside it.  What's a good name for it?  `#[warn_on_checks]`?

athas:
Closes #1023.

### DIFF
--- a/docs/language-reference.rst
+++ b/docs/language-reference.rst
@@ -1664,6 +1664,14 @@ Exploit only outer parallelism in the attributed SOAC.
 Do not perform any dynamic safety checks (such as bound checks) during
 execution of the attributed expression.
 
+``warn_on(safety_checks)``
+..........................
+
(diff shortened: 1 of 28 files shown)


docs/language-reference.rst
```diff
@@ -1664,6 +1664,14 @@ Exploit only outer parallelism in the attributed SOAC.
 Do not perform any dynamic safety checks (such as bound checks) during
 execution of the attributed expression.
 
+``warn_on(safety_checks)``
```
Munksgaard, line 1667: This should be `warn(safety_checks)`, right?

athas: Yes.

PR #1049, opened 2020-07-06
Remember to also permute the type here.

athas:


Munksgaard:
What problems did this cause? Is there a test-case?

athas:
I am too lazy to extract a test case, but the test suite for `cpprandom` fails when compiled with the OpenCL backend.

PR #1052, opened 2020-07-11
No need for treat SubExp specially in traversals.

athas:


athas:
Oh, right, we need to special-case subexpressions for copy-propagation.

PR #1064, opened 2020-07-29
Use inner parallelism when for intra-group threshold comparisons.

athas:
Makes the default threshold generated for intra-group much smaller (32).

This is based on the principle that intra-group is always worth using
(at least by default), as long as the groups are not too tiny.


athas:
@fzzle, this will speed up your code.

Issue #1051, opened 2020-07-10
A design for accumulators in the core IR


athas:
We have recently realised the need to express histogram-like
computations where each input element can give rise to an arbitrary
number of updates.  #737 lists examples (but that can be worked around
with padding or `expand`), and our recent AD research revealed more
cases that cannot simply be worked around.  Therefore, I think we
should start designing and implementing the **core IR** representation
of such a feature.  Let me repeat: this issue is **only concerned with
the IR**, *not* the source language.  This means we do not need
guaranteed soundness, but we need to have a design and rules
(preferably checkable) that make it easy for the compiler to do the
right thing.  How we expose this to the source language (if at all) is
for later.

Since the core IR does not have a well-defined notation for type
rules, I am going to approximate some of the rules through source
language syntax.  **This is still only about IR constructs.**

The basic idea that we have come up with is to introduce a type of
*accumulators*, by which we can *write* values, but not read.
Accumulators are constructed from arrays and can also be turned back
into arrays:

```
type acc [n] 't
val mkacc [n] 't : (t -> t -> t) -> *[n]t -> *(acc [n] t)
val upd [n] 't : *(acc [n] t) -> i32 -> t -> *(acc [n] t)
val unacc [n] 't : *(acc [n] t) -> [n]t
```

When you convert an array to an accumulator with `mkacc`, you also
provide an associative and commutative binary operator, which is then
used whenever `upd` is used to update a value at a given index.  This
means that the order of `upd`s does not matter.  Multiple threads can
call `upd` on the same `acc` concurrently, without risk.  However,
*how* do we write parallel updates of a single `acc`?  We could try
writing something like this:

```
let xs_acc = mkacc (+) xs
let ys = map (\i -> upd xs_acc i 1) is
```

But this is not OK.  The `upd` function consumes its result, to
enforce linear usage of the accumulator, and all values become
non-unique inside of a lambda.  We need this instead:

```
let xs_acc = mkacc (+) xs
let xs_accs = replicate n xs_acc
let ys = map (\i -> upd xs_acc i 1) is xs_accs
```

But we still have a problem - that `ys` is an array of *accumulators*.
Each of those can be turned into an array with `unacc`, but that is
not really what we want.  We need a new primitive for collapsing
accumulators:

```
val join_accs [n] t : *[](acc [n] t) -> *(acc [n] t)
```

Alternatively, `unacc` should accept an array of accumulators instead.
Maybe `mkacc` should also produce an array, since accumulators are
[...]

PR #1071, opened 2020-08-02
No anydim

athas:


athas:
This is a yak-shaving exercise that is ultimately oriented towards #1051.

PR #1075, opened 2020-08-05
More descriptive error messages when dtypes are wrong for the Python backend

philass:
This fix is for improved error message handling for the python library when futhark is compiled into a python library. Specifically for a more descriptive error message when dtypes do not match.

An example futhark program
```futhark
entry select(db : [][]i32) (col : i32) : []i32 =  db[:, col]
```

And corresponding library call from a python program 

```python
import numpy as np
from db import db

data = np.array([[1, 2, 3, 4]])
instance = db()
result =  instance.select(data, 3)
print(result)
```
Then compiled and run

```
futhark python --library db.fut && python3 fut_call_example.py

Traceback (most recent call last):
  File "fut_call_example.py", line 6, in <module>
    result =  instance.select(data, 3)
  File "/Users/philiplassen/CS/FutDB/db.py", line 1365, in select
    db_mem_36_ext))
TypeError: Argument #0 has invalid value
Futhark type: [][]i32
Argument has Python type <class 'numpy.ndarray'> and value: [[1 2 3 4]]
```

Nothing in the error message actually points the user to what is wrong. As both the value and python type that the compiler says conflict with the futhark type. The issue is of course that the dtype of the Python array is wrong. The program was expecting a dtype of i32. But numpy defaults to i64. This is an easy mistake to make as indices are generally i32. And we don't really think about the numpy default dtypes. 

My code changes give the following more descriptive error message.

```
Traceback (most recent call last):
  File "fut_call_example.py", line 6, in <module>
    result =  instance.select(data, 3)
  File "/Users/philiplassen/CS/FutDB/db.py", line 1371, in select
    db_mem_36_ext.dtype))
TypeError: Argument #0 has invalid value
Futhark type: [][]i32
Argument has Python type <class 'numpy.ndarray'> and value: [[1 2 3 4]]
Expected array with elements of dtype: int32
The array given has elements of dtype: int64
```
This very clearly indicates what the issue is, and why the error is being thrown.

### DIFF
--- a/src/Futhark/CodeGen/Backends/GenericPython.hs
+++ b/src/Futhark/CodeGen/Backends/GenericPython.hs
@@ -436,6 +436,16 @@ badInput i e t =
                           , "Futhark type: {}"
                           , "Argument has Python type {} and value: {}"]
 
+badInputType :: Int -> PyExp -> String -> PyExp -> PyExp -> PyStmt
+badInputType i e t de dg =
+  Raise $ simpleCall "TypeError"
+  [Call (Field (String err_msg) "format")
+   [Arg (String t), Arg $ simpleCall "type" [e], Arg e, Arg de, Arg dg]]
+  where err_msg = unlines [ "Argument #" ++ show i ++ " has invalid value"
+                          , "Futhark type: {}"
+                          , "Argument has Python type {} and value: {}"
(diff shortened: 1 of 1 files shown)


athas:
Nice improvement, thanks!

PR #1078, opened 2020-08-10
Add sexp-based representation for Futhark IR

Munksgaard:
The purpose of this PR is to allow users to inspect and modify the IR of a program using `futhark dev`. 

We introduce a number of new features to `futhark dev`:

 - Writing the output result as an sexp using the `--sexp` flag.
 - When reading files, `futhark dev` will distinguish between files ending in `.fut` and files ending in `.sexp`. If a file ends in `.sexp`, `futhark dev` will use the sexp based decoder to decode the IR in the file.
 - Added `--compile-opencl` and `--compile-c` flags, to allow users to compile modified IR using the sexp based notation.

There is plenty of room for improvement in the representation that I've made, and sexp-grammar should allow for a much nicer language than what we have now, but I wanted to get this merged so I can actually start using it in my interference tests.

There also isn't a whole lot of testing in place to actually make sure the IR representation is actually bidirectional. However, `futhark dev --sexp` will make sure that any encoded program is also decodeable, and that the decoded program is identical to the original program.

### DIFF
--- a/src/Futhark/Actions.hs
+++ b/src/Futhark/Actions.hs
@@ -9,14 +9,17 @@ module Futhark.Actions
   , compileCAction
   , compileOpenCLAction
   , compileCUDAAction
+  , sexpAction
   )
 where
 
+import Language.SexpGrammar as Sexp
 import Control.Monad
 import Control.Monad.IO.Class
 import System.Exit
 import System.FilePath
 import qualified System.Info
+import qualified Data.ByteString.Lazy.Char8 as ByteString
 
 import Futhark.Compiler.CLI
 import Futhark.Analysis.Alias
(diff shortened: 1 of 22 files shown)


athas:
I think it is best to factor our some of the orthogonal changes, e.g. the new actions.  I will do that.  Also, for how long do we expect to depend on an unreleased package?  I would like to avoid that at least for the next release of Futhark.

Munksgaard:
> I think it is best to factor our some of the orthogonal changes, e.g. the new actions. I will do that. 

Good idea.

> Also, for how long do we expect to depend on an unreleased package? I would like to avoid that at least for the next release of Futhark.

Good question, I'll ask them what it would take to release a new version. For the record, these are the two issues I'm waiting on: https://github.com/esmolanka/sexp-grammar/issues/12 and https://github.com/esmolanka/sexp-grammar/issues/11

Munksgaard:
Currently failing because sexp-grammar is marked as broken in nixpkgs: Tracking issue https://github.com/NixOS/nixpkgs/pull/95067

We'll need to update our sources.json whenever that PR is merged.

src/Futhark/Actions.hs
```diff
@@ -75,7 +75,7 @@ sexpAction =
   where
     helper :: ASTLore lore => Prog lore -> IO ()
     helper prog =
-      case encode prog of
+      case encodePretty prog of
```
Munksgaard, line 78: @athas Would you prefer that we use the non-prettyprinted output? It's rather verbose like this.

athas: No, we should solve the verbosity problem directly, not by removing whitespace.

Munksgaard:
It works! Rejoice!

Munksgaard:
@athas I believe this is ready to merge now. As we talked about, we can always go back and improve the representation later.

PR #1090, opened 2020-08-17
Raise Error on dimension mismatch

philass:
This fix catches a possible missed error in the Python backend. The error is when the dimensionality of the array expected by futhark does not match the dimension of the numpy array given as an arguement.  Below is a toy example.

An example futhark program
```futhark
entry select(db : [][]i32) (col : i32) : []i32 =  db[:, col]
```

And corresponding library call from a python program 

```python
import numpy as np
from db import db

data = np.array([1, 2, 3, 4]).astype(np.int32)
instance = db()
result =  instance.select(data, 3)
print(result)
```
Then compiled and run

```
futhark python --library db.fut && python3 fut_call_example.py

Traceback (most recent call last):
  File "fut_call_example.py", line 6, in <module>
    result =  instance.select(data, 3)
  File "/Users/philiplassen/CS/FutDB/db.py", line 1374, in select
    implz2081U_4211 = np.int32(db_mem_4228_ext.shape[1])
IndexError: tuple index out of range
```
I have changed the Futhark code to catch this error. So the new result of running the code snippet above is

```
Traceback (most recent call last):
  File "fut_call_example.py", line 6, in <module>
    result =  instance.select(data, 3)
  File "/Users/philiplassen/CS/FutDB/db.py", line 1374, in select
    raise TypeError("Argument #0 has invalid value\nDimensionality mismatch\nExpected Futhark type: {}\nBad Python value passed\nActual Futhark type: {}\n".format("[][]i32",
TypeError: Argument #0 has invalid value
Dimensionality mismatch
Expected Futhark type: [][]i32
Bad Python value passed
Actual Futhark type: []i32
```

### DIFF
--- a/src/Futhark/CodeGen/Backends/GenericPython.hs
+++ b/src/Futhark/CodeGen/Backends/GenericPython.hs
@@ -447,6 +447,19 @@ badInputType i e t de dg =
                           , "Expected array with elements of dtype: {}"
                           , "The array given has elements of dtype: {}"]
 
+badInputDim :: Int -> PyExp -> String -> Int -> PyStmt
+badInputDim i e typ dimf =
+  Raise $ simpleCall "TypeError"
+  [Call (Field (String err_msg) "format")
+   [Arg eft, Arg aft]]
+  where eft = String (concat (replicate dimf "[]") ++ typ)
+        aft = BinOp "+" (BinOp "*" (String "[]") (Field e "ndim")) (String typ)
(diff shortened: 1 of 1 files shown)


philass:
Im guessing this won't pass the CI, like the last two commits to Futhark. But I don't think its because my code breaks anything.

athas:
No, it is something else that I am fixing.

PR #1093, opened 2020-08-20
change lex file extension

philass:
I totally understand if this is something you don't care about (and don't accept this PR). But its been bugging me for ages that Lexer.x has been classified as Logos language.
Screenshot 2020-08-20 at 20 18 44

Simply renaming `Lexer.x` to `Lexer.l` fixes this issue (Lexer.lex) would also work.

Screenshot 2020-08-20 at 20 07 50

An alternative solution is to add a .gitattributes file and override the classification. 

This doesn't really change anything in the compiler, just how it looks on github (which is of course of the lowest priority).


philass:
Damn. I built with both Cabal and Stack on my Mac with no problems. 

the cabal error is ```suffix: ["gc","chs","hsc","x","y","ly","cpphs","hs","lhs","hsig","lhsig"]```

I guess file extension name does matter for cabal sometimes...

I could instead just do the correct classification by adding a `.gitattributes` or give up on this fruitless endeavour?

athas:
I think `.gitattributes` is the best solution.

philass:
I'm actually going to try and make a change to github's Linguist. If its accepted I think everything should be good. Otherwise I'll do the gitattributes.

philass:
I think the classification should work now. I added a .gitattributes

PR #1098, opened 2020-08-24
Run OpenCL code through oclgrind on CI.

### LINKED ISSUE #1097: Use oclgrind to test OpenCL backend on CI
Because oclgrind will catch memory errors (like valgrind), this will be really good for catching errors.  Sadly, we cannot use oclgrind's race detector, because we do intentionally have some technical race conditions (that we know are always safe on GPUs).

Oclgrind runs pretty slow, so not all datasets for all tests will be usable (and none of our benchmarks will).  Fortunately, we already have those tagged with `no_python`.

athas:
Closes #1097.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -246,7 +246,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get update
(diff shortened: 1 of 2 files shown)


athas:
This PR will fail (even if I did the GA stuff correctly) because oclgrind actually does find errors in our code!  I will have to fix those before we can merge this.

PR #1101, opened 2020-08-25
Multicore merge

athas:
This merges master (as is tradition), and also fixes the bug that made myocyte not validate.

There is still one test program that does not validate, though.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.x linguist-language=Lex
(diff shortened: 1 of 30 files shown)


HnimNart:
Excellent, thank you. 

Was it ```reduce_by_index/large.fut``` that failed?

athas:
No, it's `examples/rosettacode/life.fut` for some reason.  I will investigate.

Issue #42, opened 2015-04-02
Add coding style guidelines


Athas:
Just to have some consistency.  It would be nice if it also contained some Emacs configuration to make following the style more convenient.

RasmusWL:
Would be great... I think I'm currently breaking it :)

Munksgaard:
Was there ever created an Emacs configuration that enforce (a subset of) the coding style guidelines?

athas:
Not that I know of.  Is it even possible?

When the style guide was written, I don't think there were any good Haskell code formatters.  There is [brittany](https://hackage.haskell.org/package/brittany) now and I would certainly support mechanising the style guide (even changing it to accomodate what brittany or some other tool can handle).

Munksgaard:
I'm not that familiar with Haskell formatters and style checkers, but it should be possible to create an emacs configuration that only indents two spaces, highlights (and/or deletes) trailing whitespace and breaks lines over 80 characters. 

My main reason for asking is that my haskell indentation module (haskell-indent) uses 4 characters to indent by default, and I've been too lazy to figure out how to change it.

Munksgaard:
I don't know much about brittany, but if it's anything like rustfmt or gofmt I'm all for it.

Munksgaard:
For the record, this is what Brittany does to src/Futhark/Representation/ExplicitMemory/IndexFunction.hs: https://github.com/Munksgaard/futhark/commit/4728886e16f531802b447a1bb1d6ecbf98f3d20b?diff=split using the standard options on https://hexagoxel.de/brittany/ and version 0.12.1.1. It's opinionated about certain things, like imports, but apart from that I think the changes are not unreasonable.

athas:
I'm not a big fan of everything it does, but I guess I could live with it.  Do people normally use Brittany without options?

Munksgaard:
[...]

PR #1108, opened 2020-08-31
Ormolu formatting

Munksgaard:
The purpose of this PR is to apply a uniform style to all the Haskell code in Futhark. I've chosen ormolu as the tool to do this, because it is simple to use, is opinionated, and (as of 0.1.2.0) it is idempotent for the code currently in Futhark.

More information about ormolu can be found here: https://github.com/tweag/ormolu. Crucially, this page shows how to set up emacs integration: https://github.com/vyorkin/ormolu.el

For more discussion about code formatting and ormolu, take a look at #42.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -50,7 +50,7 @@ This command can be used to format all Haskell files in Futhark, while
 checking that the formatting is idempotent:
 
 ```
-find src unittests -name '*.hs' -print -execdir ormolu --mode inplace --check-idempotence {} +
+./tools/run-formatter.sh src unittests
 ```
 
 The idempotence check is mostly done to make sure Ormolu (which is still a young
--- a/STYLE.md
+++ b/STYLE.md
@@ -18,6 +18,62 @@ guidelines on some of the principles underlying the compiler design.
 Style Rules
 ===========
 
+Ormolu
+------
+
+Futhark uses [Ormolu](https://github.com/tweag/ormolu/) to enforce consistency
+in formatting style. The style is checked by our CI, so you should make sure any
+pull requests comply with Ormolus formatting before trying to merge.
+
+### Installation
+
+Installing Ormolu can be done through Cabal, Stack, Nix or the Arch Linux
+package manager. For instance, to install through Cabal, run the following
+command:
+
+```
+cabal install ormolu
+```
+
+If you're running Nix or NixOS, you can just use `nix-shell` to enter a
+development environment with ormolu already installed.
+
+### Basic Usage
+
+The following command formats a single file:
+
+```
+ormolu -i FILE.hs
+```
+
+This command can be used to format all Haskell files in Futhark, while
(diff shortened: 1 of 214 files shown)


Munksgaard:
Of course, merging this will wreak havoc on all other pull requests and branches branches currently in use, but if we want to do this, we may as well get over that hurdle sooner rather than later.

Munksgaard:
There seems to be an issue with reading non-ascii characters if the locale is not set correctly. As far as I can tell though, the issue is with GHC, not with ormolu as such. More information here: https://github.com/tweag/ormolu/issues/38

athas:
If we do this, then STYLE.md should be replaced with instructions on how to re-run Ormolu.

Munksgaard:
Good point, I'll try to update STYLE.md. However, I don't believe we can completely get rid of STYLE.md, since Ormolu doesn't handle things like rewriting ifs as pattern matches, preferring function pattern matching to `case`, comment best-practices, naming and so on. 

Ormolu also doesn't automatically wrap lines at 80 characters (or any character count, for that matter). It is up the the user to introduce the necessary line breaks, which ormolu will then use as a guide to format the surrounding expressions. For instance, given this long type declaration

```
onKernels :: (SegOp SegLevel KernelsMem -> ReuseAllocsM (SegOp SegLevel KernelsMem)) -> Stms KernelsMem -> ReuseAllocsM (Stms KernelsMem)
```

The user can introduce a single newline

```
onKernels :: (SegOp SegLevel KernelsMem -> ReuseAllocsM (SegOp SegLevel KernelsMem)) -> Stms KernelsMem -> 
  ReuseAllocsM (Stms KernelsMem)
```

and Ormolu will then re-format it like this

```
onKernels ::
  (SegOp SegLevel KernelsMem -> ReuseAllocsM (SegOp SegLevel KernelsMem)) ->
  Stms KernelsMem ->
  ReuseAllocsM (Stms KernelsMem)
```

Munksgaard:
@athas Well, should we?

STYLE.md
```diff
@@ -18,206 +18,150 @@ guidelines on some of the principles underlying the compiler design.
...
-pointless use of alignment, and are retained to serve as a reminder of
-this.
+This command can be used to format all Haskell files in Futhark, while
+checking that the formatting is idempotent:
 
-If in doubt, don't align.  Spurious alignment makes the code feel weird
-and off-key, and it can be remarkably ugly when not maintained.  Alignment
-in expressions is usually a bad idea.
+```
+find src unittests -name '*.hs' -print -execdir ormolu --mode inplace --check-idempotence {} +
```
athas, line 53: Can you put a script in `tools/` that does this, like we have for style checking?

athas:
According to `git blame`, pretty much every bug in the compiler is now your fault.

Munksgaard:
`git blame -w` is better, though not perfect...

PR #1113, opened 2020-09-05
Use cache-oblivious transpose on CPU.

athas:
This is a good bit faster in many cases, but a bit slower for small
arrays.  Maybe we can special case those later.


athas:
Ormolu is really doing a number on this one.

PR #1121, opened 2020-09-09
delete gitattributes

philass:
I made a PR in Linguist that fixes the Logos misclassification of Lex files https://github.com/github/linguist/pull/4952. Version 7.11 was released a few days ago, so its functionality has now taken effect. 

So we no longer need the `.gitattributes` file


Munksgaard:
Good job!

PR #1129, opened 2020-09-14
Multicore merge

athas:
This introduces double-buffering to the multicore pipeline and dramatically speeds up poseidon.

I have not run any other tests or benchmarks.


athas:
OK, all the tests pass, but I still have not run any benchmarks.  I expect a variety of things might get faster.

PR #1131, opened 2020-09-18
Better opencl commands

Munksgaard:
The purpose of this PR is to add some useful extra flags to OpenCL and C-like compiled futhark programs. Specifically, it currently adds support for a `--list-devices` flag on OpenCL executables and a `--help` flag for all C-like executables.

Here is a sample output by `--list-devices`:

```
[munksgaard@church:~/src/futhark]$ ./tests/allocs --list-devices
Platform: Intel Gen OCL Driver
[0]: Intel(R) UHD Graphics Coffee Lake Mobile GT2
Platform: Intel(R) OpenCL
[1]: Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz
```

And for --help on an OpenCL executable:

```
[munksgaard@church:~/src/futhark]$ ./tests/allocs --help
Usage: ./tests/allocs [OPTION]...
Options:

  -t/--write-runtime-to FILE Print the time taken to execute the program to the indicated file, an integral number of microseconds.
  -r/--runs INT              Perform NUM runs of the program.
                             With -t, the runtime for each individual run will be printed.
                             Additionally, a single leading warmup run will be performed (not counted).
                             Only the final run will have its result written to stdout.
  -D/--debugging             Perform possibly expensive internal correctness checks and verbose logging. Implies -L.
  -L/--log                   Print various low-overhead logging information to stderr while running.
  -e/--entry-point NAME      The entry point to run. Defaults to main.
  -b/--binary-output         Print the program result in the binary output format. The default is human-readable text, which is very slow.
  -h/--help                  Print help information and exit.
  -d/--device NAME           Use the first OpenCL device whose name contains the given string.
                             The special string #k, where k is an integer, can be used to pick the k-th device, numbered from zero.
                             If used in conjunction with -p, only the devices from matching platforms are considered.
  --default-group-size INT   The default size of OpenCL workgroups that are launched. Capped to the hardware limit if necessary.
  --default-num-groups INT   The default number of OpenCL workgroups that are launched.
  --default-tile-size INT    The default tile size used when performing two-dimensional tiling
                             (the workgroup size will be the square of the tile size).
  --default-threshold INT    The default parallelism threshold used for comparisons when selecting between code versions generated by incremental flattening.
                             Intuitively, the amount of parallelism needed to saturate the GPU.
  --print-sizes              Print all sizes that can be set with -size or --tuning.
  --size ASSIGNMENT          Set a configurable run-time parameter to the given value.
                             ASSIGNMENT must be of the form "NAME=INT".
                             Use --print-sizes to see which names are available.
  --tuning FILE              Read size=value assignments from the given file.
  -p/--platform NAME         Use the first OpenCL platform whose name contains the given string.
                             The special string #k, where k is an integer, can be used to pick the k-th platform, numbered from zero.
  --dump-opencl FILE         Don8217t run the program, but instead dump the embedded OpenCL program to the indicated file.
                             Useful if you want to see what is actually being executed.
  --load-opencl FILE         Instead of using the embedded OpenCL program, load it from the indicated file.
  --dump-opencl-binary FILE  Don8217t run the program, but instead dump the compiled version of the embedded OpenCL program to the indicated file.
                             On NVIDIA platforms, this will be PTX code.
  --load-opencl-binary FILE  Load an OpenCL binary from the indicated file.
  --build-option OPT         Add an additional build option to the string passed to clBuildProgram().
                             Refer to the OpenCL documentation for which options are supported.
                             Be careful - some options can easily result in invalid results.
  -P/--profile               Gather profiling data while executing and print out a summary at the end.
                             When -r is used, only the last run will be profiled. Implied by -D.
  --list-devices             List all devices and platforms available on the system.
                             Useful for determining what to pass to -d and --p.
```

The help information is taken from the Futhark User's Guide (eg. [here](https://futhark.readthedocs.io/en/stable/man/futhark-c.html) and friends). Perhaps we should think of some way to make sure that those two pieces of information are always in sync?

You'll notice that some lines are quite long. Automatically wrapping overly long lines seems hard (have to handle word-boundaries, soft/hard limits, etc.), so I suggest we just manually insert some newlines at the appropriate places?


athas:
We definitely want this.  Some suggestions:

* The new options should be documented [here somewhere](https://futhark.readthedocs.io/en/latest/usage.html#opencl-specific-options) (it's OK that they only work for the C backends).

* I would prefer much more concise `--help` text, such that each option fits on a line.  `--help` is for reminders, with details going in the manpage or the reference.

* Speaking of [the manpage](https://futhark.readthedocs.io/en/latest/man/futhark-opencl.html), the options should be documented there.

Munksgaard:
Better like this?

```
Usage: ./tests/allocs [OPTION]...
Options:

  -t/--write-runtime-to FILE Print the time taken to execute the program to the indicated file, an integral number of microseconds.
  -r/--runs INT              Perform NUM runs of the program.
  -D/--debugging             Perform possibly expensive internal correctness checks and verbose logging.
  -L/--log                   Print various low-overhead logging information to stderr while running.
  -e/--entry-point NAME      The entry point to run. Defaults to main.
  -b/--binary-output         Print the program result in the binary output format.
  -h/--help                  Print help information and exit.
  -d/--device NAME           Use the first OpenCL device whose name contains the given string.
  --default-group-size INT   The default size of OpenCL workgroups that are launched.
  --default-num-groups INT   The default number of OpenCL workgroups that are launched.
  --default-tile-size INT    The default tile size used when performing two-dimensional tiling.
  --default-threshold INT    The default parallelism threshold.
  --print-sizes              Print all sizes that can be set with -size or --tuning.
  --size ASSIGNMENT          Set a configurable run-time parameter to the given value.
  --tuning FILE              Read size=value assignments from the given file.
  -p/--platform NAME         Use the first OpenCL platform whose name contains the given string.
  --dump-opencl FILE         Dump the embedded OpenCL program to the indicated file.
  --load-opencl FILE         Instead of using the embedded OpenCL program, load it from the indicated file.
  --dump-opencl-binary FILE  Dump the compiled version of the embedded OpenCL program to the indicated file.
  --load-opencl-binary FILE  Load an OpenCL binary from the indicated file.
  --build-option OPT         Add an additional build option to the string passed to clBuildProgram().
  -P/--profile               Gather profiling data while executing and print out a summary at the end.
  --list-devices             List all devices and platforms available on the system.

For more information, consult the Futhark User's Guide.
```

athas:
Yes, that's much better!

Munksgaard:
Huh, the man page on [futhark c](https://futhark.readthedocs.io/en/latest/man/futhark-c.html#futhark-c-1) says that `-t` reports the time in microseconds (and so does my `--help`), but I thought it was really measured in nanoseconds?

athas:
It has always reported microseconds.

Munksgaard:
Huh, I guess you're right. I was sure `futhark bench` reported nanoseconds, but it seems like that's not the case, as you said. I must've had it confused with something else.

athas:
Also remember to mention this in the CHANGELOG.

PR #1150, opened 2020-10-08
Also deploy latest release as "stable"

### LINKED ISSUE #1149: Provide latest stable release
Is there a way to automatically download the latest _stable_ release? Currently, nightly is available at eg. https://futhark-lang.org/releases/, but there's no easy way to get the latest stable release (ie. currently 0.17.3)

The latest stable release should probably also be referenced here, instead of the latest nightly:
https://futhark.readthedocs.io/en/stable/installation.html#installing-from-a-precompiled-snapshot

Munksgaard:
Fixes #1149

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,4 +60,6 @@ jobs:
     - name: scp to server
       env:
         TARBALL: ${{ steps.metadata.outputs.TARBALL }}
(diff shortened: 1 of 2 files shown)


Munksgaard:
Closed as per discussion in #1149

athas:
Just leaving a note that the problem with this technique is that the tarball will unpack to a directory with a different name than the tarball, and that directory will still contain a version number (which will now be unpredictable).

Munksgaard:
Good catch :+1:

PR #1151, opened 2020-10-08
Use $GITHUB_PATH instead of add-path

Munksgaard:
The `add-path` method of adding stuff to $PATH has been deprecated.

More info:

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#adding-a-system-path


athas:
The Python tests fail for some reason that I assume is unrelated.

PR #1181, opened 2020-11-09
Add error handling for bad file paths

philass:
Added some error checking for the case where bad files are provided. The previous error message when a non-existent file path was provided was 
```
futhark datacmp test.data nonexistant.data
```
gives
```
nonexistant.data: openBinaryFile: does not exist (No such file or directory)
```
With the changes the same input gives the following message. 
```
futhark datacmp test.data nonexistant.data          
```
```
No such file : nonexistant.data
```
This omits the whole internal compiler error / please report thing

### DIFF
--- a/src/Futhark/CLI/Datacmp.hs
+++ b/src/Futhark/CLI/Datacmp.hs
@@ -3,19 +3,31 @@
 -- | @futhark datacmp@
 module Futhark.CLI.Datacmp (main) where
 
+import Control.Exception
 import qualified Data.ByteString.Lazy.Char8 as BS
+import qualified Data.Text.IO as T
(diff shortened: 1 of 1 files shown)


athas:
[Similar to this](https://github.com/diku-dk/futhark/blob/master/src/Futhark/Compiler/Program.hs#L127).

athas:
That is still wrong.  You must not separate the check of existence from the reading.  You must catch the exception, if any, *when the file contents are read*.

philass:
@athas Sorry about that and thanks for your patience. I think I've addressed it now. Let me know if there are any other changes you'd like me to make.

PR #1188, opened 2020-11-27
Use single-pass scan in CUDA backend

athas:
This work is by Andreas Nicolaisen and Marco Aslak Persson.

Currently it is only used for non-segmented prim-type scans in the CUDA backend.  It also combines multiple scan operators, if any, into one.  It would be nice, and probably not too difficult, to add support for segments.


athas:
@coancea what was missing here?  Something about not hardcoding the number of elements per thread?  What should the formula be?

athas:
This fails on the lss benchmark.

athas:
This program fails:

```
let lss 't (t: t) (pred1: t -> bool) (pred2: t -> t -> bool) (xs: []t) =
  let max = i32.max

  let rop (lssx, lisx, lcsx, tlx, firstx, lastx)
          (lssy, lisy, lcsy, tly, firsty, lasty) =
    let connect = pred2 lastx firsty || tlx == 0 || tly == 0
    let newlss = if connect then max (lcsx + lisy) (max lssx lssy)
                            else max lssx lssy
    let newlis = if lisx == tlx && connect then lisx + lisy else lisx
    let newlcs = if lcsy == tly && connect then lcsy + lcsx else lcsy
    let first = if tlx == 0 then firsty else firstx
    let last  = if tly == 0 then lastx else lasty
    in (newlss, newlis, newlcs, tlx+tly, first, last)

    let mop x =
      let xmatch = if pred1 x then 1 else 0
      in (xmatch, xmatch, xmatch, 1, x, x)

  in (scan rop (0,0,0,0,t,t) (map mop xs)) |> map (.0)

-- ==
-- entry: lss_iota_i8
-- input { 10000 } auto output

entry lss_iota_i8  n = i64.i32 n |> iota |> map  i8.i64 |> lss 0 (const true) (<=)
```

These are the discrepancies:

```
Value (0,2817) expected 256i32, got 257i32
Value (0,2818) expected 256i32, got 257i32
Value (0,2819) expected 256i32, got 257i32
Value (0,2820) expected 256i32, got 257i32
Value (0,2821) expected 256i32, got 257i32
Value (0,2822) expected 256i32, got 257i32
Value (0,2823) expected 256i32, got 257i32
Value (0,2824) expected 256i32, got 257i32
Value (0,2825) expected 256i32, got 257i32
Value (0,5121) expected 256i32, got 257i32
Value (0,5122) expected 256i32, got 257i32
Value (0,5123) expected 256i32, got 257i32
Value (0,5124) expected 256i32, got 257i32
Value (0,5125) expected 256i32, got 257i32
Value (0,5126) expected 256i32, got 257i32
Value (0,5127) expected 256i32, got 257i32
Value (0,5128) expected 256i32, got 257i32
Value (0,5129) expected 256i32, got 257i32
Value (0,7425) expected 256i32, got 257i32
Value (0,7426) expected 256i32, got 257i32
Value (0,7427) expected 256i32, got 257i32
Value (0,7428) expected 256i32, got 257i32
Value (0,7429) expected 256i32, got 257i32
Value (0,7430) expected 256i32, got 257i32
Value (0,7431) expected 256i32, got 257i32
Value (0,7432) expected 256i32, got 257i32
Value (0,7433) expected 256i32, got 257i32
Value (0,9729) expected 256i32, got 257i32
Value (0,9730) expected 256i32, got 257i32
Value (0,9731) expected 256i32, got 257i32
Value (0,9732) expected 256i32, got 257i32
Value (0,9733) expected 256i32, got 257i32
Value (0,9734) expected 256i32, got 257i32
Value (0,9735) expected 256i32, got 257i32
Value (0,9736) expected 256i32, got 257i32
Value (0,9737) expected 256i32, got 257i32
```

So four stretches of nine elements, equal to the per-thread work count.  Also, the errors are in the same spot every run.  Not a coincidence.

athas:
The start of the incorrect areas are spaced exactly 256*9 positions apart.  It seems like they are produced by thread 57 within each thread block, which is unfortunately a less meaningful number...

athas:
Oh, due to the transpose stuff, I don't actually have a clue which thread is responsible.

athas:
The bug was just a mixup of argument order during the distribution phase, which only matters for non-commutative scan operators.

athas:
There is still a problem for the i32 version of LSS, but now the error is random for every run.  Frustrating.  Probably something is wrong in the lookback phase.

athas:
It was a similar issue.  This is soon ready for merging.  I'll just peg gpu04 with a few final benchmarks.

PR #1190, opened 2020-12-01
Add result of comparison to debug output

Munksgaard:
I think this could be helpful when reading debug output of programs with many thresholds.


athas:
This still works with the auto-tuner (which searches the output for lines of this form)?

Munksgaard:
Good catch! Yes, it still works, because the regex being matched on by the autotuner is `"Compared ([^ ]+) <= (-?[0-9]+)"`

PR #1197, opened 2020-12-04
Accumulators

athas:
@coancea we need to discuss aspects of this design in detail, because parts here are *invasive* in the IR, and I want to make sure it does what we need.  In particular this opens the door to having arrays whose elements are *not scalars*, and that is pretty big.

The implementation also cannot currently handle histogram-like accumulators for multi-dimensional arrays, which we'll need to address.


athas:
Type checking is not complete for the core language (e.g. `MkAcc` does not consume its arrays).  That has to be fixed, but it's a little tricky, as when doing function calls, we pass the arrays as "phantom arrays".

athas:
Unfortunately we also found some rather fundamental issues in the core IR that means we cannot properly handle e.g.

```futhark
let f (acc: *acc ([]i32)) i =
  let acc = write acc (i*2) (i32.i64 i)
  let acc = write acc (i*2+1) (i32.i64 i)
  in acc

let main n m k =
  tabulate n (\i ->
                let xs = replicate m (i32.i64 i)
                in scatter_stream xs f (iota k))
```

Fixing this is on the critical path for me.  I have some ideas, but still working out the details.

athas:
When we merge this, are we really going to call the user-level constructs `scatter_stream` and `reduce_by_index_stream`, or can we come up with something more snappy?

athas:
This is no longer a particularly invasive design, and we should be OK with merging it soon.  I'm still not particularly happy with the names `scatter_stream` and `reduce_by_index_stream`, but I suppose we can live with it.  We can also hold back on providing the source language constructs for a little while.

PR #1199, opened 2020-12-12
Wasm

philass:
preliminary PR for WASM backend


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   * Fixed compiler bug on segmented rotates where the rotation amount
     is variant to the nest (#1192).
 
+
```
athas, line 36: Spurious line.

rts/c/values.h
```diff
@@ -2,10 +2,14 @@
 
 //// Text I/O
 
+
```
athas, line 5: Spurious line.

rts/c/values.h
```diff
@@ -2,10 +2,14 @@
...
 typedef int (*writer)(FILE*, const void*);
 typedef int (*bin_reader)(void*);
 typedef int (*str_reader)(const char *, void*);
 
+FILE * OUTPUT;
+FILE * STREAM;
```
athas, line 11: This naming seems inconsistent.  Should it not be `OUTPUT` and `INPUT`?

src/Futhark/CLI/WASM.hs
```diff
@@ -0,0 +1,19 @@
+{-# LANGUAGE FlexibleContexts #-}
+
+-- | @futhark c@
```
athas, line 3: futhark wasm

athas:
I made some `values.h` changes in master to give stdin/stdout a less privileged position.

.github/workflows/main.yml
```diff
@@ -42,7 +42,6 @@ jobs:
         sudo apt-get update
         sudo apt-get install ghc-8.8.4
         echo "/opt/ghc/bin" >> $GITHUB_PATH
-
```
athas, line 45: Restore the missing linebreaks.

rts/c/values.h
```diff
@@ -632,6 +692,7 @@ static int read_bin_array(FILE *f,
   *data = tmp;
 
   int64_t num_elems_read = (int64_t)fread(*data, (size_t)elem_size, (size_t)elem_count, f);
+
```
athas, line 695: Remove extra line.

src/Futhark/Actions.hs
```diff
@@ -138,6 +139,35 @@ runCC cpath outpath cflags_def ldflags = do
     Right (ExitSuccess, _, _) ->
       return ()
 
+cmdEMCC :: String
+cmdEMCC = fromMaybe "emcc" $ lookup "emcc" unixEnvironment
```
athas, line 143: What is the purpose of this?  Is there really a convention of using `emcc` as an environment variable?

rts/c/values.h
```diff
@@ -6,6 +6,9 @@ typedef int (*writer)(FILE*, const void*);
 typedef int (*bin_reader)(void*);
 typedef int (*str_reader)(const char *, void*);
 
+FILE * OUTPUT;
```
athas, line 9: Make these `static`.

Also, all-uppercase identifiers are reserved for preprocessor constants. `output_file` and `output_file` would be better names, I think.

I still don't like the use of global variables for this at all, though.  Can't we do this in the entry point functions somehow?  I guess it's good enough for now.

athas (changes requested):
Also, you'll need to merge master into your branch again.

src/Futhark/Actions.hs
philass: I tried following what was done here
```haskell
cmdCC = fromMaybe "cc" $ lookup "CC" unixEnvironment
```
But I was just blindly copying code. I think you are correct that it doesn't make sense for emcc.

I'll simply change it to 
```
cmdEmcc = "emcc"
```

athas: Just hard-code it.  No need for a variable.  `CC` is a decades-old Unix convention.

src/Futhark/CodeGen/Backends/GenericC.hs
```diff
@@ -1170,7 +1170,7 @@ onEntryPoint get_consts fname (Function _ outputs inputs _ results args) = do
          if (!($exp:(allTrue entry_point_input_checks))) {
            ret = 1;
            if (!ctx->error) {
-             ctx->error = msgprintf("Error: entry point arguments have invalid sizes.");
+             ctx->error = msgprintf("Error: entry point arguments have invalid sizes.\n");
```
athas, line 1173: Why is this necessary?

philass: Emscripten has really weird behaviour when prints statements aren't ended with new line characters

This created a couple failed tests in the futhark test suite, for example in emptydim0.fut this test failed with unexpected behaviour. Specifically these sort of test cases below.

```futhark
-- ==
-- compiled input { [[1]] [[1,2]] } error: .
-- compiled input { [[1,2]] [[1]] } error: .

let main [n] (xs: [][n]i32) (ys: [][n]i32) = (xs, ys)
````

Comment (athas) [reply src/Futhark/CodeGen/Backends/GenericC.hs:1173]: It's not just a matter of `fflush()`ing the error stream?

Comment (athas) [reply src/Futhark/CodeGen/Backends/GenericC.hs:1173]: I'll add the `\n` on master, since it should be there anyway.

**Comment by athas** [comment]
Looks like this is about ready for merging.

**Comment by philass** [comment]
Thanks for the PR! You are a legend

**Comment by philass** [comment]
Anything else you would like me to do before the merging?

**Comment by athas** [comment]
Well, do you think it should be merged now?  There are two reasons I can think of why not:

1. The `-t` option doesn't work, so `futhark bench` will malfunction.  I'll fix this one myself.
2. Is it at all realistic to use it in its current state?  You can probably answer this better than I.  As I understand it, it just dumps a `.wasm` file, which you can then load from JavaScript, but is there any way the JavaScript programmer would know which functions are available, except for studying the C API?  Maybe we should generate a *thin* low-level JavaScript wrapper.  A higher-level wrapper can be done later.

The first issue is critical, but you should consider the second one too.  I expect the second one will be addressed in due time eventually, and it's less important to me whether it's before or after the merge.

**Comment by philass** [comment]
Yeah, you are right. It would be unusable with the current level of documentation. I'll start dealing with the low level wrapper. It'll probably quite insightful for building a higher level one later.

**Comment by athas** [comment]
With the switch to [server mode](https://futhark.readthedocs.io/en/latest/server-protocol.html) for testing, this no longer works, as line-by-line blocking IO does not work in Node.

The best way forward is to focus on the library support, and then write a server mode implementation in JavaScript that uses the library code.  It's a pretty simple protocol anyway (the [Python implementation](https://github.com/diku-dk/futhark/blob/master/rts/python/server.py)) is less than 200 lines of code.

File: rts/javascript/values.js [hunk 10]
```diff
@@ -0,0 +1,63 @@
+// Start of values.js
+//
+//
+
+class ReaderInput {
+  constructor(f) {
+    this.f = f;
+    this.lookahead_buffer = [];
+  }
```
athas, line 9: *Do not* implement the text-based format!

philass:
I am closing this because the relevant code is now in the [wasm-multicore
](https://github.com/diku-dk/futhark/tree/wasm-multicore) branch. But there be some code from this PR that we still want in the wasm-multicore branch. Specifically the logic for bypassing the UTF8-decoding issues

athas:
Regarding the UTF-8 decoding stuff, I suggest simply not supporting non-server executables for the WASM backends.  It's just not worth the bother.

Issue #1179, opened 2020-11-08
Server mode


athas:
I'd like to add a flag to compiled Futhark programs that makes them act as a sort of REPL, instead of just running a single entry point.  The REPL is not supposed to be human-friendly, but rather to be used by other programs.  The REPL will allow one to load in data of any builtin type (including arrays), print said data, and call entry points and bind the results to names that can subsequently be printed, or passed to new entry points.

This will have two advantages:

* Testing and benchmarking multiple entry points and input data sets will be faster, because we can reuse the same process, rather than re-starting the process (and hence re-initialising GPU state) for every new input.

* We can more easily build tooling.  For example, [this tool for property-based testing](https://github.com/Unigurd/fucheck) is hampered by having to go through the library interface (the C API), which involves dynamic linking and other technical trickery.

Technically, we would probably do this by adding yet another compilation mode (like `--library`) called perhaps `--daemon`.  This would then replace the current CLI interface with this REPL mode, which would just make internal use of the C (or Python) API.

I expect that a proof-of-concept can be made without any modifications to the Futhark compiler at all.

FluxusMagna:
Have you decided how to handle tuning of entry points with nontrivial types? This seems like a step in that direction at least.

athas:
Yes, this would be used for that, among other things.

athas:
Implemented by #1221.

PR #1215, opened 2021-01-04
Add memory info to futhark bench json output

Munksgaard:
This commit adds support for benchmarking memory usage. It does so by changing `futhark bench` to always run with `-P`, and parses the output, searching for memory information. If any is found, it is added to the json output.

I've also added support in tools/cmp-bench-json.py for reporting the memory usage improvements, if present and significant.

screenshot

### DIFF
--- a/src/Futhark/Bench.hs
+++ b/src/Futhark/Bench.hs
@@ -7,6 +7,7 @@ module Futhark.Bench
   ( RunResult (..),
     DataResult (..),
     BenchResult (..),
+    Result (..),
     encodeBenchResults,
     decodeBenchResults,
     binaryName,
@@ -24,6 +25,7 @@ import qualified Data.Aeson as JSON
 import qualified Data.ByteString.Char8 as SBS
 import qualified Data.ByteString.Lazy.Char8 as LBS
 import qualified Data.HashMap.Strict as HM
+import qualified Data.Map as M
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import qualified Data.Text.IO as T
@@ -40,10 +42,17 @@ import System.Timeout (timeout)
 newtype RunResult = RunResult {runMicroseconds :: Int}
   deriving (Eq, Show)
 
--- | The results for a single named dataset is either an error
--- message, or runtime measurements along the stderr that was
+data Result = Result
+  { runResults :: [RunResult],
+    memoryMap :: M.Map T.Text Int,
+    stdErr :: T.Text
+  }
+  deriving (Eq, Show)
+
+-- | The results for a single named dataset is either an error message, or
+-- runtime measurements, the number of bytes used, and the stderr that was
 -- produced.
-data DataResult = DataResult String (Either T.Text ([RunResult], T.Text))
+data DataResult = DataResult String (Either T.Text Result)
   deriving (Eq, Show)
 
 -- | The results for all datasets for some benchmark program.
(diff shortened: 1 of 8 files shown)


athas:
I worry that using `-P` will reduce performance.  Is that the case?  We can always add another option that makes it print the memory statistics (which are unconditionally collected in all cases) but not enable all the other profiling instrumentation.  Maybe `-L` already does that.

Munksgaard:
> I worry that using `-P` will reduce performance. Is that the case?

A quick test on OptionPricing.fut indicates that it does look to be the case for the CUDA backend, but not for the OpenCL backend.

I see you argument about adding another option instead. I think I'll do that instead.

Munksgaard:
Is it okay if the log output comes after the program output? As in:

```
$ cat futhark-benchmarks/finpar/OptionPricing-data/small.in | futhark-benchmarks/finpar/OptionPricing -L
Using platform: NVIDIA CUDA
Using device: GeForce RTX 2080 Ti
Lockstep width: 32
Default group size: 256
Default number of groups: 272
Compared main.suff_outer_par_1 <= 1: false.
[167.055695f32]
Peak memory usage for space 'device': 3621040 bytes.
```

athas:
As long as the memory logging is on stderr, that's fine.

And I'm actually going to replace this entire mechanism in #1179.

src/Futhark/CLI/Bench.hs
```diff
@@ -297,7 +298,22 @@ runBenchmarkCase opts futhark program entry pad_to tr@(TestRun _ input_spec (Suc
...
+          Nothing ->
+            Left "Could not parse memory usage from profiling output"
+        )
+        & DataResult dataset_desc
+        & Just
+        & return
+
+getMemoryUsage :: T.Text -> Maybe Int
+getMemoryUsage t =
+  let mem_regex = "Peak memory usage for space '[^']+': ([0-9]+) bytes." :: T.Text
```
athas, line 313: What happens if there are multiple memory spaces?

Munksgaard: I am not really sure what this means, but cf. https://github.com/diku-dk/futhark/pull/1215/files/b4cb1dc490436b353326ca8268d1cde447a7961c#diff-c8ca25d5bf1f5928bbd214a58075cdf189a2dbbdee9007c06fceb8225d4238b0R315-R316, we will return `Nothing`. What would you prefer to happen?

athas: This should return a mapping from memory space to the peak, and the JSON should contain a dict.

We will certainly use both CPU and GPU memory in the same program in the relatively near future, and we might do even more funky things in the distant future, who knows.

Munksgaard:
I've changed the implementation to return a map and dict instead. This is what it looks like with more than one device:

screenshot

src/Futhark/Bench.hs
```diff
@@ -24,6 +24,7 @@ import qualified Data.Aeson as JSON
 import qualified Data.ByteString.Char8 as SBS
 import qualified Data.ByteString.Lazy.Char8 as LBS
 import qualified Data.HashMap.Strict as HM
+import Data.Map (Map)
```
athas, line 27: We generally use the qualified type name for this (`M.Map`).

src/Futhark/Bench.hs
```diff
@@ -40,10 +41,10 @@ import System.Timeout (timeout)
 newtype RunResult = RunResult {runMicroseconds :: Int}
   deriving (Eq, Show)
 
--- | The results for a single named dataset is either an error
--- message, or runtime measurements along the stderr that was
+-- | The results for a single named dataset is either an error message, or
+-- runtime measurements, the number of bytes used, and the stderr that was
 -- produced.
-data DataResult = DataResult String (Either T.Text ([RunResult], T.Text))
+data DataResult = DataResult String (Either T.Text ([RunResult], Map T.Text Int, T.Text))
```
athas, line 47: Maybe it is time to actually create a data type for this triple.

unittests/Futhark/BenchTests.hs
```diff
@@ -2,6 +2,7 @@
 
 module Futhark.BenchTests (tests) where
 
+import qualified Data.Map as Map
```
athas, line 5: `as M`!

athas (changes requested):

flip111:
Really cool feature @Munksgaard

src/Futhark/Bench.hs
Munksgaard: Done. Although, I couldn't really think of a good name... Any suggestions?

athas (approved):

Issue #1210, opened 2020-12-28
FutharkScript


athas:
I want a way to take commented Futhark programs written in [this style](https://github.com/diku-dk/futhark-website/blob/master/examples/dex-raytrace.fut) and turn them into Markdown files with visualisations (images/plots) of certain values.  The inspiration is [these reports produced by Dex](https://google-research.github.io/dex-lang/raytrace.html).

I don't think this would be particularly hard to implement.

Munksgaard:
This is literal programming, right? An easy-ish way to achieve something equivalent would be to implement Futhark support for [org-mode's babel](https://orgmode.org/worg/org-contrib/babel/).

athas:
I think that would be more work.  It's also important to me that the source programs are ordinary `.fut` files that the compiler can process without issue (it doesn't have to type-check the visualisation annotations).

I think Knuth would be sad if we called this literate programming.  It goes against one of his core goals, namely that physical declaration order does not have to be the same as the presentation order.

kfl:
It looks like a job for pandoc, it's similar to inverted `.lhs`.

kfl:
[The Padoc documentation](https://pandoc.org/using-the-pandoc-api.html) shows an example converting JSON to Word files.

athas:
I think Pandoc is overkill.  (Or rather, it will eventually be used for converting the produced Markdown to HTML.)  The technical work that has to be done is to look through the program for comments with (strawman syntax):

```
-- %image main 250
```

We'd then inspect the type of `main` and figure out how to best convert that to an image, then compile the program, run it, pass the result into some kind of data-to-image converter, and spit out a Markdown file where those comments have been replaced with HTML containing the image as a data URL.

kfl:
But that's what pandoc filters are made for? Or Am I misunderstanding you?

You want something similar to [pandoc-markdown-ghci-filter](https://hackage.haskell.org/package/pandoc-markdown-ghci-filter), but for futhark, right?

athas:
It's possible, but actually using the interpreter will be far too slow.  We must compile the program using the ordinary compiler, and I'm not sure Pandoc filters are appropriate when we must have access to the original source file.  [Actually extracting the graphics directives is trivial](https://github.com/diku-dk/futhark-website/blob/c34c1954e24ec259522df2a9f29b21bbd03638d9/site.hs#L27-L46), so Pandoc seems a heavyweight way to solve the simplest problem.

athas:
This depends on #1179.

PR #1224, opened 2021-01-09
Futharkscript

athas:
Here is an example of a FutharkScript program:

```
-- # The Example
--
-- Welcome to my great Futhark program!
--
-- Check out this shit:

entry fact (n: i32) = i32.sum (1...n)

-- Cool, right?
--
-- Now dig this:

-- :res fact 10i32

-- And it can get even fancier:

entry mkimg (h: i64) (w: i64) =
  tabulate_2d h w (\i j -> u32.f64 ((f64.i64 (i*j) / f64.i64 (h*w))*255))

-- :img mkimg 100i64 300i64

-- We can also do advanced data analysis.

entry analysis (n: i64) =
  unzip (tabulate n (\i -> (f64.i64 i, f64.sqrt (f64.i64 i))))

-- :plot2d<500,200> analysis 100i64
```

Which produces this: https://sigkill.dk/junk/example.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
     means that validation of test and benchmark results is much faster
     (close to an order of magnitude).
 
+  * The experimental `futhark script` command allows vaguely a
+    notebook-like programming experience.
+
 ### Removed
 
 ### Changed
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -273,7 +273,8 @@ man_pages = [
     ('man/futhark-bench', 'futhark-bench', 'benchmark Futhark programs', [], 1),
     ('man/futhark-doc', 'futhark-doc', 'generate documentation for Futhark code', [], 1),
     ('man/futhark-dataset', 'futhark-dataset', 'generate random data sets', [], 1),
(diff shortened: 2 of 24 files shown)


kfl:
It seems that a script cannot specify which backend to use nor use multiple backends.

I imagine that it could be useful to be able to write something like:

~~~
...
-- Using OpenCL we see the following floating point defect when evaluating `important_fun`:
-- > :backend OpenCL important_fun(420, 69)
-- Compare with evaluation on CPU
-- > important_fun(420, 69)
~~~

Following up on that, it would also be useful with a `:benchmark` directive, especially when combined with multiple backends.

(Meta: would you have prefered to get this comment on issue #1210 ?)

athas:
The backend is given by `--backend` when running the script.

A `:benchmark` directive is definitely a good idea.

Multiple backends within the same script is technically possible I think, but I think it will have to wait for later, as the current implementation is only set up to keep one instance of the program running while evaluating the script.

kfl:
My comment was prompted exactly because `--backend` is an option outside the script.

Thus, if you want to write a quick example script that demonstrate the difference between different backends (my first example was benchmarks, not floating points numbers), then you are faced with problem of having to merge the output of your runs with different backends.

Not a big problem, but maybe a future feature request.

PR #1227, opened 2021-01-15
Rename negate to neg.

athas:
We'll merge this shortly before the next release.  This function is rarely used anyway.


athas:
I've changed this to keep `negate` around as an alias for the next release, and then we can remove it.  Sadly we don't have a way of marking deprecated functions such that the compiler will complain.

Issue #1235, opened 2021-01-24
Rotating beyond the size of the array is identity


athas:
Surely that's not by design?

philass:
It seems that this is only an issue in the Interpreter. 

When I test against the Python, C, and OpenCL backends. I get expected behaviour (not the identity).
```
Testing rotation on
\[1, 2, 3, 4, 5\]
Sanity check
testing 0
--------------------
[1i32, 2i32, 3i32, 4i32, 5i32]
^ Python
[1i32, 2i32, 3i32, 4i32, 5i32]
^ Sequential C
[1i32, 2i32, 3i32, 4i32, 5i32]
^ OpenCl
---------------------
Testing rotation of 1
---------------------
[2i32, 3i32, 4i32, 5i32, 1i32]
[2i32, 3i32, 4i32, 5i32, 1i32]
[2i32, 3i32, 4i32, 5i32, 1i32]
---------------------
Testing rotation of 6
---------------------
[2i32, 3i32, 4i32, 5i32, 1i32]
[2i32, 3i32, 4i32, 5i32, 1i32]
[2i32, 3i32, 4i32, 5i32, 1i32]
---------------------
Testing rotation of 11
---------------------
[2i32, 3i32, 4i32, 5i32, 1i32]
[2i32, 3i32, 4i32, 5i32, 1i32]
[2i32, 3i32, 4i32, 5i32, 1i32]
----------------------
Now testing negative rotations
----------------------
---------------------
Testing rotation of -1
---------------------
[5i32, 1i32, 2i32, 3i32, 4i32]
[5i32, 1i32, 2i32, 3i32, 4i32]
[5i32, 1i32, 2i32, 3i32, 4i32]
---------------------
Testing rotation of -6
---------------------
[5i32, 1i32, 2i32, 3i32, 4i32]
[5i32, 1i32, 2i32, 3i32, 4i32]
[5i32, 1i32, 2i32, 3i32, 4i32]
---------------------
Testing rotation of -11
---------------------
[5i32, 1i32, 2i32, 3i32, 4i32]
[5i32, 1i32, 2i32, 3i32, 4i32]
[5i32, 1i32, 2i32, 3i32, 4i32]
```

When tested against the Interpreter
```
[0]> rotate (1) [1, 2, 3, 4, 5]
[2i32, 3i32, 4i32, 5i32, 1i32]
[1]> rotate (6) [1, 2, 3, 4, 5]
[...]

PR #1236, opened 2021-01-25
Fix rotation in interpreter

philass:
I believe this should make the interpreters rotation have the expected behaviour and fix the issue in #1235.

### DIFF
--- a/src/Language/Futhark/Interpreter.hs
+++ b/src/Language/Futhark/Interpreter.hs
@@ -1626,7 +1626,7 @@ initialCtx =
       fun2t $ \i xs -> do
         let (shape, xs') = fromArray xs
         return $
-          let idx = rem (asInt i) (length xs')
+          let idx = if null xs' then 0 else rem (asInt i) (length xs')
            in if idx > 0
                 then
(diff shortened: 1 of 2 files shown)


athas:
Please also add a test case to [rotate0.fut](https://github.com/diku-dk/futhark/blob/master/tests/rotate0.fut) so we can be sure this does not reoccur.

philass:
Added the test. Should be all good now.

src/Language/Futhark/Interpreter.hs
```diff
@@ -1626,13 +1626,14 @@ initialCtx =
...
         let (shape, xs') = fromArray xs
         return $
-          if asInt i > 0
-            then
-              let (bef, aft) = splitAt (asInt i) xs'
-               in toArray shape $ aft ++ bef
-            else
-              let (bef, aft) = splitFromEnd (- asInt i) xs'
-               in toArray shape $ aft ++ bef
+          let idx = rem (asInt i) (length xs')
```
athas, line 1629: What if the array is empty?

Please also add a test case for this.

philass: Good catch

PR #1245, opened 2021-02-08
Use i64 indexes, fix stringlit definition

NPN:
Bit unrelated, but I found it surprising that while any signed type can be used to index an array, only `i64` can be used to slice an array. The docs mention the latter, but "array index must always be of type `i64`" implies that code like `a[0i8]` shouldn't work.

On that note, perhaps it should be more clearly stated somewhere that indexes are `i64`. Unless I've missed something, it's only mentioned in:

* The blog posts about switching to 64-bit indexes
* Chapter 2 of the Futhark book: "Size parameters are always of type `i64`"
* Chapter 3 of the Futhark User's Guide: `3.4.3.9. a[i:j:s]` and `3.4.3.21. e : t` under the "Semantics of Simple Expressions" (Also, the language grammar for `index` implies that any `exp` is valid, though maybe the grammar only cares about syntax and not semantics.)

Maybe it would be good to mention this when arrays are introduced in the book and in the language reference under `3.4.3.8. a[i]`. I can add this if needed.

### DIFF
--- a/docs/language-reference.rst
+++ b/docs/language-reference.rst
@@ -554,8 +554,8 @@ A variable name; evaluates to its value in the current environment.
 `stringlit`
 ...........
 
-Evaluates to an array of type ``[]i32`` that contains the code points
-of the characters as integers.
+Evaluates to an array of type ``[]u8`` that contains the characters
+encoded as UTF-8.
 
 ``()``
 ......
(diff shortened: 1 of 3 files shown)


athas:
> Bit unrelated, but I found it surprising that while any signed type can be used to index an array, only i64 can be used to slice an array. The docs mention the latter, but "array index must always be of type i64" implies that code like a[0i8] shouldn't work.

Ideally, you could use integers of any type.  The restriction to `i64` for slices is because of the size type system.  As a rather useful convenience, we want `a[0:i]` to produce an array of size `i`, but that is only valid if `i` is of type `i64`.  I think it would be confusing if `a[0:i]` silently produced an unknowable size whenever `i` happened to be some other integer type, particularly since that might manifest as a mysterious size incompatibility later on in the program.  I think this restriction could be lifted if we implemented better diagnostics and/or warnings in the compiler for such cases.

Thanks for the fixes, and feel free to also make similar fixes to the book.  Unfortunately, these things can be pretty hard to spot when you're the author of the text...

PR #1261, opened 2021-03-02
Parser for the Futhark core language.

### LINKED ISSUE #1259: Adding a parser for the core language
This is not urgent, but I think we should have a parser for the core language.  The sexp representation is serviceable, but it is horrible to read.  Compare

```Futhark
fun {i32, i32, i32, i32} main (a_3940 : i32, b_3941 : i32) = {
  let {main_res_3942 : i32} = add32(1i32, a_3940)
  let {main_res_3943 : i32} = sub32(b_3941, 1i32)
  let {main_res_3944 : i32} = ashr32(b_3941, 1i32)
  in {a_3940, main_res_3942, main_res_3943, main_res_3944}
}
```

and

```lisp
(prog
  ()
  (fundef
    nil
    {}
    main
    (i32 i32 i32 i32)
    ((param a_3940 i32) (param b_3941 i32))
    (nil
      ((let
         (() ((main_res_3942 i32)))
         (aux () {} nil)
         ((add i32 wrap) 1i32 a_3940))
       (let
         (() ((main_res_3943 i32)))
         (aux () {} nil)
         ((sub i32 wrap) b_3941 1i32))
       (let
         (() ((main_res_3944 i32)))
         (aux () {} nil)
         ((ashr i32) b_3941 1i32)))
      (a_3940 main_res_3942 main_res_3943 main_res_3944))))
```

A well defined core language representation would also make it feasible to write transformations as external tools ("plugins"), which might lead to interesting possibilities.

I already wrote a quick proof-of-concept for a subset of the language, and I think it's quite viable.  The biggest problem is that there are parts of our IR that we've taken to printing as comments, which will not fly.  For example conditionals:

```Futhark
entry {i32} main (b_3928 : bool, x_3929 : i32, y_3930 : i32) = {
  let {main_res_3931 : i32} =
    -- Branch returns: {i32}
    if b_3928
    then {x_3929} else {y_3930}
  in {main_res_3931}
}
```

This will need a different syntax.  Maybe

```Futhark
if b_3928
then {x_3929} else {y_3930}
: {i32}
```

Doesn't have to be fancy or even very neat (nobody will write this by hand), but it should be simple to parse and easy to read.

athas:
Closes #1259.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -370,6 +370,9 @@ jobs:
         echo "$HOME/.local/bin" >> $GITHUB_PATH
 
     - run: |
+        tools/testparser.sh tests
+
+    - run: |
         cd pkgtests && sh test.sh
 
     - run: |
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ core.*
 # Ignore directories often created by testing
 data/
 lib/
+testparser/
--- a/futhark.cabal
+++ b/futhark.cabal
@@ -147,6 +147,7 @@ library
       Futhark.IR
       Futhark.IR.Aliases
       Futhark.IR.Decorations
+      Futhark.IR.Parse
       Futhark.IR.Kernels
       Futhark.IR.Kernels.Kernel
       Futhark.IR.Kernels.Simplify
--- a/futhark.cabal
+++ b/futhark.cabal
@@ -325,7 +325,6 @@ library
     , process >=1.4.3.0
     , process-extras >=0.7.2
     , regex-tdfa >=1.2
-    , sexp-grammar >= 2.2.1
     , srcloc >=0.4
     , template-haskell >=2.11.1
     , temporary
@@ -382,7 +381,6 @@ test-suite unit
     , megaparsec
     , mtl
     , parser-combinators
-    , sexp-grammar
     , tasty
     , tasty-hunit
     , tasty-quickcheck
--- a/futhark.cabal
+++ b/futhark.cabal
@@ -61,6 +61,7 @@ library
       Futhark.Analysis.Metrics
       Futhark.Analysis.Metrics.Type
       Futhark.Analysis.PrimExp
+      Futhark.Analysis.PrimExp.Parse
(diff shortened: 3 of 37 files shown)


Munksgaard:
I never got around to it when implementing the sexp-based parser/prettyprinter, but it would be nice to have the ability to easily print and parse a given program and verify that the result is the same. If I remember correctly, a blocker for this was that the error message locations are lost during pretty-printing/parsing, but perhaps there's a way around that.

Ideally, this could be a script that we could run once in a while on _all_ the futhark tests and benchmarks, just to make sure the parser is sound.

What do you think? Would that be doable?

athas:
It's probably doable.  The prettyprinter is currently lossy in the sense that it doesn't print full source locations.  We can fix that, but when/if we add more detailed source locations for *every* statement in the future, this will likely become completely unreadable.

For now, I was just going to test that two passes through the parser/prettyprinter will yield the (textually) same program.  That doesn't imply the first pass won't mess it up or lose information, though.

athas:
It can now handle MCMem and KernelsMem.  The only missing piece is recording the entry point metadata correctly.  @Munksgaard, at your leisure, can you check whether this can replace the sexp representation for your purposes?

Munksgaard:
Ah! I was just about to note that entry points didn't work. I'll try again.

Munksgaard:
Seems to work well for me. It's a lot easier to mess around with the IR in this form than in the sexp-based form!

athas:
I think eventually we can simplify some of the notation.  In particular it would be nice to have a more concise representation for when the index function is the "obvious" row-major one.  But that can come later.  Is it OK with you that I remove the sexp stuff as part of this PR?

Munksgaard:
> I think eventually we can simplify some of the notation. In particular it would be nice to have a more concise representation for when the index function is the "obvious" row-major one. But that can come later. Is it OK with you that I remove the sexp stuff as part of this PR?

Sure, go ahead.

athas:
Interestingly, with the removal of the sexp grammar, this is a net increase of less than 200 lines of code.  Probably will be more work to maintain, though, but at least the code is easier to understand.

PR #1266, opened 2021-03-07
fix futhark-test examples in docs, and add tests.

philass:
I think there are some uncaught mistakes. In the man pages, specifically with regards to `futhark-test`. 

The first program example gives the following when run with vanilla `futhark test`
```
futhark test example1.fut
example1.fut:
Compiling with --backend=c:
Running compiled program:
Running ./example1:
Entry point: main; dataset: #1 ("[4i32, 3i32, 2i32, 1i32] 5i64"):
Expected error:
  Assertion.*failed
Got error:
  Error: Index [5] out of bounds for array of shape [4].

Backtrace:
-> #0  example1.fut:10:3-6
   #1  example1.fut:9:1-10:6


│ programs │ 0      │ 1      │ 0/1       │
│ runs     │ 1      │ 1      │ 0/2       │
```

The second example program also doesn't pass the test suite provided for it. In fact it doesn't even compile.
```
futhark test example2.fut
example2.fut:
Compiling with --backend=c:
Error at example2.fut:8:32-32:
Cannot apply "add" to "x" (invalid type).
Expected: (i32, i32)
Actual:   i32

If you find this error message confusing, uninformative, or wrong, please open an issue at
https://github.com/diku-dk/futhark/issues.
```
I have gone on to fix the tests so that they compile, and pass the test suites provided for them. I have also added test cases in a new folder `man/test` so that if some changes break the examples in the doc the CI will fail.

### DIFF
--- a/docs/man/futhark-bench.rst
+++ b/docs/man/futhark-bench.rst
@@ -134,20 +134,20 @@ EXAMPLES
 The following program benchmarks how quickly we can sum arrays of
 different sizes::
 
-  -- How quickly can we reduce arrays?
-  --
-  -- ==
-  -- nobench input { 0 }
-  -- output { 0 }
-  -- input { 100 }
-  -- output { 4950 }
-  -- compiled input { 100000 }
-  -- output { 704982704 }
-  -- compiled input { 100000000 }
-  -- output { 887459712 }
-
-  let main(n: i32): i32 =
-    reduce (+) 0 (iota n)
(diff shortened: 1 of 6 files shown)


philass:
My last commit fixes the docs for futhark-bench and adds a test for the example program. This seems to have slipped through the cracks with the i32 -> i64 transition.
```
 let main(n: i32): i32 =
    reduce (+) 0 (iota n)
```

PR #1273, opened 2021-03-18
Better scatter docs

Munksgaard:
Also refactored some of the code used when further processing scatters.


athas:
I have been thinking we should start using [Notes](https://www.stackbuilders.com/news/the-notes-of-ghc) for these long comments, but that can come later.

PR #1279, opened 2021-03-19
Interpret byte-arrays (u8) as greyscale images

### LINKED ISSUE #1278: futhark literate :img should interpret u8 as greyscale
"should" is perhaps a strong word, but `u8` seems to me like a better format to store greyscale information in.

Munksgaard:
Fixes #1278

### DIFF
--- a/docs/man/futhark-literate.rst
+++ b/docs/man/futhark-literate.rst
@@ -140,10 +140,11 @@ The following directives are supported:
 
 * ``> :img e``
 
-  Visualises ``e``, which must be of type ``[][]i32``, ``[][]u32`` or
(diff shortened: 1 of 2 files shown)


athas:
Then how will we represent 8 bit RGB images?

Munksgaard:
> Then how will we represent 8 bit RGB images?

Hm, good question...

Munksgaard:
It might also be nice to be able to interpret [][]bool as a a black and white picture.

athas:
`f32` is in practice often used for greyscale values.  Nobody uses eight-bit colours for data processing, but `u8` is certainly a common storage format.

Issue #1281, opened 2021-03-21
NumPy DeprecationWarnings in PyOpencl backend


EvilMav:
PyOpenCL modules use NumPy type aliases which are identical to using python types and are therefore deprecated: [Using the aliases of builtin types like np.int is deprecated](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations).

This results in hundreds of warnings like:
```
  .../emg-thz/emg/thz/futhark/rcsj_ocl.py:17977: DeprecationWarning: `np.long` is a deprecated alias for `np.compat.long`. To silence this warning, use `np.compat.long` by itself. In the likely event your code does not need to work on Python 2 you can use the builtin `int` for which `np.compat.long` is itself an alias. Doing this will not modify any behaviour and is safe. When replacing `np.long`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    byte_count=np.long((replicate_arg_24549 * np.int32(4))))

```

Very annoying but probably a trivial fix...

EvilMav:
Warnings seem to come for two types: `np.bool` and `np.long`. The codebase can be greped for them. `np.bool` can be replaced with bool. I'm not sure it `np.long` should be replaced with generic `int` or fixed precision `np.int64`.

athas:
Definitely `np.int64`.

EvilMav:
> Definitely `np.int64`.

The old way was equivalent to `int` if I read the docs correctly though ;) Should I prepare a pull request or are you already on it? (On a side note - are there any beginner docs on how to run tests for the compiler? For those trying to make themselves at least somewhat useful but being lost on the unfamiliar haskell terrain ;) )

athas:
It should be `np.int64`.  Probably it didn't make a difference in any of the spots where we were using `np.long`, but we try to avoid the builtin `int` as much as possible (although it makes no difference for the runtime system code, I suppose).

[...]

PR #1283, opened 2021-03-21
Pyopencl typedeprecation issue1281

EvilMav:
Fix for issue #1281: replaced deprecated NumPy types: np.bool => bool (built-in) and np.long => np.int64.

### DIFF
--- a/src/Futhark/CodeGen/Backends/PyOpenCL.hs
+++ b/src/Futhark/CodeGen/Backends/PyOpenCL.hs
@@ -185,7 +185,7 @@ compileProg mode class_name prog = do
 -- We have many casts to 'long', because PyOpenCL may get confused at
 -- the 32-bit numbers that ImpCode uses for offsets and the like.
 asLong :: PyExp -> PyExp
-asLong x = Py.simpleCall "np.long" [x]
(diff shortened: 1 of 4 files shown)


src/Futhark/CodeGen/Backends/PyOpenCL.hs
```diff
@@ -185,7 +185,7 @@ compileProg mode class_name prog = do
 -- We have many casts to 'long', because PyOpenCL may get confused at
 -- the 32-bit numbers that ImpCode uses for offsets and the like.
 asLong :: PyExp -> PyExp
-asLong x = Py.simpleCall "np.long" [x]
```
athas, line 188: This may be one case where we do need to use `int` instead, since this function is used for values that end up getting passed to PyOpenCL, which may not like Numpy types.  We'll see what the CI says.

athas:
Thanks!  I'll merge this if/when the CI succeeds.

src/Futhark/CodeGen/Backends/PyOpenCL.hs
EvilMav: I've tried both and both seems to have passed the tests on my machine.

PR #1287, opened 2021-03-25
Adding a prototype of reduce to the MPI backend

michael-elkh:
Right now, the prototype only works if the size of the input array is a multiple of the number of MPI nodes.


athas:
I just gave you commit access so that you can merge into the `mpi` branch yourself (I'm not doing anything with it).

This is pretty cool!  I thought you'd need to do more harm to the runtime/startup code of Futhark programs before this would work.

PR #1313, opened 2021-04-15
Output size-comparison logging information on CUDA

### LINKED ISSUE #1312: Autotuner cannot handle futhark-benchmarks/accelerate/fft/fft.fut
It just rams all the thresholds to the maximum, which is much worse than the defaults.

Munksgaard:
Fixes #1312

### DIFF
--- a/src/Futhark/CodeGen/Backends/CCUDA.hs
+++ b/src/Futhark/CodeGen/Backends/CCUDA.hs
@@ -15,7 +15,7 @@ import Control.Monad
 import Data.List (intercalate)
 import Data.Maybe (catMaybes)
 import Futhark.CodeGen.Backends.CCUDA.Boilerplate
-import Futhark.CodeGen.Backends.COpenCL.Boilerplate (commonOptions)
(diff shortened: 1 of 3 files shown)


athas:
I foresee 100% likelyhood that this bug or a similar one will be reintroduced in the future, due to the duplication.

Could you put the code for emitting the statement in the `Futhark.CodeGen.Backends.COpenCL.Boilerplate` module (or somewhere similar) where both the COpenCL and CUDA backends could use it?

Ideally also the PyOpenCL backend, but that might be too ambitious...

Munksgaard:
Sure.

Ideally, we should also test that the autotuner works, somehow... Any ideas?

athas:
I suggest we add a `tests_adhoc` directory where we can put completely arbitrary scripts that test whatever we want.  That would also allow us to write tests for `futhark imports` and other random things.

Munksgaard:
> I suggest we add a `tests_adhoc` directory where we can put completely arbitrary scripts that test whatever we want. That would also allow us to write tests for `futhark imports` and other random things.

Sounds like a plan, at least for now, but perhaps that should be a separate PR.

PR #1317, opened 2021-04-16
Only generate those threshold parameters that are actually used.

athas:
There might still be gaps in the parameter numbering due to pruning
done during memory expansion.


athas:
This will probably require us to retune some benchmarks.  I'll do that as the errors come in!

PR #1323, opened 2021-04-21
Add option for compile free test runs

philass:
One of my main work flows that I need while developing the WASM/JS backend is the following
- Run a futhark test program
an example
`futhark test  example1.fut  --backend=wasm --runner=node --notty`

Then because my code is still quite raw, I fail the test. I then want to directly edit the executable `./example1` and run it through the test session again. This way I can quickly check which changes I need to make to the Javascript generation without having to recompile futhark everytime I make a minor tweek to get closer to the desired implementation.

As far as I can tell there is no way to run through the test suite again without having to recompile the source. Where all I really want to do it pass my edited executable back into the test session.

I have added a flag '-r' that operates like no flag, when there is no executable present. However when there is an executable present it will not recompile. But simply run it through the test session.

I though this might be helpful to others so I made a PR.

With regards to the implementation it may violate TOCTOU 

```
239  RunCases ios structures warnings -> do
240       exists <- liftIO $ doesFileExist $ "." </> dropExtension program
241       if mode == Run && exists
242         then runCompiledProgram (FutharkExe futhark) program progs ios
```

I think I can get rid of this issue without too much hassle, but I just wanted to check if there was interest in getting this into the compiler before I spend more time on this.

### DIFF
--- a/src/Futhark/CLI/Test.hs
+++ b/src/Futhark/CLI/Test.hs
@@ -27,6 +27,7 @@ import Futhark.Util.Pretty (prettyText)
 import Futhark.Util.Table
 import System.Console.ANSI
 import qualified System.Console.Terminal.Size as Terminal
+import System.Directory
 import System.Environment
 import System.Exit
 import System.FilePath
@@ -235,6 +236,11 @@ runTestCase (TestCase mode program testcase progs) = do
           ExitSuccess -> return ()
           ExitFailure 127 -> throwError $ progNotFound $ T.pack futhark
           ExitFailure _ -> throwError $ T.decodeUtf8 err
+    RunCases ios _ _ | mode == SkipCompile -> do
+      exists <- liftIO $ doesFileExist $ binaryName program
+      if exists
+        then runCompiledProgram (FutharkExe futhark) program progs ios
+        else throwError $ T.pack (binaryName program) <> " does not exist, but --skip-compilation passed.\n"
     RunCases ios structures warnings -> do
       -- Compile up-front and reuse same executable for several entry points.
       let backend = configBackend progs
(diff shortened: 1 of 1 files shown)


athas:
*If* we want this, then it should be called `--skip-compilation` like `futhark bench`.  But I would like to see at least one use case that isn't solely useful for compiler hackers.

philass:
I could probably come up with some super obscure use case, But yeah this was kind of intended purely for compiler hacking. I've found it super helpful, and its dramatically reduced my iteration cycles. But if I'm the only person who gets utility out of it then I respect the decision not to add it. 

Out of curiosity, does some of the reluctance come from the reasons for the `Use with caution` warning on the `bench --skip-compilation`

athas:
The simplest way to avoid bugs is to avoid features.  This is why the compilers take so few flags.  But I think `--skip-compilation` on `futhark test` is useful when adding new test cases without modifying the program.

athas:
You're getting into the weeds here.  `--skip-compilation` should not be a `TestMode`, because that is not composable (does it run the interpreter?).  Instead it should be a field in the `TestConfig` that is checked before compiling the programs.

philass:
@athas adding it is an option to `TestConfig` makes sense. The only issue is that the compilation step is done in the function `runTestCase`, which only takes `TestCase` as an argument. 

`TestCase` consists is defined as follows
```{
TestCase = TestCase
   { _testCaseMode :: TestMode,
     testCaseProgram :: FilePath,
     testCaseTest :: ProgramTest,
     _testCasePrograms :: ProgConfig
   }
```

The issue is that it doesn't contain `TestConfig` which would contain `configSkipCompile`. 

The easiest implementation would be to add `SkipCompile` as a field to `ProgConfig`, in which case minimal tweaking is necessary.

Without doing this it seems that multiple functions would need to be adapted to carry around `TestConfig` instead of just `ProgConfig`.

I don't know if you have any opinions on this. Also feel free to drop this if its becoming more trouble than its worth.

athas:
I think this is more trouble than it's worth *for now*, at least.  `futhark test` has been overdue for refactoring for a while.

PR #1338, opened 2021-05-10
Handle failing sample runs correctly

### LINKED ISSUE #1336: Autotuner bug
As described [here](https://github.com/mavion/annf/blob/47ed643f4f4c79ba193284fcd55fc1caee4f173f/README.md), `futhark autotune` will produce wrong autotuning results if all datasets cause the tuning runs to fail (eg. by running out of memory). 

Thanks to Kristian Bøjer Andreasen for pointing this out.

Munksgaard:
If the sampling run fails, we should truly return "no information". Currently,
we return `(tMin, tMax)`, but if this is the first dataset we're tuning, and the
sample run fails, we should return `Nothing`. Then, if the collective result of
tuning a threshold is `Nothing`, we should use `thresholdMin` to select the
alternative code version. Hopefully, that won't fail.

Fixes #1336

### DIFF
--- a/src/Futhark/CLI/Autotune.hs
+++ b/src/Futhark/CLI/Autotune.hs
@@ -231,17 +231,21 @@ tuneThreshold ::
   (String, Path) ->
   IO Path
 tuneThreshold opts datasets already_tuned (v, _v_path) = do
-  (_, threshold) <-
-    foldM tuneDataset (thresholdMin, thresholdMax) datasets
-  return $ (v, threshold) : already_tuned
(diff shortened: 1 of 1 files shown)


athas:
This is the kind of time where I wish we had gotten around to writing a test harness for the autotuner!

Munksgaard:
I agree...

@futbot: autotune

Munksgaard:
Oh damn, that's somebody's actual user. Sorry @futbot.

PR #1339, opened 2021-05-10
Add test-autotuner.sh

Munksgaard:


Munksgaard:
At least this should make it easier for us to verify that the autotuner works correctly by manually running it on gpu04 or something similar.

PR #1343, opened 2021-05-11
Build haddock documentation as part of CI tests

Munksgaard:


Munksgaard:
An alternative (or perhaps additional) measure would be to add "-haddock" as a ghc-option, as detailed [here](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/using.html?highlight=haddock#ghc-flag--haddock). What do you think @athas?

athas:
I have no opinion.

PR #1349, opened 2021-05-15
Allow FutharkScript-generated test input.

athas:
Two main problems remain:

1. FutharkScript is probably too limited for this to be truly useful.
   In particular, there is no variable binding or destructuring.  This is
   at least easy to fix.

2. There are awkward corner cases with entry points that accepts
   tuples (which become opaque types that _look_ like tuples) and
   return tuples (which are decomposed into their fields).  Probably
   the C API should expose tuples and records more intelligently, but
   this will not happen in the short term.


athas:
I added simple `let` bindings.  It works really well now (although point 2 still stands).

PR #1353, opened 2021-05-21
Fix #1352

### LINKED ISSUE #1352: Crashes in the CUDA backend crash futhark server

Munksgaard:


athas:
Since the test suite doesn't run the CUDA backend, do you solemly swear that you have run this on actual NVIDIA® CUDA® hardware?

PR #1375, opened 2021-06-12
Cuda code-generation of chained-scan with decoupled lookback

mortentc:
As per my Master's Thesis "Regular Segmented Single-pass Scan in Futhark", this PR implements code generation of a single-pass segmented scan in CUDA in [SegSinglePass.hs](https://github.com/mortentc/futhark/blob/master/src/Futhark/CodeGen/ImpGen/Kernels/SegScan/SegSinglePass.hs).
The current code-generator still defaults to Persson & Nicolaisen's implementation when the [collection to be scanned is not segmented](https://github.com/mortentc/futhark/blob/e695d835a7835cf487b6ffa9e88cbc34c7c89fc3/src/Futhark/CodeGen/ImpGen/Kernels/SegScan.hs#L65-L68), which avoids unnecessarry index calculations, but does not make use of the [analytical memory model for sequential work](https://github.com/mortentc/futhark/blob/e695d835a7835cf487b6ffa9e88cbc34c7c89fc3/src/Futhark/CodeGen/ImpGen/Kernels/SegScan/SegSinglePass.hs#L115-L132).


athas:
How difficult do you think it would be to incorporate the analytical model in the non-segmented scan?

athas:
This is solid work thanks!  I think the conflict is something fairly trivial (I'll take a look).

athas:
Oh, the conflict is only for rebasing; no worry then.

athas:
Could you please fix the style errors, though?  [All Futhark code is formatted with Ormolu.](https://github.com/diku-dk/futhark/blob/master/STYLE.md#ormolu)

mortentc:
> How difficult do you think it would be to incorporate the analytical model in the non-segmented scan?

There should be no difficulties, in the non-segmented scan the degree is [hardcoded](https://github.com/mortentc/futhark/blob/e695d835a7835cf487b6ffa9e88cbc34c7c89fc3/src/Futhark/CodeGen/ImpGen/Kernels/SegScan/SinglePass.hs#L117), so replacing the line 'm=9' with the [analytical model](https://github.com/mortentc/futhark/blob/e695d835a7835cf487b6ffa9e88cbc34c7c89fc3/src/Futhark/CodeGen/ImpGen/Kernels/SegScan/SegSinglePass.hs#L115-L132) should be enough.

> Could you please fix the style errors, though? All Futhark code is formatted with Ormolu.
I will take a look at the style, apologies.

athas:
Thanks, excellent work!

PR #1377, opened 2021-06-14
Shrink SegRed memory usage.

athas:


athas:
@mortvest This is the patch that will (hopefully) fix your issue.

mortvest:
> @mortvest This is the patch that will (hopefully) fix your issue.

It did. Thanks

PR #1379, opened 2021-06-15
Conceptually rename "lore" to "representation".

athas:
Specific changes:

* Decorations is now RepTypes.

* ASTLore is now ASTRep.

* The 'lore' type parameter is now generally named 'rep'.

Various other small follow-on renames.  In some cases, the "Lore"
suffix becomes "Dec" for "decoration".


zfnmxt:
Definitely a better name.

PR #1389, opened 2021-07-06
Add --compile-opencl-server to futhark dev

Munksgaard:


athas:
I've long thought we should actually expose the real actions (including running the C compiler) to `futhark dev`.

Munksgaard:
Yeah, that would be great. This particular flag is helpful for me to be able to run futhark bench on manually modified IRs.

PR #1391, opened 2021-07-08
Merge master into wasm-multicore.

athas:
Note: this removes the `terminate` stuff from `futhark test`, as the server client code is now in a separate library.  Another solution will have to be found.  Maybe don't merge this until after your defense.


philass:
Thanks for helping with the merge!

I'll wait to merge until after the defence.

Deleting the terminate, shouldn't be an issue. The addition of terminate stopped Node from hanging, in such a way that it can continue with the next test. It unfortunately didn't terminate the way I would have liked. Running 'ps' shows that all the node sessions have not actually been killed, and still exist in the background. I imagine this is why 'test wasm-multicore' mysteriously fails after 100+ tests (Github actions might have a limit for background processes). So scrapping it all together is no bother


Like we talked about earlier, the pthread hang in Node is a known [issue](https://github.com/emscripten-core/emscripten/issues/9763) in Emscripten, but it seems that there is a [PR](https://github.com/emscripten-core/emscripten/pull/12963) that just landed, and may actually fix the problem. 

I'll look into that after I'm done with the longest hour of my life

athas:
I also found out why you cannot run multiple tests concurrently.  I'll add the fix to this PR.

src/Futhark/Actions.hs
```diff
@@ -154,15 +154,15 @@ runCC cpath outpath cflags_def ldflags = do
...
+runEMCC :: String -> String -> FilePath -> [String] -> [String] -> [String] -> Bool -> Bool -> FutharkM ()
+runEMCC cpath outpath classpath cflags_def ldflags expfuns rts lib = do
   ret <-
     liftIO $
       runProgramWithExitCode
         "emcc"
         ( [cpath, "-o", outpath]
             ++ ["-lnodefs.js"] -- "-s", "INITIAL_MEMORY=201326592"] -- ,"-s", "ALLOW_MEMORY_GROWTH=1"]
-            ++ (if rts then ["-s", "--post-js", "futharkClass.js"] else [])
+            ++ (if rts then ["-s", "--post-js", classpath] else [])
```
philass, line 165: This file `classpath` isn't needed after this statement. Would you recommend deleting it or do you think its helpful to have an unused file that only contains the API laying around?

athas: I was wondering about that myself.  In that case I would actually suggest making it a temporary file.  However, the Futhark compiler already leaves intermediate build artifacts in the directory, so it's not crucial to remove it.

philass:
Also there is a lot of greasy code that needs cleanup,

There are comments that are no longer necessary littered everywhere, unused functions, mixed syntax (I still need to change some of the Haskell to  use quasi quotes instead of unlines)... and so much more that would drive somebody mad.    I haven't forgotten about them, they have just been on the back burner. I'll attack these very soon.

athas:
I'll do a proper code review once you propose this for merging.  That'll also require writing up some documentation.

athas:
I fixed the issue that made the wasm-multicore tests never terminate.  I think it is safe for you to merge this.

PR #1392, opened 2021-07-11
Remove "context" part of patterns.

athas:
This is a significant simplification.  Now, if something returns N
values, its type list will have N entries, always. Probably we will be
chasing the dividends of this simplification for years, as the
context/value distinction was truly entrenched.

There are some programs that now crash the compiler.  The main ones I
noticed are related to being unable to copy-propagate
certificate-protected variables into Results.  This should be fixed
separately, and this commit rebased.

Entry points annotations (and some parts of the ExplicitAllocations
pass) still assume that the "context parts" (sizes) of a pattern are
first, but that is an easy property to maintain - and we might relax
it in the future.


athas:
Based on a quick check, this also seems to make the compiler quite a bit faster (about 1.2x for heston32).

PR #1394, opened 2021-07-12
Add room for WASM documentation.

athas:
This commit creates the files and sets up the links.  The manpages are proper drafts, but the actual API spec is just a placeholder.


athas:
This PR also merges in some more changes from master.

PR #1398, opened 2021-07-18
Replace trace/break with attributes.

### LINKED ISSUE #1355: Problems with trace and break
Neither **trace** or **break** work as intended with  <|, |>, <-<  or >-> operators.
- **trace** does not print the correct line/char number. Prints /prelude/functional.fut:.... [instead](https://gist.github.com/mortvest/3950e06b0ef3a6236345fdd3e5dbe724)
- **break** behaves very strangely. Breakpoint is recognized, but with a scope consisting of a single variable "x" (even when there are no variables called "x" in the entire [program](https://gist.github.com/mortvest/091b715e424b58a10597de238a70f839)).

Moreover, **break** is ignored when it is a part of a let-binding in a [program](https://gist.github.com/mortvest/edfa9e1e0f514f065aba7811383c6b16). Here, if you just load the file, nothing would happen.

### LINKED ISSUE #1357: Add support for trace in C backend
While tracing in OpenCL might be difficult to do, it should be pretty straightforward in the C backend, right?

Perhaps we could even add support for breaks, by hooking up to GDB?

athas:
Closes #1357.

Closes #1355.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+  * The `#[trace]` and `#[break]` attributes now replace the `trace`
+    and `break` functions (although they are still present in
+    slightly-reduced but compatible form).
+
+  * Tracing now works in compiled code, albeit with several caveats
+    (mainly, it does not work for code running on the GPU).
+
 ### Removed
 
 ### Changed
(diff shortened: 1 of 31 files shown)


athas:
Currently ignores attempts to trace arrays in compiled code.  I'll fix that.

athas:
Now also supports tracing arrays, although it just prints the elements on a single line separated by whitespace, without any separators.

I also think we will generate invalid GPU code if a trace operation happens to find its way inside a kernel.  I'll add some filtering.

FluxusMagna:
For the sake of consistency, perhaps `opaque` should be replaced by an attribute as well?

athas:
Yes, I'll do that in a separate change.  It's not as urgent, because `opaque` works fine, while `trace` and `break` had horrible hacks in the interpreter to provide the stack trace a human would expect.

PR #1399, opened 2021-07-19
Make the WASM backends work with benchmarking.

athas:


philass:
This LGTM. Thanks for the PR

philass:
Looks like CI fail is unrelated so I'll go ahead and merge.

PR #1400, opened 2021-07-19
Make 'futhark dev' use the correct representation names.

athas:
So:

  * '--gpu' is what used to be '--kernels'
  * '--gpu-mem' is what used to be '--gpu'
  * '--mc-mem' is what used to be '--multicore'
  * '--seq-mem' is what used to be '--cpu'.

For input file names:

  * '.fut_gpu' is what used to be '_fut_kernels'
  * '.fut_gpu_mem' is what used to be '.fut_kernels_mem'

@munksgaard and @coancea: you will care about this.

Time to update years of muscle memory!

### DIFF
--- a/src/Futhark/CLI/Dev.hs
+++ b/src/Futhark/CLI/Dev.hs
@@ -543,7 +543,7 @@ commandLineOptions =
       "Run the default optimised kernels pipeline"
       kernelsPipeline
       []
-      ["kernels"],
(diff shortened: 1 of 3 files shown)


Munksgaard:
Excellent! I've been thinking about what naming scheme would be best for this, but haven't been able to come up with anything nice. I like yours.

PR #1413, opened 2021-07-27
Bump benchmarks to include lud fix

Munksgaard:
@athas: How did we not catch this error before? Shouldn't buildbot system run
the benchmarks and see that the 16by16.in input to lud was invalid?


Munksgaard:
Also, what is [that error](https://github.com/diku-dk/futhark/pull/1413/checks?check_run_id=3170092257)?

athas:
> Also, what is [that error](https://github.com/diku-dk/futhark/pull/1413/checks?check_run_id=3170092257)?

My guess is a race condition in the CI cache management.

athas:
> @athas: How did we not catch this error before? Shouldn't buildbot system run
> the benchmarks and see that the 16by16.in input to lud was invalid?

Invalid how?  The C value reader is somewhat more forgiving than the Haskell implementation.

Munksgaard:
> Invalid how? The C value reader is somewhat more forgiving than the Haskell implementation.

Running `futhark test --backend=opencl lud.fut` failed for me before this commit.

PR #1416, opened 2021-07-29
Improve interference tracking of loop memory parameters.

athas:
We conservatively treat all memory arguments to a DoLoop to interfere
with each other, as well as anything used inside the loop.  This could
potentially be improved by looking at the interference computed by the
loop body wrt. the loop arguments, but probably very few programs
would benefit from this.


athas:
Do we really only have a single test (`tests/psum.fut`) for this pass?

Munksgaard:
Yes...

Munksgaard (approved):

PR #1423, opened 2021-08-03
Enforce Data.Map and Data.Set imports as M and S, respectively

Munksgaard:


Munksgaard:
We could add more restrictions. Inspiration [here](https://github.com/ndmitchell/hlint/blob/master/.hlint.yaml#L10-L32)

Issue #1402, opened 2021-07-19
The API generated by the WASM backends is unsafe


athas:
The API requires manual memory management, which is unavoidable because JavaScript does not support finalisers (wtf!).  However, if the user happens to use an object after it has been "freed", we will end up reading garbage memory (albeit still sandboxed), and this *is* avoidable.  The proxy classes FutharkArray and FutharkOpaque should contain a "valid" field that is set to false when they are freed, and which we check before using them.  This will allow us to throw an exception instead of corrupting memory.

philass:
Can we close this or is there still more to do here?

athas:
Closed in #1427.  Let's hope we don't have to reopen it.

kfl:
@athas FYI JavaScript *does* support finalizers. The [ECMAScript spec for `FinalizationRegistry`](https://tc39.es/ecma262/multipage/managing-memory.html#sec-finalization-registry-objects) and the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry). And if it's really important to support old IE, I guess that you could roll your own with [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap).

PR #1427, opened 2021-08-04
Wasm free check

philass:
This PR addresses #1402.


athas:
It would be nice to eventually have some tests that check that we catch this, but it's not like the C backend has a lot of negative tests either.

philass:
It did occur to me. `tests_lib` doesn't have any JavaScript/WASM tests. 

I'm happy to populate the `tests_lib` with some JavaScript examples. 

Would you like me to add these validity tests with this PR or make a new PR with more comprehensive tests of the JS/WASM backend as a library?

athas:
That's better done in a separate commit.

PR #1439, opened 2021-08-11
Initial setup for WASM/JS Node library tests

philass:
Finally got a test set up sort of working with ES6, for library tests with Node.

Let me know if this setup is the kind of thing you had in mind. If so I'll start populating with tests that tests meaningful functionality.

### DIFF
--- a/tests_lib/javascript/test_a.js
+++ b/tests_lib/javascript/test_a.js
@@ -1,0 +1,20 @@
+// Tests for a.fut, the increment function
+
+import assert from 'assert/strict';
+
+// Hack for Running generated ES6 modules from Emscripten with Node
+// https://github.com/emscripten-core/emscripten/issues/11792#issuecomment-877120580
+import {dirname} from "path";
+globalThis.__dirname = dirname(import.meta.url).substring(7);
+import { createRequire } from 'module';
+globalThis.require = createRequire(import.meta.url);
+
+// Imports from the generated ES6 Module
+import { newFutharkContext } from './a.mjs';
+
+var fc;
+newFutharkContext().then(x => {
+  fc = x;
+  var res = fc.incr(7)
+  assert(res == 8);
+});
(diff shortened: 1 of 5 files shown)


athas:
Does the package.json thing also work if we add more `.js` files?  To see whether this actually works:

1. Add another test program.
2. Add it to the CI.

You should put in a `test.sh` script that runs the tests, just so it has the same interface as the other test directories.

philass:
The package.json works for multiple `.js` files. I just checked with another test program. 

```
$ ls
Makefile     a.fut        b.fut        package.json test_a.js    test_b.js
$ make
futhark wasm --library a.fut
node --experimental-wasm-simd test_a.js
futhark wasm --library b.fut
node --experimental-wasm-simd test_b.js
rm b.mjs a.mjs
```

I'll get on `2` right away.

tests_lib/javascript/test_a.js
```diff
@@ -0,0 +1,20 @@
+// Tests for a.fut, the increment function
+
+import assert from 'assert/strict';
+
+// Hack for Running generated ES6 modules from Emscripten with Node
+// https://github.com/emscripten-core/emscripten/issues/11792#issuecomment-877120580
+import {dirname} from "path";
+globalThis.__dirname = dirname(import.meta.url).substring(7);
```
philass, line 8: The real grease is the substring(7) at the end of this line. Without this (file://) is weirdly added to the front of the `__dirname` making it unable to load the `.wasm` module.

philass:
After some more investigation, we don't actually need `package.json`. As long as we name the test files with `.mjs` we can omit the `package.json` all together.

philass:
The `tests_lib` have a different interface than the other test directories. For the CI both the `C` and `Python`, `tests_lib` directories are called directly with `Make` in the CI.

If that is an oversight, I can also quickly whip up a `test.sh` for the C and Python directories and put them in the CI

athas:
No that's fine, I just forgot.  Make it work like the current lib tests.

Issue #1406, opened 2021-07-20
WebAssembly tests for the browser


philass:
Right now we currently only test the WebAssembly backends with Node locally.

It would make sense to also add some tests that run in the browser. We should look at headless Chrome and Firefox and see if there is a way we can use them for writing some tests that run with github actions.

ruler501:
Testing with the worker environment would probably also be good. The webasm code I use in my projects is compiled aas either `ENVIRONMENT=node,worker` or `ENVIRONMENT=browser,worker`.

philass:
There are still more tests to be added. But I am comfortable closing this as we now have an example test running the WASM/Javascript backend in the browser.

PR #1446, opened 2021-08-15
Initial web tests

philass:
This is what I managed to whip up for the web tests, to address #1406.

I'll try and add some more tests over time.

Any feedback would be appreciated. I'm still pretty green with respect to Node and web testing.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -392,6 +392,10 @@ jobs:
       with:
         node-version: '16.x'
 
+    - name: Install Jest
+      run: |
+        npm install jest --global
+
     - uses: actions/download-artifact@v2
       with:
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -407,7 +411,8 @@ jobs:
         node --version
         export EMCFLAGS="-sINITIAL_MEMORY=2147418112 -O1" # 2gb - 64kb... largest value of memory
         futhark test -c --backend=wasm --runner=./tools/node-simd.sh --no-tuning --exclude=no_wasm tests examples
-        make -C tests_lib/javascript -j
+        make -C tests_lib/javascript
+        make -C tests_lib/web
```
athas, line 415: Put this in a separate action.

philass: Will do!

PR #1450, opened 2021-08-18
Closes #1449.

### LINKED ISSUE #1449: OpenCL error : Program uses double-precision floats
I was doing some benchmarks from `futhark-benchmarks`. And I ran into the following issue.

For the following program `scan.fut`

```
entry sum_iota_f64 n = i64.i32 n |> iota |> map f64.i64 |> scan (+) 0
```
I compile and run with
```
futhark opencl scan.fut 
./scan: Program uses double-precision floats, but this is not supported on the chosen device: Intel(R) UHD Graphics 630
```

I am running this on MacOS.

philass:
this commit blacklists `Intel(R) UHD Graphics 630` addressing #1449.

### DIFF
--- a/rts/c/opencl.h
+++ b/rts/c/opencl.h
@@ -415,7 +415,8 @@ static int is_blacklisted(const char *platform_name, const char *device_name,
       strcmp(cfg->preferred_device, "") != 0) {
     return 0;
   } else if (strstr(platform_name, "Apple") != NULL &&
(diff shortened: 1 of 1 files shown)


athas:
The Intel GPU is fine; it just doesn't support double precision.  This is also an incomplete blacklist, because most Intel GPUs do not support double precision.  I do not want to add more invisible magic to device selection.  The existing blacklist is only because Apple's CPU device is *spectacularly* broken.

PR #1483, opened 2021-10-11
Move to aeson-2.

athas:


athas:
This cannot actually be merged yet, because Nixpkgs hasn't caught up.  But the actual compiler changes are minor.

Issue #1542, opened 2021-12-06
Allow loading datasets in `futhark literate`


Snektron:
Currently, `futhark literate` can load images from the file system using `$loadimg`. It would be useful if there was an option to load arbitrary futhark datasets, for example using something like `$loaddata "file"`.

athas:
This would be a good feature.  Similarly, generating random data would also be useful.

athas:
Implemented in #1544.

PR #1544, opened 2021-12-06
literate: loaddata builtin

Snektron:
Implements #1542 (just the data loading part)

As discussed on Matrix, when a data file contains multiple values an tuple of values is returned, which must be destructered for use as FutharkScript does not support applying functions with tuples:
```futhark
--> let (a, b) = $loaddata "multiple.in" in foo a b
```
When there is only a single value a regular value is returned, and no destructuring is required:
```futhark
--> foo ($loaddata "single.in")
```

I also fixed up an incorrect error message when `$loadimg` is called with incorrect arguments.


athas:
What will this do if the provided file does not exist or otherwise cannot be read?

Snektron:
When a file does not exist, the compiler returns
`futhark: missing.in: openBinaryFile: does not exist (No such file or directory)`

and compilation is aborted. Note that this still generates the source file, executable, and images directory.

When the data file cannot be read:
```
> $loaddata "test.in" failed:
Failed to read datafile test.in
```
I guess here it mentions the filename twice with might be undesirable.

Another problem might be is that there is no checking for what the type of the parameter of `$loaddata` is, and so

`-- > ($loaddata 1)`

yields 
```
futhark: : openBinaryFile: does not exist (No such file or directory)
```

athas:
The former is OK (I was worried it might print Internal Compiler Error or some such), but the latter should be fixed.  I don't think it would be very difficult.

PR #1558, opened 2022-01-03
 Issue a warning if non-monotonicity is detected while autotuning. 

### LINKED ISSUE #1554: Autotune makes my code benchmark execute slower
Hi

I have run this with benchmark opencl both before and after using autotune, and on gpu03 I have noticed the general execution has slowed down after using autotune.

```
let argmax arr =
  reduce_comm (\(a,i) (b,j) ->
                 if a < b
                 then (b,j)
                 else if b < a then (a,i)
                 else if j < i then (b,j)
                 else (a,i))
              (0f32, 0)
              (zip arr (indices arr))

let gaussian_elimination [n] [m] (A: [m][n]f32): [m][n]f32 =
  loop A for i < i64.min m n do
    -- Find nonzero pivot.
    -- Written carefully to avoid irregular parallelism.
    let value j x = if j >= i then f32.abs x else -f32.inf
    let j = A[:,i] |> map2 value (indices A) |> argmax |> (.1)
    let f = (1-A[i,i]) / A[j,i]
    let irow = map2 (f32.fma f) A[j] A[i]
    in tabulate m (\j ->
                     let f = A[j,i] * -1
                     in map2 (\x y -> if j == i then x else f32.fma f x y)
                             irow A[j])

let matrix_inverse [n] (A: [n][n]f32): [n][n]f32 =
  let n2        = n + n
  let augmented = map (\ i1 -> 
                    map (\ i2 -> 
                      if i1+n == i2
                      then 1f32
                      else if i2 < n
                           then A[i1,i2]
                           else 0f32) 
                              (iota n2 )) 
                                (iota n)
  let changed = gaussian_elimination augmented
  let inverse = map (\arr -> take n (reverse arr) |> reverse) changed
  in inverse


let main [k][n] (A: [k][n][n]f32): [k][n][n]f32 =
  map (\mat -> matrix_inverse mat) A


entry test_main = main
--
--
-- ==
-- entry: test_main
-- random input { [100000][2][2]f32 }
-- random input { [100000][4][4]f32 }
-- random input { [100000][6][6]f32 }
-- random input { [100000][8][8]f32 }
-- random input { [100000][10][10]f32 }
-- random input { [100000][12][12]f32 }
-- random input { [100000][14][14]f32 }
-- random input { [100000][16][16]f32 } 
```

Munksgaard:
Detects whether tuning a parameter decreases the performance on a given
dataset. Ideally, this should never happen and (excepting bugs in the autotuner)
is a sign of non-monotonicity in the generated Futhark program.

Fixes #1554.

### DIFF
--- a/src/Futhark/CLI/Autotune.hs
+++ b/src/Futhark/CLI/Autotune.hs
@@ -312,8 +312,10 @@ tuneThreshold opts server datasets already_tuned (v, _v_path) = do
                 putStrLn $ unwords ("Got ePars: " : map show ePars)
 
               newMax <- binarySearch runner (t, tMax) ePars
-              let newMinIdx = pred <$> elemIndex newMax ePars
-              let newMin = maxinum $ catMaybes [Just tMin, newMinIdx]
+              let newMinIdx = do
+                    i <- pred <$> elemIndex newMax ePars
+                    if i < 0 then fail "Invalid lower index" else return i
+              let newMin = maxinum $ catMaybes [Just tMin, fmap (ePars !!) newMinIdx]
               return $ Just (newMin, newMax)
 
     bestPair :: [(Int, Int)] -> (Int, Int)
--- a/src/Futhark/CLI/Autotune.hs
+++ b/src/Futhark/CLI/Autotune.hs
@@ -8,6 +8,7 @@ import Control.Monad
 import qualified Data.ByteString.Char8 as SBS
 import Data.Function (on)
 import Data.List (elemIndex, intersect, isPrefixOf, minimumBy, sort, sortOn)
+import qualified Data.Map as M
 import Data.Maybe
 import qualified Data.Set as S
 import qualified Data.Text as T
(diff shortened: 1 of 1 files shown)


Munksgaard:
I've also included a fix to an actual bug in the autotuner, where we didn't correctly narrow the threshold range used for iterative tuning. Perhaps we should re-tune the benchmarks after merging this, just to see if it makes a difference.

src/Futhark/CLI/Autotune.hs
```diff
@@ -388,8 +414,12 @@ tune opts prog = do
   let progbin = "." </> dropExtension prog
   putStrLn $ "Running with options: " ++ unwords (serverOptions opts)
 
-  withServer (futharkServerCfg progbin (serverOptions opts)) $ \server ->
-    foldM (tuneThreshold opts server datasets) [] $ tuningPaths forest
+  fst
```
athas, line 417: This looks terrible; can you please use either `fmap fst . withServer` or put the `fst` inside the lambda?

PR #1574, opened 2022-01-27
More precise error codes.

### LINKED ISSUE #1570: Better handling of out-of-memory on GPU backends
By "better handling" I don't mean anything more fancy than explicitly saying that's the problem, what the memory consumption is of the failing program, and the size of the failed allocation.

(This only goes for device OOM; `malloc()` failures cannot be handled sanely by Futhark I think.)

athas:
Also more precise OOM errors.

Closes #1570.


athas:
@FluxusMagna I hope you can quickly add support for reporting OOM in futhask!  Perhaps you even have a program where this has been a problem?  I have tested that the detection works (at least in simple cases), but it's difficult to test these error paths systematically.

FluxusMagna:
That should be pretty easy. I will try it this weekend if I have time. There is already code for triggering garbage collection on errors, but it currently triggers for all errors once, so I just need to specialize that to the new error code. I have several futhark functions that will run out of memory quite easily, so that part is a non-issue.

PR #1575, opened 2022-01-28
Virtualise intragroup SegOps

athas:
These improvements to the code generator lets us have intragroup kernels with any amount of inner parallelism (although any use of local memory must still fit), by the use of virtualisation.

This is still not used by flattening however, as performance suffers sometimes even when virtualisation is not needed (10% on LocVolCalib for example).  This could be addressed by generating two kernels (one virtualised and one not) during code generation and picking the best one.  I shudder at the thought of even more code explosion, though.  We'll see what we do with this.


athas:
I think the errors are transient and due to infrastructure issues elsewhere.

PR #1576, opened 2022-01-28
Adding DPP project enhancements to reverse-AD ('ad-genred-opt' branch)

Nephz:
Includes enhancements for reverse-AD:
* reduce (*)
* reduce(map2(binop)) 2d_arr interchange 

After the oral part of the project we extended the interchange to work for all binops as suggested by you.


athas:
Closing due to staleness because ad-genred-opt has been merged into clean-ad.  @coancea will have to merge this some other way.

PR #1588, opened 2022-02-04
Refactor error handling in generated host code.

### LINKED ISSUE #1584: Shrink code bloat due to error handling
The C code generator must take care to avoid leaking memory during program failures, but the way it currently does it is very stupid and leads to code bloat.  This should be improved.

athas:
Previously, every error point would contain code to free every memory
block in scope.  This caused significant code duplication.  Now they
jump to a shared function epilogue that frees all memory blocks.

Closes #1584.


athas:
I saw a large program where this cuts compile times by 75%.

PR #1594, opened 2022-02-09
Prohibit newlines in string literals

Munksgaard:
As per the reference, newlines in string literals should not be allowed.

### DIFF
--- a/docs/language-reference.rst
+++ b/docs/language-reference.rst
@@ -197,8 +197,9 @@ first class.  See :ref:`hofs` for the details.
 
 .. productionlist::
    stringlit: '"' `stringchar`* '"'
-   charlit: "'" `stringchar` "'"
-   stringchar: <any source character except "\" or newline or quotes>
+   stringchar: <any source character except "\" or newline or double quotes>
+   charlit: "'" `char` "'"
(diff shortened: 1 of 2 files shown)


Munksgaard:
@athas How/where do I create a test for this?

athas:
Negative tests for parse errors is a hopeless effort.

athas:
Doesn't this change disallow `'"'`?

Munksgaard:
Actually, I think it disallows `"'"`.

The reference states:

```
stringlit  ::=  '"' [stringchar](https://futhark.readthedocs.io/en/stable/language-reference.html#grammar-token-stringchar)* '"'
charlit    ::=  "'" [stringchar](https://futhark.readthedocs.io/en/stable/language-reference.html#grammar-token-stringchar) "'"
stringchar ::=  <any source character except "\" or newline or quotes>
```

which might actually disallow both `'"'` and `"'"`? And escaped characters...

athas:
Please fix the specification.  We want to allow `"'"` and `'"'`.  We also want to allow escapes, but feel free to punt on that.  Currently the rule is something like "whatever Haskell does for string literals".

Munksgaard:
@athas: Something like this?

PR #1601, opened 2022-02-17
Add dynamic check to avoid copies

Munksgaard:


athas:
I think you need to check more than the offset.  Also, this definitely requires a comment justifying why it's valid and sensible.

athas (changes requested):
Check everything.

src/Futhark/CodeGen/ImpGen.hs
```diff
@@ -1372,14 +1372,37 @@ fullyIndexArray' (MemLoc mem _ ixfun) indices = do
...
+        ( fromBool (dst_name == src_name && length dst_lmads == 1 && length src_lmads == 1)
+            .&&. IxFun.lmadOffset dst_lmad .==. IxFun.lmadOffset src_lmad
+            .&&. foldl
+              ( \acc (dim1, dim2) ->
+                  acc
+                    .&&. IxFun.ldStride dim1 .==. IxFun.ldStride dim2
+                    .&&. IxFun.ldRotate dim1 .==. IxFun.ldRotate dim2
+                    .&&. IxFun.ldShape dim1 .==. IxFun.ldShape dim2
+                    .&&. (fromBool $ IxFun.ldPerm dim1 == IxFun.ldPerm dim2)
+                    .&&. (fromBool $ IxFun.ldMon dim1 == IxFun.ldMon dim2)
```
athas, line 1397: Could you factor this out as a separate function that, given two IxFuns, produces a `TPrimExp Bool` that dynamically tells whether the IxFuns are the same?  It could perhaps even live in the IxFun module itself (also for testing purposes).

Munksgaard: Of course. What's a good naming scheme for this? `dynamicEqualsLMADDim` is my best idea, but it's a bit of a mouthful...

athas: I think `equivLMAD` and `equivIxFun` are fine names!

PR #1603, opened 2022-02-22
Export Warnings for language server

haoranpb:
WIP links:
[futhark-language-server](https://github.com/haoranpb/futhark-language-server)
[vscode-futhark](https://github.com/haoranpb/vscode-futhark)


athas:
I think it is better not to have the "context" part. It's not useful in general.

haoranpb:
removed "context" and formatted, should pass ci now

not sure about the naming though

PR #1606, opened 2022-03-09
Added flake.nix and flake-compat for compatibility.

Alexnortung:
## Motivation for this change

This change replaces niv with nix flakes, such that installation of futhark for nix projects is much easier.
There is also a problem with nixpkgs where the latest version of futhark or one of its dependencies appears broken.

Migrating to flakes should not be a pain for other nix users who are not yet using flakes (since it is still in unstable), thus I have included flake compat such that the commands `nix-shell` and `nix-build` still has the same functionality.

## Features for nix flakes users

Installation is easy in other nix flakes by using `inputs.futhark.url = "github:diku-dk/futhark?dit=nix`
For flake users:
`nix develop ./nix` super fast dev shells
`nix build ./nix` or `nix build ./nix#futhark` to build either the release or the package itself.
`nix run ./nix` runs the current version of futhark that is being worked on.

inputs can still be updated easily just like with niv with e.g. the following command `nix flake lock --update-input nixpkgs`


athas:
Interesting idea (although CI currently fails).  But to merge this you'll also have to write some documentation about how to update the snapshots that are currently handled with `niv`.

Alexnortung:
I'm not sure I can fix the issue with `--argstr` which makes the ci fail. I will close this pull request and open a new once everything seems tidied up.

PR #1608, opened 2022-03-12
Loc info for parse Error

haoranpb:
The code compiles but the location info are still missing somehow, for example the code below will still emit `NoLoc`:

```futhark
def average (xs: []f64) = reduce (* 0 xs / f64.i64 (length xs)
```

Just managed to trace it back to [parseError](https://github.com/diku-dk/futhark/blob/f18ec76225b4c7c652d3b2148a382b9fbd4c5c36/src/Language/Futhark/Parser/Monad.hs#L241), `NoLoc` starts from there, will keep investigating


athas:
Strange.  To my eyes, it looks like EOF tokens have their location set properly.

haoranpb:
Should be good to go, except for the style issues @athas

PR #1611, opened 2022-03-19
Typed hole

haoranpb:
Accidentally deleted all changes made since our last meeting...

Make a PR to keep the progress


haoranpb:
@athas  Could you please take a look if I'm starting everything correctly? Can't remember how we started exactly anymore, thanks in advance!

athas:
With the design we discussed, the `HOLE` terminal also needs a `Name` field and the `Hole` AST constructor should have just a `vn`, not a `QualName vn`.  Further, the `PatType` should be wrapped in `f`.

haoranpb:
Got error `Unexpected token: '?'` when compiling

```futhark
def test = ?foo
```

I guess I forgot something in `lexer.x`?

athas:
Yes, you need to add a tokenization rule (a regular expression) for the new token.

athas:
Correction: The `Hole` AST constructor should name a `Name`, not a `vn`.

haoranpb:
Still got `NoInfo` for the type, but other than that should be rdy

src/Language/Futhark/TypeChecker/Terms.hs
```diff
@@ -269,6 +269,7 @@ checkApplyExp e = do
 checkExp :: UncheckedExp -> TermTypeM Exp
 checkExp (Literal val loc) =
   pure $ Literal val loc
+checkExp (Hole name t loc) = typeError loc mempty $ "Found typed hole: " <+> text (show name) <+> text (show t) <> "."
```
athas, line 272: Linebreak.  Also, don't bother printing the `t`, since it has no information at this point.

haoranpb:
After reading the [article](http://dev.stephendiehl.com/fun/006_hindley_milner.html) and some googling, I'm still not sure how to solve the error I'm getting.

I changed the code as discussed on Tuesday:

```futhark
checkExp (Hole name _ loc) = do
  t <- newTypeVar loc "t"
  pure $ Hole name (Info $ fromStruct t) loc
```

And I'm getting the error below, any tips @athas ?

```
  • Ambiguous type variable ‘as0’ arising from a use of ‘newTypeVar’
    prevents the constraint ‘(Monoid as0)’ from being solved.
    Probable fix: use a type annotation to specify what ‘as0’ should be.
    These potential instances exist:
      instance Ord k => Monoid (M.Map k v)
        -- Defined in ‘Data.Map.Internal’
      instance Ord a => Monoid (S.Set a)
        -- Defined in ‘Data.Set.Internal’
      instance Monoid Uniqueness -- Defined in ‘Language.Futhark.Core’
      ...plus 17 others
```

athas:
Don't use `fromStruct`.

haoranpb:
I probably spent too much time digging around by myself.

Removing `fromStruct ` worked, it compiled successfully and then  I'm a bit confused about what to do next, I went through file `Term.hs`, and I guess I need to at least change the function [checkFunBody](https://github.com/diku-dk/futhark/blob/5b2a4d5f5e73aff87b1db7c132d7f22148060f1b/src/Language/Futhark/TypeChecker/Terms.hs#L1707)? I tried to print out the result of `body'`, but I can't seem to find `Hole` in it

athas:
Why do you think you need to do something next?  What more do you think needs to be done in the type checker?  I see no reason why you should need to change `checkFunBody`.

haoranpb:
I'm testing it out locally and getting some strange behaviours at runtime. With an example futhark code like:

```futhark
def average (xs: []f64) = reduce (+) 0 ?foo / f64.i64 (length xs)
```

I'm getting an error `Size "n₅" is ambiguous.`

And if I move the typed hole a bit:

```futhark
def average (xs: []f64) = reduce (+) ?foo xs / f64.i64 (length xs)
```

the program compiles without any errors, and `futhark query` gets no info at `?foo`. 

I'm thinking about throwing errors (with type) for the Holes right before type checker finishes

athas:
> I'm testing it out locally and getting some strange behaviours at runtime. With an example futhark code like:
> 
> ```futhark
> def average (xs: []f64) = reduce (+) 0 ?foo / f64.i64 (length xs)
> ```
> 
> I'm getting an error `Size "n₅" is ambiguous.`

That is to be expected, since the size of the array `?foo` is supposed to represent cannot be known.  Typed holes do not change the fundamentals of type checking.  (That this particular type error message is not great is not your problem.)

athas:
> I'm testing it out locally and getting some strange behaviours at runtime. With an example futhark code like:
> 
> ```futhark
> def average (xs: []f64) = reduce (+) 0 ?foo / f64.i64 (length xs)
> ```
> 
> I'm getting an error `Size "n₅" is ambiguous.`

That is to be expected, since the size of the array `?foo` is supposed to represent cannot be known.  Typed holes do not change the fundamentals of type checking.  (That this particular type error message is not great is not your problem.)

> I'm thinking about throwing errors (with type) for the Holes right before type checker finishes

Don't do this.  Type checking a program with holes should finish without type errors.

athas:
I'm considering changing the syntax to just `???`, since I don't think we have any current use for the names.

PR #1618, opened 2022-03-29
Initialize Futhark Language server

haoranpb:
Initialize the Futhark Language server

Current features:
1. Hover Information
2. Error/Warning diagnostic

Know issue:
1. https://github.com/haoranpb/futhark-language-server/issues/1

Usage:
1. Clone this PR
2. Compile from source
3. Install the compiled binary in $PATH
4. Install the [extension](https://marketplace.visualstudio.com/items?itemName=DIKU-haoranpb.futhark-vscode) in vscode


src/Futhark/CLI/LSP.hs
```diff
@@ -0,0 +1,42 @@
...
+import Control.Monad.IO.Class (MonadIO (liftIO))
+import Futhark.LSP.Handlers (handlers)
+import Futhark.LSP.Utils (debug, emptyState)
+import Language.LSP.Server
+import Language.LSP.Types
+import System.Log.Logger (Priority (DEBUG))
+
+main :: String -> [String] -> IO ()
+main _prog _args = do
+  stateMVar <- newMVar emptyState
```
athas, line 16: We prefer `snake_case` for local non-function variables; see [the style guide](https://github.com/diku-dk/futhark/blob/master/STYLE.md#naming).

src/Futhark/CLI/LSP.hs
```diff
@@ -0,0 +1,42 @@
...
+{-# LANGUAGE OverloadedStrings #-}
+
+module Futhark.CLI.LSP (main) where
+
+import Control.Concurrent.MVar (newMVar)
+import Control.Monad.IO.Class (MonadIO (liftIO))
+import Futhark.LSP.Handlers (handlers)
+import Futhark.LSP.Utils (debug, emptyState)
+import Language.LSP.Server
+import Language.LSP.Types
```
athas, line 11: While we're normally OK with unqualified imports in the compiler, LSP exports such an enormous amount of names that it may be better to explicit list the ones we need.  Could you try it out and see if it becomes too unwieldy?

src/Futhark/LSP/Compile.hs
```diff
@@ -0,0 +1,59 @@
...
+import Control.Concurrent.MVar (MVar, putMVar, takeMVar)
+import Control.Monad.IO.Class (MonadIO (liftIO))
+import Futhark.Compiler.Program (LoadedProg, noLoadedProg, reloadProg)
+import Futhark.LSP.Diagnostic (errorToDiagnostics, sendDiagnostics, warningsToDiagnostics)
+import Futhark.LSP.Utils (State (..), debug, emptyState)
+import Language.Futhark.Warnings (listWarnings)
+import Language.LSP.Server (LspT)
+import Language.LSP.Types (TextDocumentVersion, filePathToUri, toNormalizedUri)
+
+-- try to take state from MVar, if it's empty (Nothing), try to compile.
```
athas, line 12: Use Haddock documentation comments for functions exported from the module.

src/Futhark/LSP/Diagnostic.hs
```diff
@@ -0,0 +1,57 @@
...
+sendDiagnostics :: NormalizedUri -> [Diagnostic] -> TextDocumentVersion -> LspT () IO ()
+sendDiagnostics uri diags version = publishDiagnostics 100 uri version (partitionBySource diags)
+
+mkDiagnostic :: Range -> DiagnosticSeverity -> T.Text -> Diagnostic
+mkDiagnostic range severity msg = Diagnostic range (Just severity) Nothing (Just "futhark") msg Nothing Nothing
+
+warningsToDiagnostics :: [(SrcLoc, Doc)] -> [Diagnostic]
+warningsToDiagnostics =
+  map
+    ( \(srcloc, msg) -> do
```
athas, line 35: This `do` is redundant.

src/Futhark/LSP/Utils.hs
```diff
@@ -0,0 +1,20 @@
+module Futhark.LSP.Utils
```
athas, line 1: This is not a good name for this module.  And how does it conceptually differ from the "Tools" module?

athas (changes requested):

src/Futhark/LSP/Utils.hs
haoranpb: I was thinking functions, etc defined in the `Utils` (utilities) module should be imported more than once.

While for `Tool` module, I put `getHoverInfoFromState` (only used once in `Handlers`) in it simply because it would be a bit crowded in `Handlers` otherwise.

What about I remove `Tool` and rename `Utils` to `Utilities`?

src/Futhark/CLI/LSP.hs
athas: That's fine.  It's a fairly harmless extension.

src/Futhark/LSP/Utils.hs
athas: I don't like the idea of "util" modules.  We have one already, `Futhark.Util`, but it is only for *truly* generic things.  The definition of the central state type for the LSP server is not really a "utility" - it is absolutely foundational.  Other parts of the compiler calls this kind of modules `Monad`, because they tend to define a monad used by other code.  That's not the case here.  Maybe `State` is a good name?  I imagine most of what will end up here will be state management.

(And actually, the `debug` function could be put in `Futhark.Util` as it is truly general.)

athas:
I merged the holes implementation and made holes produce warnings, so that should work straightforwardly now.

src/Futhark/LSP/Utils.hs
haoranpb: Yep, that sounds like a better idea!

src/Futhark/LSP/Handlers.hs
```diff
@@ -0,0 +1,62 @@
...
+import Control.Concurrent.MVar (MVar)
+import Control.Lens ((^.))
+import Control.Monad.IO.Class (MonadIO (liftIO))
+import Futhark.LSP.Compile (tryReCompile, tryTakeStateFromMVar)
+import Futhark.LSP.State (State (..))
+import Futhark.LSP.Tool (getHoverInfoFromState)
+import Futhark.Util (debug)
+import Futhark.Util.Pretty (pretty)
+import Language.LSP.Server (Handlers, LspM, getVersionedTextDoc, notificationHandler, requestHandler)
+import Language.LSP.Types
```
haoranpb, line 14: This is still importing lots of staff. There would be `27` more lines of import if I explicit import everything

src/Futhark/CLI/LSP.hs
```diff
@@ -0,0 +1,65 @@
...
+{-# LANGUAGE OverloadedStrings #-}
+
+module Futhark.CLI.LSP (main) where
+
+import Control.Concurrent.MVar (newMVar)
+import Control.Monad.IO.Class (MonadIO (liftIO))
+import Futhark.LSP.Handlers (handlers)
+import Futhark.LSP.State (emptyState)
+import Futhark.Util (debug)
+import Language.LSP.Server
```
athas, line 12: OK, seeing how many imports this needs, it's getting a bit silly.  Half of the file becomes imports!  It's probably better to move it back to how it used to be.

src/Futhark/LSP/Handlers.hs
athas: Yes, that's too much.  But you can consider a qualified import later if things get out of hand (`import qualified Language.LSP.Types as LSP`).

src/Futhark/LSP/Tool.hs
```diff
@@ -0,0 +1,37 @@
...
+    Nothing -> pure Nothing
+    Just loaded_prog -> do
+      let imports = lpImports loaded_prog
+      case atPos imports $ Pos path l c 0 of
+        Nothing -> pure Nothing
+        Just (AtName qn def _loc) -> do
+          case def of
+            Nothing -> pure Nothing
+            Just (BoundTerm t defloc) -> do
+              pure $ Just $ T.pack $ pretty qn ++ " :: " ++ pretty t ++ "\n\n" ++ "**Definition: " ++ locStr (srclocOf defloc) ++ "**"
```
athas, line 30: Why two colons?

haoranpb: I was thinking about copying the haskell-language-server style here. Say if we hover on `qn` here, HLS will show `qn :: QualName VName`

athas: That is because Haskell uses `::` for type ascription.  Futhark (and most other functional languages) use `:`.

haoranpb: Good point, fixed in https://github.com/diku-dk/futhark/pull/1618/commits/22a5de28199481672a7eddae94e7765f2423889e

PR #1622, opened 2022-04-04
lsp: adopt virtual file system

haoranpb:
Adopt virtual file system so that futhark language server never interacts with filesystem directly.

Document from [Specification](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_didOpen):

> the server must not try to read the document’s content using the document’s Uri


athas:
Can you clean up all the unrelated indentation changes?  It's hard to see what you actually did here.

haoranpb:
sure, the Haskell extension keeps automatically formatting on save

athas:
That is normally not a problem, but it looks like you are using a slightly older version of the extension, or perhaps one configured differently (althought I thought ormolu supported no configuration).  I'm not sure why the style checker in CI doesn't complain about it.

haoranpb:
Disabled Haskell extension's formatting, [ormolu extension](https://marketplace.visualstudio.com/items?itemName=sjurmillidahl.ormolu-vscode) gives the proper format

src/Futhark/Compiler/Program.hs
```diff
@@ -132,26 +132,22 @@ newImportMVar m = do
 
 -- | Read the content and modification time of a file.
 -- Check if the file exits in VFS before interact with file system directly.
-contentsAndModTime :: FilePath -> Maybe VFS -> IO (Maybe (Either String (T.Text, UTCTime)))
-contentsAndModTime filepath maybe_vfs = do
+contentsAndModTime ::
+  FilePath -> M.Map NormalizedUri VirtualFile -> IO (Maybe (Either String (T.Text, UTCTime)))
```
athas, line 136: Use a type abbreviation `type VFS = M.Map NormalizedUri VirtualFile`.  (Or some other concise name.)

src/Futhark/Compiler/Program.hs
```diff
@@ -51,8 +51,8 @@ import Language.Futhark.Prop (isBuiltin)
 import Language.Futhark.Semantic
 import qualified Language.Futhark.TypeChecker as E
 import Language.Futhark.Warnings
-import Language.LSP.Types (normalizedFilePathToUri, toNormalizedFilePath)
-import Language.LSP.VFS (VFS (vfsMap), virtualFileText)
+import Language.LSP.Types (NormalizedUri, normalizedFilePathToUri, toNormalizedFilePath)
+import Language.LSP.VFS (VirtualFile, virtualFileText)
```
athas, line 55: I don't think this is appropriate.  Something is wrong if this module has to depend on LSP types.  You should define the VFS contents as `M.Map FilePath T.Text` (with the `FilePath` understood as an absolute path).  The LSP server itself then has to preprocess the `VirtualFile`s and URIs and whatever *before* calling anything in this module.  It is important to maintain layering, and putting LSP details inside this foundational module is a layering violation.

haoranpb: Great point, removed all imports from lsp package in `Program.hs`

PR #1623, opened 2022-04-04
extract isBuiltin to Language.Futhark.Prop

haoranpb:
Bump into `isBuiltin` when trying to define something similar for the language server.

Quick question in the meantime, where can I find [prelude](https://github.com/diku-dk/futhark/tree/master/prelude) in the filesystem after installing futhark?


athas:
This function should go in `Language.Futhark.Prop`.  The `Futhark.Util` module is for non-Futhark-specific utilities.

PR #1629, opened 2022-04-07
Reactive benchmarking (closes #1335)

### LINKED ISSUE #1335: Reactive benchmarking
`futhark bench` currently uses a fixed number of runs.  This is in many cases inflexible.  It would be nice to support a reactive mode where the number of runs is based on how long a single run takes, and how quickly we establish confidence in the measurement results.  Probably we should look at what other tools like [criterion](http://www.serpentine.com/criterion/) do.

aleksanderJunge:
A solution to   https://github.com/diku-dk/futhark/issues/1335

Trying to automatically determine the number of runs needed to conclude a benchmark


athas:
I think we either need a stricter bound on maximum runtime or to be less picky regarding convergence.  Some benchmarks (e.g. bfast) run for a very long time (16k runs).

Munksgaard:
Shouldn't both be specifiable as arguments to `futhark bench`?

aleksanderJunge:
You are right, sometimes it seems to run for quite a while.

I think perhaps a time limit would be preferable as default, as some of the benchmarks converge pretty quickly despite being in the highest autocorrelation group (e.g. if only the first few runs are slower than the subsequent runs).

Part of the problem may be if later runs also have a high degree of variance, (for instance running a benchmark with a varying background load seems to greatly increase the number of runs required to converge, I tested this by running 2 benchmarks simultaneously).

bfast with noise (18k runs):
runs_background_noise

bfast without noise (6k runs):
bfast_no_noise

aleksanderJunge:
but perhaps the convergence criteria is also a bit too conservative.

athas:
Maybe this is a "don't blame the messenger" thing. Bfast really does have odd behaviour, such as very high autocorrelation due to initially slow runs and (very) slowly increasing runtimes.  It takes a while to hit that 0.001 RSD.

athas:
In the convergence loop, what is the justification for using `resample` before computing the RSD?  Would it not be better (or at least equally good) to compute the RSD using all samples?

athas:
It seems it mostly serves to ignore outliers, but is it justifiable to do that by random sampling?

athas:
I just did another `futhark bench` run with this branch on futhark01 and it took 14m.  Previously it took 7m.  I've seen much longer at other times (close to an hour at worst), but now I'm thinking that might have been because the machine was actually loaded and the measurements very uncertain.  I am leaning towards merging this as-is (at least after we're done with the SC22 artifacts this week - I don't want to mess those up).  While it may sometimes run for too long, it's probably fine to tweak it later.

I am a little nervous about the benchmark JSONs growing much larger than they used to be and the server hosting futhark-lang.org running out of space, but I suppose we can deal with that if it happens.

athas:
It was only that fast because I had fat-fingered the convergence.  But that's still useful news: just by imposing a 0.5s minimum measuring time (which is *definitely* a good idea), we're already 2x slower.  The cost of also doing convergence should be counted relative to *that*.

aleksanderJunge:
> In the convergence loop, what is the justification for using `resample` before computing the RSD? Would it not be better (or at least equally good) to compute the RSD using all samples?

"to a man with a hammer everything looks like a nail", I had just read a book about bootstrapping and it seemed like a great idea to apply it. However, it is actually only beneficial when an analytical form of the estimator is unknown which is not the case here. So calculating RSD on the samples directly without resampling should be better I think (and save a lot of time).

side note:
in the convergenceCriteria list, shouldn't (0.065, 0.0025) be (0.65, 0.0025) and (0.045, 0.0050) be (0.45, 0.0050) or is this intentional?

aleksanderJunge:
the analytical form would be SD(X) / sqrt( |X| )
where X is the sample vector.

athas:
Oh yeah, that's a bad typo.

athas:
Running the full benchmark suite with convergence takes about an hour.  With a minimum of 0.5s or 10 runs (i.e. no convergence), it takes 14 minutes.  Only a few benchmarks and datasets benefit from actually running to convergence:

```
futhark-benchmarks/accelerate/canny/canny.fut
  data/lena256.in:                                                      1.25x
  data/lena512.in:                                                      1.11x

futhark-benchmarks/accelerate/hashcat/hashcat.fut
  rockyou.dataset:                                                      1.24x

futhark-benchmarks/finpar/LocVolCalib.fut
  LocVolCalib-data/large.in:                                            0.98x
  LocVolCalib-data/medium.in:                                           1.01x
  LocVolCalib-data/small.in:                                            1.23x

futhark-benchmarks/micro/mmm/mmm-batch.fut
  [64][128][32][32]f32 [64][128][32][32]f32:                            1.21x

futhark-benchmarks/rodinia/lavaMD/lavaMD.fut
  data/10_boxes.in:                                                     1.27x
  data/3_boxes.in:                                                      1.59x
```

Interestingly, most of the microbenchmarks do not benefit.  Probably the 0.5s minimum runs are good enough in that case.

I wrote a small Python script that reads a bench.json and shows the cumulative runtime and number of runs for each dataset:

```Python
#!/usr/bin/env python3

import json
import sys

data = json.load(open(sys.argv[1]))

for benchmark in data:
    for d in data[benchmark]['datasets']:
        dataset = data[benchmark]['datasets'][d]
  	print('{} {} {}:	{} runs'.
              format(sum(dataset['runtimes'])/1000000, benchmark, d, len(dataset['runtimes']), ))
```

When piped into `sort -n` it's useful to see where the time is actually spent.  For example, these are the ten most time-consuming datasets:

```
21.187857 futhark-benchmarks/parboil/mri-q/mri-q.fut data/large.in: 7066 runs
21.401452 futhark-benchmarks/misc/knn-by-kdtree/driver-knn.fut 256i32 [2097152][7]f32 [10000000][7]f32: 10 runs
22.006383 futhark-benchmarks/rodinia/particlefilter/particlefilter.fut data/128_128_10_image_10000_particles.in: 4605 runs
22.292535 futhark-benchmarks/accelerate/crystal/crystal.fut #5 ("4000i32 30.0f32 50i32 1i32 1.0f32"): 3937 runs
22.721532 futhark-benchmarks/accelerate/nbody/nbody.fut data/100000-bodies.in: 1167 runs
22.729389 futhark-benchmarks/rodinia/lud/lud.fut data/2048.in: 2094 runs
24.073706 futhark-benchmarks/finpar/LocVolCalib.fut LocVolCalib-data/medium.in: 1396 runs
24.523285 futhark-benchmarks/misc/bfast/bfast-cloudy.fut data/africa.in: 931 runs
249.07418 futhark-benchmarks/finpar/LocVolCalib.fut LocVolCalib-data/small.in: 9369 runs
261.254545 futhark-benchmarks/misc/heston/heston32.fut data/10000_quotes.in: 12921 runs
```

I really suspect there is not a need to hammer on heston32.fut for over four minutes, so probably we can still consider calibrating some of the constants.

athas:
With the newest changes, it takes a *lot* of samples.  I suspect the autocorrelation is high due to the initial warmup period.  Since our goal is to measure sustained performance, maybe a shortcut would be to remove the initial measurements as outliers.

aleksanderJunge:
> With the newest changes, it takes a _lot_ of samples. I suspect the autocorrelation is high due to the initial warmup period. Since our goal is to measure sustained performance, maybe a shortcut would be to remove the initial measurements as outliers.

Does the new way of measuring the relative standard error use significantly more samples? hmm.

Removing the first runs could work in a lot of cases I think. But the challenge may be to identify when the warmup period ends, as it varies from benchmark to benchmark, while some benchmarks have multiple plateaus of performance increases. Maybe it would be possible to do some test of the collected data after _runMinimum_ and check whether the autocorrelation decreases if we remove _x_ samples from the front of the list. 

btw, what does the 1.25x mean in the above?:

futhark-benchmarks/accelerate/canny/canny.fut
  data/lena256.in:                                                      1.25x
  ...

aleksanderJunge:
Maybe it's the updated _convergenceCriteria_ that's making it use more samples?
I found the analytical way of calculating relative std err is usually within 1% of the old one (and is the more accurate measure).

athas:
Aleksander Junge ***@***.***> writes:

> Does the new way of measuring the relative standard error use significantly more samples? hmm.

No, it doesn't.  I posted before it was done running, so it was just a
gut feeling.  There is no real difference.

> Removing the first runs could work in a lot of cases I think. But the
> challenge may be to identify when the warmup period ends, as it varies
> from benchmark to benchmark, while some benchmarks have multiple
> plateaus of performance increases.

I'm going to try to just remove the samples done during the minimum
measurements phase and see what happens.

> btw, what does the 1.25x mean in the above?:

That when benchmarking with convergence, the observed performance has a
1.25x speedup compared to only doing the minimum number of measurements.
The specific number doesn't matter; it reveals that doing the minimum
number of runs failed to uncover the true sustained performance.

-- 
\  Troels
/\ Henriksen

athas:
Alright, I did the change.  Now the results from phase 1 are removed when we enter phase 2.  My experiments show that the behaviour is excellent: the results we obtain are essentially the same, but they are gathered *much* faster.  The convergence loop is only entered for particularly noisy benchmarks.  The display is now a bit messy, because it doesn't understand that the run count is conceptually reset, but that can be fixed.

PR #1630, opened 2022-04-07
Use maps in all Primitives tests

Yakokse:


athas:
Writing this patch must have sucked.  Thanks for doing it!

PR #1636, opened 2022-04-13
lsp: fix multiple open files

haoranpb:
when we navigate between **opened** files in vscode, no more "open file event" is fired, `state` is not re-compiled.

Re-pro example:
1. if we first open `test_1.fut` and then open `test_2.fut`
2. the `state` stores the result for `test_2.fut`
3. navigate back to `test_1.fut` and try to get hover info

Together with https://github.com/diku-dk/futhark-vscode/commit/130185a5eaf0e5b9ae117549e83e7a89e50e646a , we fire an custom event whenever the focus is changed, to re-compile


haoranpb:
Hmm, yes, indeed. Not sure if the project-based root is a more common approach though, I was playing around with haskell-language-server, I can see some activities runs when I switch between opened files.

If we treat the current workspace as root, then we need to span up a new server for every futhark file in the workspace, and we still need to listen to the [onDidChangeActiveTextEditor](https://github.com/diku-dk/futhark-vscode/blob/130185a5eaf0e5b9ae117549e83e7a89e50e646a/src/extension.ts#L26) event to switch to the corresponding language server.

Given that futhark file is usually fast to compile (it took around 1s for https://github.com/diku-dk/futhark-benchmarks/blob/master/misc/heston/heston.fut), I think it's a good trade off , what do you think?

athas:
I think it is something worth mentioning in your report.

PR #1638, opened 2022-04-14
lsp: add documentation link for build-in functions

haoranpb:
Add documentation link for prelude functions

Still unsure about whether to include definitions for prelude functions, it's true that the definition is a bit useless, but the UI looks a little bit wired

Screen Shot 2022-04-14 at 15 37 08

### DIFF
--- a/src/Futhark/LSP/Tool.hs
+++ b/src/Futhark/LSP/Tool.hs
@@ -38,10 +38,11 @@ getHoverInfoFromState state (Just path) l c = do
       Just $
         (prettyText qn <> " : " <> prettyText t)
           <> if isBuiltinLoc defloc
-            then mempty
+            then "\n\n" <> futharkDocUrl defloc
             else "\n\n**Definition: " <> T.pack (locStr (srclocOf defloc)) <> "**"
     bound
-      | isBuiltinLoc (boundLoc bound) -> Just "Builtin definition."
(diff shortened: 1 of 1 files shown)


athas:
I don't think this is a useful feature.  It just links to a page, not to the specific function.  Linking to the specific function in the prelude documentation is difficult and still not particularly useful.  Why should the prelude receive special treatment over separate libraries?  It would be a bad precedent.  We should not include things that are not fully general.  Extracting the doc comment (e.g. `valBindDoc`) from the declaration would be much more useful.

haoranpb:
> Why should the prelude receive special treatment over separate libraries?

Actually, I got started on this feature because I feel like prelude is treated a bit differently.

The "Go to definition" feature works for all files / libraries (except for prelude), because they are all stored on disk (libraries will be stored under `lib/` after `futhark pkg sync`). Since the feature doesn't work for prelude, I was thinking maybe the link to documentation site would make up for it.

Also inspired by the haskell language server, but they do link to the particular function.

Screen Shot 2022-04-15 at 08 29 46

> Extracting the doc comment (e.g. valBindDoc) from the declaration would be much more useful.

Agree, that would be a much better feature.

---

I made the PR in a rush because I was hoping to catch the release, but I guess it doesn't have to and I can give it more thought after I finish the tasks at hand

PR #1640, opened 2022-04-16
lsp: fix diagnostics are not flushed

haoranpb:
This reverts commit cd000141224b7f2fb9634bad85f7b19435c22ee4.

Not sure what's the implementation detail in `lsp` package, but diagnostics are not flushed without `diagnosticSource`.

Re-pro steps:
1. open the correct futhark file
2. trigger an error/warning to get diagnostic
3. fix the problem just triggered
4. the diagnostic won't disappear

### DIFF
--- a/src/Futhark/LSP/Compile.hs
+++ b/src/Futhark/LSP/Compile.hs
@@ -9,7 +9,7 @@ import Control.Monad.IO.Class (MonadIO (liftIO))
 import qualified Data.Map as M
 import qualified Data.Text as T
 import Futhark.Compiler.Program (LoadedProg, lpWarnings, noLoadedProg, reloadProg)
-import Futhark.LSP.Diagnostic (maxDiagnostic, publishErrorDiagnostics, publishWarningDiagnostics)
+import Futhark.LSP.Diagnostic (diagnosticSource, maxDiagnostic, publishErrorDiagnostics, publishWarningDiagnostics)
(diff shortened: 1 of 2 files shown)


athas:
Can we at least make it invisible?  It seems pointless that there is a string "futhark" next to every diagnostic.  Do other language extensions have this?

haoranpb:
A quick search on [haskell-language-server](https://github.com/haskell/haskell-language-server), I find [ideErrorText](https://github.com/haskell/haskell-language-server/blob/7518a3a7eb9c11b2dbbff0aed1d982fd3e3505a9/ghcide/src/Development/IDE/Types/Diagnostics.hs#L52) and [diagFromErrMsg](https://github.com/haskell/haskell-language-server/blob/7518a3a7eb9c11b2dbbff0aed1d982fd3e3505a9/ghcide/src/Development/IDE/GHC/Error.hs#L61).

```haskell
ideErrorText :: NormalizedFilePath -> T.Text -> FileDiagnostic
ideErrorText = ideErrorWithSource (Just "compiler") (Just DsError)

diagFromErrMsg :: T.Text -> ...
diagFromErrMsg diagSource ...
```

I'm guessing they included `DiagnosticSource` because of their [plugin system](https://github.com/haskell/haskell-language-server/tree/master/plugins), so one plugin won't accidentally flush (delete) diagnostics from the others, which we don't have to worry about

PR #1645, opened 2022-04-17
lsp: positionmap when use stale type-checked result

### LINKED ISSUE #1621: PositionMapping when use stale compile result
Moved from https://github.com/haoranpb/futhark-language-server/issues/1, as the repository is archived.

When using stale compile result, if one new line is inserted at the beginning of the file, everything would be messed up, need to figure out a way to convert source locations in the current version of a document back into the version of the document for which the rule was last computed successfully.

Related implementations from [haskell-language-server](https://github.com/haskell/haskell-language-server/blob/411db02883d6985efb37918a0b9100d727fe8554/ghcide/src/Development/IDE/Core/UseStale.hs)

Idea from @athas 

> If you also store the full contents of the file/buffer at the last time the program managed to type-check, you can compute a diff that maps old source locations to new source locations.

haoranpb:
close #1621  , referencing [deltaFromDiff](https://github.com/haskell/haskell-language-server/blob/7518a3a7eb9c11b2dbbff0aed1d982fd3e3505a9/ghcide/src/Development/IDE/Core/PositionMapping.hs#L185)

The logic for computing mapping and converting between current and stale positions is ready.

But there is still some problems when we navigate between files, should be fixed in another PR probably. Using [heston](https://github.com/diku-dk/futhark-benchmarks/tree/master/misc/heston) for example:

1. Open `heston.fut`
2. Go to def (or simply open) `least_squares.fut` (State will be re-compiled, AST won't have `heston.fut` anymore)
3. Tigger some errors in `least_squares.fut`
4. Go back to `heston.fut` and try to Go to def to `least_squares.fut`

### DIFF
--- a/src/Futhark/LSP/Compile.hs
+++ b/src/Futhark/LSP/Compile.hs
@@ -11,7 +11,7 @@ import Data.Maybe (fromMaybe)
 import qualified Data.Text as T
 import Futhark.Compiler.Program (LoadedProg, lpWarnings, noLoadedProg, reloadProg)
 import Futhark.LSP.Diagnostic (diagnosticSource, maxDiagnostic, publishErrorDiagnostics, publishWarningDiagnostics)
-import Futhark.LSP.State (State (..), emptyState)
+import Futhark.LSP.State (State (..), emptyState, updateStaleContent)
 import Futhark.Util (debug)
 import Language.Futhark.Warnings (listWarnings)
 import Language.LSP.Server (LspT, flushDiagnosticsBySource, getVirtualFile, getVirtualFiles)
@@ -29,7 +29,7 @@ tryTakeStateFromMVar state_mvar file_path = do
   old_state <- liftIO $ takeMVar state_mvar
   case stateProgram old_state of
     Nothing -> do
-      new_state <- tryCompile file_path noLoadedProg
+      new_state <- tryCompile old_state file_path noLoadedProg
       liftIO $ putMVar state_mvar new_state
       pure new_state
     Just _ -> do
@@ -42,7 +42,7 @@ tryReCompile state_mvar file_path = do
   debug "(Re)-compiling ..."
   old_state <- liftIO $ takeMVar state_mvar
   let loaded_prog = getLoadedProg old_state
-  new_state <- tryCompile file_path loaded_prog
+  new_state <- tryCompile old_state file_path loaded_prog
   case stateProgram new_state of
     Nothing -> do
       debug "Failed to (re)-compile, using old state or Nothing"
@@ -53,11 +53,11 @@ tryReCompile state_mvar file_path = do
 
 -- | Try to compile, publish diagnostics on warnings and errors, return newly compiled state.
 --  Single point where the compilation is done, and shouldn't be exported.
-tryCompile :: Maybe FilePath -> LoadedProg -> LspT () IO State
-tryCompile Nothing _ = pure emptyState
-tryCompile (Just path) old_loaded_prog = do
+tryCompile :: State -> Maybe FilePath -> LoadedProg -> LspT () IO State
+tryCompile _ Nothing _ = pure emptyState
+tryCompile state (Just path) old_loaded_prog = do
   vfs <- getVirtualFiles
-  res <- liftIO $ reloadProg old_loaded_prog [path] (transformVFS vfs)
+  res <- liftIO $ reloadProg old_loaded_prog [path] (transformVFS vfs) -- NOTE: vfs only keeps track of current opened files
   flushDiagnosticsBySource maxDiagnostic diagnosticSource
   case res of
     Right new_loaded_prog -> do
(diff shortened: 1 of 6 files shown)


src/Futhark/LSP/Tool.hs
```diff
@@ -54,10 +54,16 @@ findDefinitionRange state (Just path) l c = do
 findDefinitionRange _ _ _ _ = Nothing
 
 queryAtPos :: State -> Pos -> Maybe AtPos
-queryAtPos state pos =
-  case stateProgram state of
-    Nothing -> Nothing
-    Just loaded_prog -> atPos (lpImports loaded_prog) pos
+queryAtPos state pos = do
```
haoranpb, line 57: Need to get the current file content via [getVirtualFile](https://hackage.haskell.org/package/lsp-1.4.0.0/docs/Language-LSP-Server.html#v:getVirtualFile) here, not sure how to do that except by adding `LspM () ...` to the return type of `queryAtPos`, `findDefinitionRange` and `getHoverInfoFromState`. But feels wrong to do that, any tips? @athas

athas: Perhaps you can call that function outermost and pass on the result as an extra parameter.

athas:
Looks pretty cool.  I wonder how slow it is to compute those diffs all the time, but probably not very.  I'm a bit skeptical about the use of lists for mapping line numbers, but perhaps it's not a big deal.

src/Futhark/LSP/Compile.hs
```diff
@@ -45,29 +52,39 @@ tryReCompile :: IORef State -> Maybe FilePath -> LspT () IO ()
 tryReCompile state_mvar file_path = do
   debug "(Re)-compiling ..."
   old_state <- liftIO $ readIORef state_mvar
-  new_state <- tryCompile file_path old_state
+  let loaded_prog = getLoadedProg old_state
+  new_state <- tryCompile old_state file_path loaded_prog
+  debug $ show $ staleData old_state
```
athas, line 57: I think this debugging statement produces too much input to leave in.

src/Futhark/LSP/PositionMapping.hs
```diff
@@ -0,0 +1,72 @@
...
+import Data.Algorithm.Diff (Diff, PolyDiff (Both, First, Second), getDiff)
+import Data.Bifunctor (Bifunctor (bimap, first, second))
+import qualified Data.Text as T
+import Futhark.Util.Loc (Loc (Loc), Pos (Pos))
+import Language.LSP.VFS (VirtualFile)
+
+-- | A mapping between current file content and the stale (last successful compiled) file content
+-- currently, only supports entire line mapping
+-- more detailed mapping might be achieved via referring to haskell-language-server@efb4b94
+newtype PositionMapping = PositionMapping ([Int], [Int])
```
athas, line 20: I am not comfortable using lists here, as lookups are slow *especially* if at some point we start doing many lookups (e.g. for semantic highlighting).  Use a Vector instead, like I think you have done elsewhere.  Also, document what the two sequences actually mean.

src/Futhark/LSP/State.hs
```diff
@@ -2,17 +2,50 @@
...
+import Futhark.LSP.PositionMapping (PositionMapping, StaleFile (..))
+import Language.LSP.VFS (VirtualFile)
 
 -- | The state of the language server.
-newtype State = State
+data State = State
   { -- | The loaded program.
-    stateProgram :: Maybe LoadedProg
+    stateProgram :: Maybe LoadedProg,
+    -- | The last succussful type-checked files.
```
athas, line 21: successful

src/Futhark/LSP/State.hs
```diff
@@ -2,17 +2,50 @@
...
+    staleData :: M.Map FilePath StaleFile
   }
 
 -- | Initial state.
 emptyState :: State
-emptyState = State Nothing
+emptyState = State Nothing M.empty
+
+-- | Get the contents of a stale (last succuessfully complied) file's contents.
```
athas, line 29: successfully

athas (changes requested):

src/Futhark/LSP/PositionMapping.hs
haoranpb: Good point

src/Futhark/LSP/PositionMapping.hs
```diff
@@ -0,0 +1,78 @@
...
+import Futhark.Util.Loc (Loc (Loc), Pos (Pos))
+import Language.LSP.VFS (VirtualFile)
+
+-- | A mapping between current file content and the stale (last successful compiled) file content.
+-- The first component of the pair provides the mapping from stale position to current,
+-- e.g. old2new[2] = 4 means "line 2" in the stale file corresponds to "line 4" in the current file.
+-- The second component vice versa.
+-- Currently, only supports entire line mapping,
+-- more detailed mapping might be achieved via referring to haskell-language-server@efb4b94
+newtype PositionMapping = PositionMapping (V.Vector Int, V.Vector Int)
```
athas, line 24: Instead of `newtype`, use `data` with properly named fields.

src/Futhark/LSP/PositionMapping.hs
```diff
@@ -0,0 +1,78 @@
...
+  { -- | The last successfully compiled file content.
+    -- Using VirtualFile for convenience, we can use anything with {version, content}
+    staleContent :: VirtualFile,
+    -- | PositionMapping between current and stale file content.
+    -- Nothing if last type-check is successful.
+    staleMapping :: Maybe PositionMapping
+  }
+
+instance Show StaleFile where
+  show (StaleFile _ mapping) = show mapping
```
athas, line 38: Non-`derived` instances of `Show` are a bad idea.  Let people write `show . staleMapping` if they want to see that part.

athas (changes requested):

PR #1654, opened 2022-04-25
Multi gpu

demiguard:
This is the initial starting point for multi GPU support to futhark


rts/c/cuda.h
```diff
@@ -249,6 +257,45 @@ static int cuda_device_setup(struct cuda_context *ctx) {
...
+      devices[used_devices] = devID;
+      used_devices++;
+    }
+  }
+  if(ctx->cfg.debugging){
+    fprintf(stderr, "Using %d devices\n", used_devices);
+  }
+  ctx->kernel_iterator = 0;
+  ctx->device_count = used_devices;
+  ctx->devices = (CUdevice*)malloc(sizeof(CUdevice)*used_devices);
```
athas, line 278: I prefer `calloc()` for this.  Yes, I know the rest of the code isn't consistent.  Be better than I was!

rts/c/cuda.h
```diff
@@ -290,7 +337,9 @@ static const char *cuda_nvrtc_get_arch(CUdevice dev) {
     { 7, 0, "compute_70" },
     { 7, 2, "compute_72" },
     { 7, 5, "compute_75" },
-    { 8, 0, "compute_80" }
+    // NVRTC doesn't support compiling to 8.0, See nvrtc.h
+    { 8, 0, "compute_75" },
+    { 8, 6, "compute_75" }
```
athas, line 342: Wait, what?  Even if we want this, it should probably be a separate PR.  Have you tested that the CUDA backend works on these new GPUs?

rts/c/cuda.h
```diff
@@ -249,6 +257,45 @@ static int cuda_device_setup(struct cuda_context *ctx) {
...
+  ctx->device_count = used_devices;
+  ctx->devices = (CUdevice*)malloc(sizeof(CUdevice)*used_devices);
+  ctx->contexts = (CUcontext*)malloc(sizeof(CUcontext)*used_devices);
+  ctx->modules = (CUmodule*)malloc(sizeof(CUmodule)*used_devices);
+  ctx->kernel_done = (CUevent*)malloc(sizeof(CUevent)*used_devices*2);
+  ctx->finished = (CUevent*)malloc(sizeof(CUevent)*used_devices);
+
+  for(int devIdx = 0; devIdx < used_devices; devIdx++){
+    CUDA_SUCCEED_FATAL(cuDeviceGet(&ctx->devices[devIdx], devices[devIdx]));
+    CUDA_SUCCEED_FATAL(cuCtxCreate(&ctx->contexts[devIdx],
```
athas, line 286: The documentation for `cuCtxCreate()` suggests that it is better to use `cuDevicePrimaryCtxRetain()`.

src/Futhark/CodeGen/Backends/CCUDA.hs
```diff
@@ -134,6 +134,13 @@ cliOptions =
              optionArgument = NoArgument,
              optionDescription = "Gather profiling data while executing and print out a summary at the end.",
              optionAction = [C.cstm|futhark_context_config_set_profiling(cfg, 1);|]
+           },
+         Option
+           { optionLongName = "use-single-device",
```
athas, line 139: This is fine for development, but we'll need to think about what the configuration API should actually be like.  It's fiddly stuff that isn't conceptually very interesting or difficult, but can still be very time-consuming.  I think Cosmin and I will try to work something out.

src/Futhark/CodeGen/Backends/CCUDA.hs
```diff
@@ -326,7 +351,8 @@ callKernel (LaunchKernel safety kernel_name args num_blocks block_size) = do
       grid[perm[1]] = $exp:grid_y;
       grid[perm[2]] = $exp:grid_z;
 
-      void *$id:args_arr[] = { $inits:args'' };
+      int device_id = 0; 
+      void *$id:args_arr[] = { $inits:args'' , &device_id };
```
athas, line 355: I would prefer if `device_id` was the first argument, if possible.  (So that all the arguments that are _always_ passed come first.)

src/Futhark/CodeGen/Backends/CCUDA.hs
```diff
@@ -337,12 +363,39 @@ callKernel (LaunchKernel safety kernel_name args num_blocks block_size) = do
...
+            grid_MD[perm[2]] = $exp:grid_z_MD;
+            CUDA_SUCCEED_FATAL(cuCtxPushCurrent(ctx->cuda.contexts[devID]));
+            CUDA_SUCCEED_FATAL(cuLaunchKernel(ctx->$id:(kernelMultiDevice kernel_name)[devID],
+                                                  grid_MD[0], grid_MD[1], grid_MD[2],
+                                                  $exp:block_x, $exp:block_y, $exp:block_z,
+                                                  $exp:shared_tot, NULL,
+                                                  $id:args_arr, NULL));
+            CUDA_SUCCEED_FATAL(cuEventRecord(ctx->cuda.kernel_done[devID * 2
+                                             + ctx->cuda.kernel_iterator], NULL));
+            for(int other_dev = 0; other_dev < ctx->cuda.device_count; other_dev++){
```
athas, line 382: I am not at all sure what is going on here, and I remember we talked about it.  Any other reader would be completely mystified.

But more significantly, I'm not completely sure it is correct.  You allocate two event objects per device (right?).  How can we be sure only one kernel is in flight at a time?  Or doesn't that matter?  Is it OK that the host-side events are overwritten?

src/Futhark/CodeGen/Backends/CCUDA/Boilerplate.hs
```diff
@@ -325,6 +337,26 @@ generateContextFuns cfg cost_centres kernels sizes failures = do
           concatMap forKernel (M.keys kernels)
             ++ concatMap forCostCentre cost_centres
 
+      forKernelMD name =
+        ( [C.csdecl| typename CUfunction* $id:(kernelMultiDevice name);|],
+          [C.cstms| ctx->$id:(kernelMultiDevice name) = malloc(sizeof(CUfunction)*ctx->cuda.device_count);
```
athas, line 342: `calloc()`

src/Futhark/CodeGen/ImpGen/GPU.hs
```diff
@@ -225,15 +225,15 @@ withAcc pat inputs lam = do
     locksForInputs atomics ((c, (_, _, op)) : inputs')
       | Just (op_lam, _) <- op,
         AtomicLocking _ <- atomicUpdateLocking atomics op_lam = do
-          let num_locks = 100151
```
athas, line 228: I think you are messing up this indentation because you are using a slightly too old version of the `ormolu` formatter.

athas (changes requested):

athas:
This looks quite good, and delightfully simple.  I left some review comments.  I'll also take the liberty of merging master into your branch and fixing some style issues.

rts/c/cuda.h
demiguard: It's mostly a convenience thing for me, since I've a 8.6 compute capable device, so i do not have to manually edit programs after i compile them. As far as my testing go, I've not encountered any errors nor do i get any errors when i run a "stack test", but I agree it doesn't really belong to this PR.

src/Futhark/CodeGen/Backends/CCUDA.hs
demiguard: The events exists to provide a kernel to kernel synchronization. If you have 2 kernels and put them on the same stream, you know that the first kernel is finished by the time kernel 2 starts and any data dependencies on kernel 2 from kernel 1 is fulfilled. If you launch kernels on multiple streams, which I'm forced to because I've multiple devices, I lose such assurances.
In our case is "divide" the kernels into smaller version. Say for example: kernel_1_1, kernel_1_2, kernel_1_2  kernel_2_2.
There might exists data dependencies from kernel_1_1 to kernel_2_2, hence kernel_1_1 needs to be finished before kernel_2_2 can be launched.

Hence the streams I create looks like this:

Lingo: Sx: Null stream on Device x, Kxy: Kernel_x_y, RDxEy: Record event y from device x, WDxEy: stream wait on device x event y

S1: K11 RD1E1 WD2E1 K21 RD1E2 WD2E2 K32 RD1E1 WD2E1 ...
S2: K12 RD2E1 WD1E1 K22 RD2E1 WD1E2 K32 RD2E1 WD1E1 ...

The re usage of events is ok, because re-recording events just overwrites the state of the event. which is also why we need two events because otherwise a stream could end up waiting on the same event twice and that would cause a deadlock.

The documentation for reusage of events: https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EVENT.html#group__CUDA__EVENT_1g95424d3be52c4eb95d83861b70fb89d1

rts/c/cuda.h
athas: Then I'll make the change in master.

athas: I don't really understand the reference to NVRTC not supporting CC 8.0.  We've been using it for over a year.

demiguard: I think the most plausible explanation is that I'm using an outdated version of nvrtc

athas:
This PR now valides on a *single* GPU.  I have not yet measured whether there are any performance regressions.

PR #1657, opened 2022-04-26
Reduce synchronous GPU memory transfers

### LINKED ISSUE #1045: Optimising scalar computation interspersed between GPU operations
Consider an implementation of Gaussian elimination:

```futhark
  let gauss_jordan [m] [n] (A:[m][n]t) =
    loop A for i < i32.min m n do
    -- Find nonzero value.
    let j = A[i:,i] |> map T.abs |> argmax |> (.1) |> (+i)
    let f = T.((i32 1-A[i,i]) / A[j,i])
    let irow = map2 (T.fma f) A[j] A[i]
    in tabulate m (\j ->
                     let f = T.negate A[j,i]
                     in map2 (\x y -> if j == i then x else T.fma f x y)
                             irow A[j])
```

`argmax` is a reduction.  The problem here is that we copy the scalar `j` from the GPU to the host just to compute the scalar `f` (which also involves copying twice from `A` to the host!), which is then immediately used in more GPU operations.  This is wasteful.  Indeed, most of this program's runtime is spent copying those scalars around.  It would be much better if we just did those operations on the GPU instead.  I can think of two ways of accomplishing this.

### 1. Extending SegRed with a postlude operator

When the SegRed that implements `argmax` writes the final result, it does in principle have the opportunity to do an arbitrary amount of scalar computation before actually writing it.  We could extend SegReds with a "postlude operator" that is applied to the final result(s) before they are written.  A bottom-up compiler pass would then identify SegRed results that are used in subsequent arithmetic operations and do a kind of backward fusion to move it into this postlude operator.  SegReds already return their results as single-element arrays (that is, GPU-resident), so this would fairly elegantly reduce the amount of needed CPU-GPU communication.  This could also be added to SegScan, giving us the ability to fuse `scan` into `map`.

### 2. Adding a "device code" block construct

We could extend the Kernels IR with a construct that contains a Body and indicates that this Body should be run as a single-thread GPU kernel.  This would allow us to fairly easily migrate code to the GPU to cut down on communication costs.

PhilipBorgesen:
A pull request as agreed upon with @athas. Fixes #1045.

The main additions and changes of interest:
* Adds a sequential GPUBody kernel to the GPU representation, implemented by @athas.
* Implemented merging of GPUBody kernels and other simplifications within the GPU pipeline.
* The GPU pipeline now moves statements into GPUBody kernels to reduce synchronous GPU memory transfers. The model used to drive the transformations favor the OpenCL compiler backend.
* Array indexing done by for-in loops can now be sunk.
* Indexing into single-element arrays are no longer hoisted from kernels.

For details see the module descriptions of:
* [src/Futhark/Pass/MergeGPUBodies.hs](https://github.com/PhilipBorgesen/futhark/tree/master/src/Futhark/Pass/MergeGPUBodies.hs)
* [src/Futhark/Pass/ReduceDeviceSyncs.hs](https://github.com/PhilipBorgesen/futhark/tree/master/src/Futhark/Pass/ReduceDeviceSyncs.hs)
* [src/Futhark/Analysis/MigrationTable.hs](https://github.com/PhilipBorgesen/futhark/tree/master/src/Futhark/Analysis/MigrationTable.hs)
* [src/Futhark/Analysis/MigrationTable/Graph.hs](https://github.com/PhilipBorgesen/futhark/tree/master/src/Futhark/Analysis/MigrationTable/Graph.hs)

Preliminary benchmarking of [futhark-benchmarks](https://github.com/diku-dk/futhark-benchmarks) programs affected by the changes show mean runtime improvements between 10 and 50 percent of those programs that are affected, depending on hardware. The bars are the average improvement over the datasets for each program entry point. A height of 1.0 means 100% better, i.e. twice as fast as the baseline.
billede
billede
billede
billede

### DIFF
--- a/futhark.cabal
+++ b/futhark.cabal
@@ -216,6 +216,7 @@ library
       Futhark.LSP.Handlers
       Futhark.LSP.Tool
       Futhark.LSP.State
+      Futhark.LSP.PositionMapping
       Futhark.MonadFreshNames
       Futhark.Optimise.BlkRegTiling
       Futhark.Optimise.CSE
@@ -336,6 +337,7 @@ library
     , bmp >=1.2.6.3
     , containers >=0.6.2.1
     , cryptohash-md5
+    , Diff >=0.4.1
     , directory >=1.3.0.0
     , directory-tree >=0.12.1
     , dlist >=0.6.0.1
--- a/src/Futhark/LSP/Compile.hs
+++ b/src/Futhark/LSP/Compile.hs
@@ -7,14 +7,21 @@ module Futhark.LSP.Compile (tryTakeStateFromIORef, tryReCompile) where
 import Control.Monad.IO.Class (MonadIO (liftIO))
 import Data.IORef (IORef, readIORef, writeIORef)
 import qualified Data.Map as M
+import Data.Maybe (fromMaybe)
 import qualified Data.Text as T
 import Futhark.Compiler.Program (LoadedProg, lpFilePaths, lpWarnings, noLoadedProg, reloadProg)
 import Futhark.LSP.Diagnostic (diagnosticSource, maxDiagnostic, publishErrorDiagnostics, publishWarningDiagnostics)
-import Futhark.LSP.State (State (..), emptyState)
+import Futhark.LSP.State (State (..), emptyState, updateStaleContent, updateStaleMapping)
+import Futhark.LSP.Tool (computeMapping)
 import Futhark.Util (debug)
 import Language.Futhark.Warnings (listWarnings)
-import Language.LSP.Server (LspT, flushDiagnosticsBySource, getVirtualFiles)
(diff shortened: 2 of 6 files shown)


src/Futhark/Analysis/MigrationTable.hs
```diff
@@ -28,8 +28,8 @@
 -- by "Futhark.Pass.ReduceDeviceSyncs".
 --
 -- For details on how the graph is constructed and how the vertex cut is found,
--- see the master thesis "Eliminating Synchronous GPU Memory Transfers" by
--- Philip Børgesen (2022).
+-- see the master thesis "Reducing Synchronous GPU Memory Transfers" by Philip
```
PhilipBorgesen, line 31: The thesis will be available later this summer.

athas:
Can you please make a feature branch instead?  Git goes nuts trying to handle multiple masters.

PR #1658, opened 2022-04-26
Reduce synchronous GPU memory transfers

### LINKED ISSUE #1045: Optimising scalar computation interspersed between GPU operations

PhilipBorgesen:
A pull request as agreed upon with @athas. Fixes #1045.

The main additions and changes of interest:
* Adds a sequential GPUBody kernel to the GPU representation, implemented by @athas.
* Implemented merging of GPUBody kernels and other simplifications within the GPU pipeline.
* The GPU pipeline now moves statements into GPUBody kernels to reduce synchronous GPU memory transfers. The model used to drive the transformations favor the OpenCL compiler backend.
* Array indexing done by for-in loops can now be sunk.
* Indexing into single-element arrays are no longer hoisted from kernels.

For details see the module descriptions of:
* [src/Futhark/Optimise/MergeGPUBodies.hs](https://github.com/PhilipBorgesen/futhark/tree/delivery/src/Futhark/Optimise/MergeGPUBodies.hs)
* [src/Futhark/Optimise/ReduceDeviceSyncs.hs](https://github.com/PhilipBorgesen/futhark/tree/delivery/src/Futhark/Optimise/ReduceDeviceSyncs.hs)
* [src/Futhark/Optimise/ReduceDeviceSyncs/MigrationTable.hs](https://github.com/PhilipBorgesen/futhark/tree/delivery/src/Futhark/Optimise/ReduceDeviceSyncs/MigrationTable.hs)
* [src/Futhark/Optimise/ReduceDeviceSyncs/MigrationTable/Graph.hs](https://github.com/PhilipBorgesen/futhark/tree/delivery/src/Futhark/Optimise/ReduceDeviceSyncs/MigrationTable/Graph.hs)

Preliminary benchmarking of [futhark-benchmarks](https://github.com/diku-dk/futhark-benchmarks) programs affected by the changes show mean runtime improvements between 10 and 50 percent of those programs that are affected, depending on hardware. The bars are the average improvement over the datasets for each program entry point. A height of 1.0 means 100% better, i.e. twice as fast as the baseline.
billede
billede
billede
billede


PhilipBorgesen:
New pull request from a feature branch as requested by @athas.

athas:
I forgot, what was the story with the additional synchronisation in the CUDA backend?  (Strictly speaking unrelated to this PR, I know.)

PhilipBorgesen:
> I forgot, what was the story with the additional synchronisation in the CUDA backend? (Strictly speaking unrelated to this PR, I know.)

Let's take the Gauss-Jordan elimination example from #1045:
```
  -- argmax returns (m, i) where m is the largest value in arr and i its index.
  let argmax arr =
    reduce_comm (\(a,i) (b,j) ->
      if      a < b then (b, j)
      else if b < a then (a, i)
      else if j < i then (b, j)
      else               (a, i)
    ) (0f32, 0) (zip arr (indices arr))

  -- gauss_jordan performs Gauss-Jordan elimination without pivoting.
  let gauss_jordan [m] [n] (A:[m][n]f32) =
    loop A for i < i64.min m n do
      -- Find the nonzero value of greatest magnitude.
      let j = A[i:,i] |> map f32.abs |> argmax |> (.1) |> (+i)
      let f = (1 - A[i,i]) / A[j,i]
      let irow = map2 (f32.fma f) A[j] A[i]
      in tabulate m (\j ->
          let f = f32.neg A[j,i]
          in map2 (\x y ->
            if j == i then x else f32.fma f x y
          ) irow A[j]
        )
```
`argmax` runs a reduction that cannot fail (no assertions). The resulting read that is used to compute `j` is implemented as follows under CUDA (similar occurs twice more to compute `f`):
```
int64_t defunc_2_reduce_comm_res_7153;
        int64_t read_res_7949;
        
        {
            cudaEvent_t *pevents = NULL;
            
            if (ctx->profiling && !ctx->profiling_paused) {
                pevents = cuda_get_events(&ctx->cuda,
                                          &ctx->copy_scalar_from_dev_runs,
                                          &ctx->copy_scalar_from_dev_total_runtime);
                CUDA_SUCCEED_FATAL(cudaEventRecord(pevents[0], 0));
            }
            CUDA_SUCCEED_OR_RETURN(cuMemcpyDtoH(&read_res_7949, mem_7783.mem +
                                                (int64_t) 0 * sizeof(int64_t),
                                                sizeof(int64_t)));
            if (pevents != NULL)
                CUDA_SUCCEED_FATAL(cudaEventRecord(pevents[1], 0));
        }
        if (futhark_context_sync(ctx) != 0)
            return 1;
        defunc_2_reduce_comm_res_7153 = read_res_7949;
```
`cuMemcpyDtoH` is a synchronous read. Afterwards `if (futhark_context_sync(ctx) != 0) return 1;` is executed, and `futhark_context_sync` blocks until the entire command queue has finished (the `cuCtxSynchronize()`):
```
    CUDA_SUCCEED_OR_RETURN(cuCtxPushCurrent(ctx->cuda.cu_ctx));
    CUDA_SUCCEED_OR_RETURN(cuCtxSynchronize());
    if (ctx->failure_is_an_option) {
        int32_t failure_idx;
        
        CUDA_SUCCEED_OR_RETURN(cuMemcpyDtoH(&failure_idx, ctx->global_failure,
                                            sizeof(int32_t)));
        ...
```
So its two synchronizations for every read, and the entire pipeline is emptied. I don't know why it is implemented this way (there likely is a good reason) but under OpenCL `futhark_context_sync` is only called after a read if `ctx->failure_is_an_option` is `true`. Also, if `ctx->failure_is_an_option` is `true` the value will be read asynchronously, the `failure_idx` in `futhark_context_sync` will be read asynchronously, and then the program waits for the pipeline to finish (thus finishing the reads as well). That is one synchronous operation for every read, instead of 2 or 3 as under CUDA.

OpenCL also stores constants in static variables and writes them asynchronously. Under CUDA all values, incl. constants, are written synchronously. Note that `cuMemcpyDtoHAsync` and `cuMemcpyHtoDAsync` variants exist.

I have implemented a transformation that turns `A[i] = x` into `A[i] = gpu { x }` but only if `x` is a scalar variable. If `x` is a constant then the OpenCL implementation is faster than running a kernel to transfer `x` to device. This means that the CUDA backend still performs synchronous writes, despite that they entirely are eliminated under OpenCL.

So in summary, I believe these three optimizations would do good for the CUDA backend:
* Only call `futhark_context_sync` if a kernel could have failed, thus eliminating pipeline flushes.
* When a kernel could have failed, read the array value and `failure_idx` asynchronously.
* Transfer constants to device asynchronously via `static` variables.

Again, disclaimer, I have not studied why the CUDA backend does as it does, so there might be complications in doing as I suggest. On the surface some of the suggestions appears to be low-hanging fruit.

athas:
Thanks for the reminder.  Those are probably oversights from when the CUDA backend was implemented (the corresponding OpenCL functions are asynchronous).  I'll fix them on master.  It should not affect your measurements, as they are on the OpenCL backend.  I intend to merge this branch once I've taken a quick look at the slowdowns and determined whether there's an easy way to avoid them (but probably not, and they don't truly matter, as we discussed).

PhilipBorgesen:
> It should not affect your measurements, as they are on the OpenCL backend.

No problem. My intention is to include both CUDA and OpenCL measurements in my academic writing but everything is based on v0.21.10, the latest release. I have another branch with the optimizations based on that tag.

src/Futhark/Optimise/ReduceDeviceSyncs/MigrationTable.hs
```diff
@@ -0,0 +1,1622 @@
...
+      -- the total number of reads can hence either decrease, remain the same,
+      -- or even increase, subject to the ordering of reads and kernels that
+      -- perform assertions.
+      --
+      -- Since it is possible to implement the same failure checking scheme as
+      -- OpenCL using asynchronous reads (and doing so would be a good idea!)
+      -- we consider this to be acceptable.
+      --
+      -- TODO: Implement the OpenCL failure checking scheme under CUDA. This
+      --       should reduce the number of synchronizations per read to one.
```
PhilipBorgesen, line 425: @athas: The differences between how reads are handled between OpenCL and CUDA are also documented here. It sounds like you might be looking at the `TODO`.

athas: Yes, that's fixed on master.

athas:
I also modified futhark-benchmarks itself so that mandelbrot and tunnel no longer have the indexing at the end (and hence the slow GPUBody statements).

I think those were the biggest relative slowdowns (even if tiny in absolute terms), but I'll take another look later.

src/Futhark/Optimise/ReduceDeviceSyncs/MigrationTable.hs
PhilipBorgesen: Then I guess the CUDA section of the comment and the headers just can be removed, since the OpenCL section then applies to both backends. Are we maintaining other backends where these changes are relevant for? Alternatively the header can be modified.

PhilipBorgesen:
> I also modified futhark-benchmarks itself so that mandelbrot and tunnel no longer have the indexing at the end (and hence the slow GPUBody statements).
> 
> I think those were the biggest relative slowdowns (even if tiny in absolute terms), but I'll take another look later.

I will keep using the old versions in my benchmarking and writing. They reflect a real inefficiency that is present in the compiler which naturally occurs. I did not find a way a to solve the flaw in the migration model without increasing the runtime complexity considerable. The current algorithm is |VC|*|E| in the worst case where VC is the minimum vertex cut of the migration graph and E is its set of edges. Accurately modelling if branches turned the optimization into a combinational problem with exponential time complexity, if I recall correctly. I deemed that the small inefficiency in some programs would not be worth the increase in compilation time for all programs. These programs reflect that trade-off.

PhilipBorgesen:
I also reran the futhark-benchmark benchmarks on the NVIDIA A100 using a convergence phase, without removing outliers, and without automatic rerunning for collection of low-RSD samples. It worked well and was considerably faster. The updated measurements indicate an overall larger runtime improvement.

I have verified the most extreme benchmarks (both best improvements and worst degrades) by rerunning them.

plot_v0 21 10-A100
plot_v0 21 10-CUDA

PhilipBorgesen:
I have just rerun micro benchmarks for the unconditional transformation `A[i] = x` -> `A[i] = gpu { x }`. I am not sure what have happened but I no longer get runtime improvements. Instead I get a 20-50% decrease, depending on hardware.
However, if the migrated value can be reused for multiple updates, or if the migration can be merged with other GPUBody kernels, then the transformed version is still faster.

I am not sure what the effects are on realistic programs. I don't think the problematic case is common but I consider patching it out. What are your thoughts @athas?

```
-- ==
-- entry: writeVar writeVar2 writeVar2_
-- random input { [1000]i32 i32 }

entry writeVar (A: *[1000]i32) (v: i32): *[1000]i32 =
  let (B, _) =
    loop (A, v) for i < 1000 do
      #[unsafe] (A with [i] = v, v+1)
  in B

entry writeVar2 (A: *[1000]i32) (v: i32): *[1000]i32 =
  let (B, _) =
    loop (A, v) for i < 999 do
      let A = #[unsafe] A with [i] = v
      let A = #[unsafe] A with [i+1] = v
      in (A, v+1)
  in B

entry writeVar2_ (A: *[1000]i32) (v: i32): *[1000]i32 =
  let (B, _) =
    loop (A, v) for i < 999 do
      let A = #[unsafe] A with [i] = v
      let A = #[unsafe] A with [i+1] = v+1
      in (A, v+1)
  in B
```

athas:
It does not look like an optimisation to me.  If the only motivation is microbenchmarks, and the results are this muddled, then I suggest leaving it out.

PhilipBorgesen:
> It does not look like an optimisation to me. If the only motivation is microbenchmarks, and the results are this muddled, then I suggest leaving it out.

I have limited the optimization to where it can prevent a read. Based on futhark-benchmarks it appears to be a rare pattern; most array writes depend on array reads, which the new compiler pass still optimizes to eliminate the reads. In futhark-benchmarks the transformation heavily affected the `github.com/diku-dk/cpprandom/random_tests.fut` module, but I really don't care about optimizing test modules... Outside test code there was a single instance occurring in `rodinia/nn/nn.fut`. The update still depended upon a read, but the read could not be prevented because it was used as an index.

After making the change and rebenchmarking, `rodinia/nn/nn.fut` went from a slowdown of 6% to an improvement of 8% (OpenCL). Under CUDA the improvement was a more modest 2%.

PhilipBorgesen:
> I also modified futhark-benchmarks itself so that mandelbrot and tunnel no longer have the indexing at the end (and hence the slow GPUBody statements).
> 
> I think those were the biggest relative slowdowns (even if tiny in absolute terms), but I'll take another look later.

I see now that we spoke about different things. I was thinking about another benchmark where the only change done by the optimization is copying the result of reductions done in opposite branches, delaying them out of their if statement to no use.

I still intend to report benchmarks for the old versions. They are still valid samples and they do reflect inefficiencies that can come up as a result of the transformation, however small they are.

PR #1667, opened 2022-05-03
Support `vjp` nesting.

zfnmxt:
Work to support nesting of `vjp`. It's still broken, unfortunately (`ad/accx.fut` fails for all `x > 1`). With some jank hacking I managed to get `ad/acc3.fut` and `ad/acc4.fut` to work and validate, but I removed all of that and just kept the barebones stuff that actually works.

The issue with `ad/acc3.fut` and `ad/acc4.fut` basically boils down to the fact don't distinguish between accumulators and lists of accumulators (this causes issues with the non-accumulator typed adjoints of accumulators). I don't know what's wrong with `ad/acc2.fut`, I haven't looked closely at it, it just returns the wrong values.


athas:
Now only acc3.fut fails.

athas:
Should all work now.  Let's wait for CI.

PR #1670, opened 2022-05-04
Graph-based fusion engine

amar-topalovic:
In this version, diagonal fusion is enabled.
LocVolCalib doesn't compile yet, fusion across reshapes are a work in progress.


athas:
I have merged master into this PR to simplify the diff (and resolve the conflicts).  I don't think I modified any of your fusion code.

athas:
I fixed some bugs in the new fusion, and some in the rest of the compiler that the new fusion code merely served to expose.

There are not many problems left.  There are still some missing copies, as well as missing transpose/reshape/stream-related fusion.  But it's looking pretty good now.

athas:
To my delight, your work is actually quite a bit faster than the old fusion pass - mostly because old fusion had to run twice to get everything right.  It doesn't have a huge impact on overall compilation time because fusion was never the bottleneck, but it's lovely to see a minor improvement.

athas:
This is looking really good now.  I still want to refactor a bit of the graph construction which is quite ad hoc when it comes to inner fusion.  There's also still one disabled test, which involves fusing into loops by turning them into sequential streams - but maybe we should just drop that feature.  I don't think any programs need it.

athas:
I think this is about ready to merge.  Any remaining problems can be fixed later.  There are no regressions to speak of on the benchmarks (except for some that are tuning-related), and a few improvements.

PR #1693, opened 2022-06-25
Expose structure of records in C API.

athas:


athas:
This doesn't yet contain any information in the manifest, and is missing the construction functions.  But the major book-keeping obstacles have been fixed.

athas:
This basically works now, but I'm considering whether to add support for it in the server protocol before merging.

PR #1706, opened 2022-07-18
Make rotates manifest.

athas:
Rotates are pretty uncommon in Futhark.  When they do occur, they are
almost always passed to some other subsequent operation (such as
'map') where they will be fused or otherwise simplified away.  It is
exceedingly rare for rotates to be kept around as delayed arrays.  I
have not actually observed it outside of test programs.

As a result, I think it is better to make rotates always manifest.
This simplifies our index function representation, where rotates
currently add a complexity tax on most other operations.

The main downside is that the code generator(s) must now manually
handle rotates whenever we want them to parallel.  This is not a big
problem, because as stated above, rotates are usually simplified away.


athas:
As expected, this seems to have no impact on benchmark performance.

PR #1707, opened 2022-07-18
Add Match construct to IR

### LINKED ISSUE #1393: Add multiway branches to IR
Our IR supports binary if-then-else, which means that source-language `match` expressions turn into sometimes very long if-then-else chains.  Since we use ANF, this also results in many intermediate results.  This makes intermediate and generated code uglier, and the compiler slower.  We should consider replacing `if` by a multiway branching construct to the IR which will have the following properties:

1. Efficient translation of `match`.
2. Translation into C `switch` statements when possible.

It's not so easy to do this as it might seem.  I propose the following construct for a start:

```
condexp ::= "cond" condcase* "else" "->" body
condcase ::= "if" subexp "==" constant "->" body
```

We require a fallback case at the end.  A case `case x == y -> body` matches if `x` is equal to `y`.  We do not support arbitrary comparisons, only strict equality.  We can express `if p then a else b` as `cond if p == true -> a else b`.

A `cond` expression can be compiled to `case` if the `subexp`s being scrutineed in the various cases are all the same.

athas:
This PR replaces `If` with `Match`.  Conceptually simple.  This PR looks monstrous because I had to slay several dragons in ExplicitAllocations along the way.

Closes #1393.


athas:
With this PR, primexp generalisation is also unused, so I am removing it.

athas:
Fixed the performance regression.  The fix is somewhat ugly, but not uglier than what used to be in ExplicitAllocations.  We can factor it into a separate pass later; this PR really wasn't supposed to be about fiddling with ExplicitAllocations.

PR #1713, opened 2022-07-28
Makefile installs futhark in wrong location if destination directory does not exist.

CamdenKuwahara:
If the installation directory does not exist, the futhark binary will be installed as a binary with the name of the destination folder.
For example, by default running `make install` will install futhark at `~/.local/bin/futhark`. If the directory `~/.local/bin/` does not exist, it would be installed as `~/.local/bin`. 
This patch fixes this issue by  specifying  the desired installation name in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ build:
 	cabal build
 
 install: build
-	install -D $$(cabal -v0 list-bin exe:futhark) $(PREFIX)/bin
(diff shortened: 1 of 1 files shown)


athas:
Man, I thought the point of `install` was to take care of these stupid bugs...

Thanks for the fix!

PR #1714, opened 2022-07-29
Update description for tokens command

zshipko:
Small fix to correct the description of the `tokens` command


athas:
`futhark thanks`!

PR #1720, opened 2022-08-04
Remove parallel stream SOACs.

athas:
Also removed reduce_stream and map_stream from source language.

The main payoff is that this allows simplification of the SegOp
representation and flattening, because we don't have to represent
streams with weird semantics anymore.


athas:
Cosmin says he's OK with crippling one of his children, so here we go.

PR #1721, opened 2022-08-06
Migration bug fixes

PhilipBorgesen:
This patch fixes three bugs related to code migration introduced by https://github.com/diku-dk/futhark/pull/1658:
1) Multi-element array slice updates did not block parent statements from being migrated, so it was possible for such slice updates to be migrated. For large slice updates, this could degrade performance.
2) The transformation to support partial loop migration was bugged in two ways, which could cause internal compiler errors. For details see commit 181f20ad9216de7c008d129898eeebfcc1f51fab.


PhilipBorgesen:
@athas: I also tried disabling the migration of top-level constant declarations to deal with the minor slowdown of the `poseidon-bench.fut:arity11` benchmark (in futhark-benchmarks). You might recall that this benchmark slowed down a bit across all of the four GPUs (A100, MI100, Radeon Pro 560, Intel HD Graphics 630) that I reported measurements for.

Disabling migration of top-level constant declarations only affected the `poseidon-bench.fut:arity11` and `poseidon-bench.fut:arity8` benchmarks in the suite. I only tested on Radeon Pro 560 (I could not access A100 or MI100) where I found that `poseidon-bench.fut:arity11` became ~500 microseconds faster and `poseidon-bench.fut:arity8` became ~500 microseconds slower.

Based on these inconclusive and insignificant results I have left the migration of top-level constant declarations as is.

PhilipBorgesen (commented):
Note: `blocking6_update.fut`, `loop14_into.fut` and `loop15_into.fut` are the only new tests. The other tests are simply renamed (to keep similarly themed tests adjacent in the number sequence).

athas:
Good fixes, thanks!  I agree with the decision not to tweak migration of top-level constants.

PR #1732, opened 2022-09-27
Refactor SegLevel for GPU representation.

athas:
Now we have three levels: SegThread, SegGroup, and SegThreadInGroup.

Also, the specification of the grid size has been decoupled and made optional.  This allows us to generate flat parallelism without being forced to worry about the low-level details of how many threads to launch (although flattening will still presently mostly do that).  We still have freedom to decide if we wish, for example in tiling, which has also been slightly rewritten.


Munksgaard:
I really need to get my short-circuiting branch merged...

athas:
Yes.

PR #1737, opened 2022-10-13
Add .git to .gitignore

Munksgaard:
The two first lines of our .gitignore explicitly unignore the .git folder.

This creates problems for tools such as ripgrep which rely on parsing the .gitignore file. See for instance the discussion here: https://github.com/BurntSushi/ripgrep/issues/2329

On another note, perhaps we might want to get rid of those two lines? As an alternative, we could codify that the /scratch directory is ignored, and could be used for the same purpose as the root directory currently is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 /*
 !/*/
 
+# The previous lines explicitly unignores .git
+/.git
+
 # Ignore Emacs autosave files.
 
 \#*\#


athas:
I rather like being able to treat the root as a scratch directory.  It's rarely a problem.

Munksgaard:
> It's rarely a problem.

True, but it has bitten me at least once: When trying to rip out z3 from my short-circuiting branch, I used `rg` to search for all occurrences of z3 in the repository. Because of those two lines, it didn't show the entries in `futhark.cabal`. Not a huge problem, I'll admit.

athas:
I use `git grep` in those cases.

PR #1766, opened 2022-11-17
Short circuiting simple maps

### LINKED ISSUE #1765: Short-circuiting should also make scalar maps in-place when possible
The `map`s in these functions should be in-place (the key thing is the uniqueness):

```
def f (xs: *[]i32) = map (+2) xs
def g (xs: *[]i32) (ys: []i32) = map2 (+) xs ys
```

I told @Munksgaard how to go about it while mildly inebriated yesterday night, so he should know how to do it.

Munksgaard:
In some simple cases, use `SegMap` as short-circuit point.

For `SegOp`, we currently only handle `SegMap`, and only under the following circumstances:

 1. The `SegMap` has only one return value/pattern value.

 2. The `KernelBody` contains an `Index` statement that is indexing an array using only the values from the `SegSpace`.

 3. The array being indexed is last-used in that statement, is free in the `SegMap`, is unique or has been recently allocated (specifically, it should not be a non-unique argument to the enclosing function), has elements with the same bit-size as the pattern elements, and has the exact same `IxFun` as the pattern of the `SegMap` statement.

There can be multiple candidate arrays, but the current implementation will always just try the first one.

The first restriction could be relaxed by trying to match up arrays in the `KernelBody` with patterns of the `SegMap`, but the current implementation should be enough to handle many common cases.

Fixes #1765


athas:
What the fuck, how can that possibly work?

src/Futhark/Optimise/ArrayShortCircuiting/ArrayCoalescing.hs
```diff
@@ -1398,33 +1401,33 @@ genSSPointInfoGPUMem
...
+          src `nameIn` last_uses,
+          Just memblock@(MemBlock src_pt _ src_mem src_ixf) <-
+            getScopeMemInfo src scopetab,
+          src_mem `nameIn` last_uses,
           -- The 'alloc' table contains allocated memory blocks, including
           -- unique memory blocks from the enclosing function. It does _not_
           -- include non-unique memory blocks from the enclosing function.
-          dst_mem `M.member` alloc td_env,
-          dst `nameIn` frees,
-          src_ixf == dst_ixf,
```
Munksgaard: Okay, that's just lazy...

src/Futhark/Optimise/ArrayShortCircuiting.hs
```diff
@@ -141,7 +155,7 @@ generalizeIxfun
   (PatElem vname (MemArray _ _ _ (ArrayIn mem ixf)))
   m@(MemArray pt shp u _) = do
     coaltab <- asks envCoalesceTab
-    if vname `M.member` coaltab
+    if vname `M.member` foldMap vartab coaltab
```
athas, line 158: Looks like an expensive way to search all of them, but I guess it's fine.

src/Futhark/Optimise/ArrayShortCircuiting/ArrayCoalescing.hs
```diff
@@ -87,23 +94,23 @@ emptyBotUpEnv =
 --- Main Coalescing Transformation computes a successful coalescing table    ---
 --------------------------------------------------------------------------------
 
--- | Given a 'Prog' in 'SegMem' representation, compute the coalescing table
--- for each function.
+-- | Given a 'FunDef' in 'SegMem' representation, compute the coalescing table
```
athas, line 97: It really is a Prog, not a FunDef.

src/Futhark/Optimise/ArrayShortCircuiting/ArrayCoalescing.hs
```diff
@@ -1326,30 +1338,120 @@ mkCoalsHelper3PatternMatch pat e lutab td_env successCoals_tab activeCoals_tab i
...
+  Maybe [SSPointInfo]
+genSSPointInfoSeqMem _ _ _ _ _ =
+  Nothing
+
+-- | Given an 'Op', return a list of potential short-circuit points
+--
+-- For 'SegOp', we currently only handle 'SegMap', and only under the following
+-- circumstances:
+--
+--  1. The 'SegMap' has only one return/pattern value.
```
athas, line 1373: How difficult would this be to improve?  Doesn't have to be part of this PR, of course.

src/Futhark/Optimise/ArrayShortCircuiting/LastUse.hs
```diff
@@ -45,41 +46,96 @@ type LastUseOp rep = Op (Aliases rep) -> Names -> LastUseM rep (LUTabFun, Names,
...
-      pure $ consts_lu <> lus
+-- lastUseProg ::
+--   ( Mem rep inner,
+--     CanBeAliased (Op rep)
+--   ) =>
+--   LastUseOp rep ->
+--   Prog (Aliases rep) ->
+--   LUTabFun
+-- lastUseProg onOp prog =
+--   runLastUseM onOp (m prog)
```
athas, line 96: Commented-out code.

src/Futhark/Optimise/ArrayShortCircuiting/LastUse.hs
```diff
@@ -45,41 +46,96 @@ type LastUseOp rep = Op (Aliases rep) -> Names -> LastUseM rep (LUTabFun, Names,
...
+    CanBeAliased (Op rep),
+    HasMemBlock (Aliases rep)
+  ) =>
+  Names ->
+  FunDef (Aliases rep) ->
+  LastUseM rep (LUTabFun, Names)
+lastUseFun bound_in_consts f =
+  inScopeOf f $
+    flip lastUseBody (mempty, bound_in_consts) $
+      funDefBody f
```
athas, line 130: `inSCopeOf f $ lastUseBody (funDefBody f) (mempty, bound_in_consts)`

It's not like this convoluted use of `flip` gets you out of using parens anyway.

src/Futhark/Optimise/ArrayShortCircuiting/LastUse.hs
```diff
@@ -205,31 +291,32 @@ lastUseExp (Match _ cases body _) used_nms = do
...
-      fpar_nms = namesFromList $ map (identName . paramIdent . fst) var_ses
-      used_nms' = (free_in_body <> freeIn (map snd var_ses)) `namesSubtract` fpar_nms
-      used_nms_res = used_nms0 <> used_nms' <> freeIn (bodyResult body)
-
-      -- the last-uses at loop-statement level are the loop free variables that
-      -- do not belong to @used_nms0@; this includes the initializers of b), @lu_ini_b@
-      lu_arrs = used_nms' `namesSubtract` used_nms0
-  pure (lutab_res, lu_arrs, used_nms_res)
+lastUseExp (DoLoop var_ses lf body) used_nms0 =
+  inScopeOf lf $ do
```
athas, line 295: You can make this diff smaller by moving this to the same line as the `=`.

athas (changes requested):
I'm pretty comfortable with these changes.  I did not closely study changes to the short-circuiting transform, because I do not understand it well in the first place.

src/Futhark/Optimise/ArrayShortCircuiting/ArrayCoalescing.hs
Munksgaard: We would have to attempt to match candidate sources to the patterns in the destination. We would also need to pass in the `InhibitTab` from `BotUpEnv` in order to keep track of which combinations have failed so far. I don't think it would be that hard to do.

athas (approved):

athas:
Oof, you didn't squash.  This one will make for rough bisection in the future.

PR #1767, opened 2022-11-18
Run tests on all pull requests, and pushes to master

Munksgaard:


athas:
How does this differ from what is currently happening?

Munksgaard:
Currently, all tests are run twice when a new commit is pushed to a PR: once because of the `push` directive, and once because of the `pull_request` directive. With this change, CI will only run once for each commit to a PR (including the first time the PR is submitted).

The downside is that pushes to branches without an associated PR will not be CI tested.

PR #1771, opened 2022-11-22
Extend short-circuiting to MCMem.

athas:
Currently fails for map14.fut - it's not clear to me why, but I suspect it's a name collision issue.  It works for GPUMem, though.


athas:
It fails for the best backend.  I'll take a look.

PR #1773, opened 2022-11-24
Lotte and Ulrik's work refering to rev-AD implementation for s

coancea:
…an and histos together with tests


Munksgaard:
Wtf...

athas:
Fails on the ISPC backend.  My hunch is that we may be generating weird scans of a kind the ISPC backend has never seen before.

athas:
Some of these test programs are way too large and have tons of unused entry points.  I'd really like to clean this up because otherwise it will never happen, and they will slow down the test suite forever.

athas:
Actually it's not so bad.  It's only the two "benchtests" files that are messy.

PR #1788, opened 2022-12-08
Fix multicore codegen for scan.

### LINKED ISSUE #1777: soacs/scan6.fut fails with multicore backend on large machines
It is because of memory short circuiting, but I'm not sure exactly why.  Probably some subtlety related to how we use memory in the array-operand case for scan code generation.

athas:
The old codegen was truly weird, and executed the kernel body multiple times.  There is still room to improve, but it is much saner now.

Fixes #1777.


athas:
Note sure I took care of ISPC, but I guess time will tell.

PR #1790, opened 2022-12-09
MemRefAggreg: Include current destination uses when checking overlap

### LINKED ISSUE #1789: clEnqueueCopyBuffer failed with error code -8 (Memory copy overlap)
The following program
```
entry main [n] (xs: *[n]f32) : [n]f32 =                                                                                                                        
  let i = n/3                                                                                                                                                  
  in xs with [:2*i] = copy xs[i:] 
```
compiled with backend `opencl` fails at runtime (input `[1,2,3,4,5,6,7,8,9]`) with error:
```
./test: test.c:7862: OpenCL call
  clEnqueueCopyBuffer(ctx->opencl.queue, xs_mem_6179.mem, xs_mem_6179.mem, (size_t) (i_6149 * (int64_t) 4), (size_t) (int64_t) 0, (size_t) (j_m_i_6150 * (int64_t) 4), 0, ((void *)0), ctx->profiling_paused || !ctx->profiling ? ((void *)0) : opencl_get_event(&ctx->opencl, &ctx->copy_dev_to_dev_runs, &ctx->copy_dev_to_dev_total_runtime))
failed with error code -8 (Memory copy overlap)
```
That is, an in-place update where the reading and writing parts overlap causes a crash, despite the use of `copy`.

Munksgaard:
Fixes #1789

### DIFF
--- a/src/Futhark/Optimise/ArrayShortCircuiting/MemRefAggreg.hs
+++ b/src/Futhark/Optimise/ArrayShortCircuiting/MemRefAggreg.hs
@@ -228,7 +228,7 @@ recordMemRefUses td_env bu_env stm =
             if m_b `nameIn` original_mem_aliases
               then (wrt_lmads' <> lmads'', Set mempty)
               else (wrt_lmads', lmads'')
-          no_overlap = noMemOverlap td_env prev_use wrt_lmads''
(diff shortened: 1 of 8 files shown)


athas:
Why a draft?

Munksgaard:
Not on purpose

PR #1810, opened 2022-12-21
futhark-literate: Add :audio directive

Munksgaard:
So far it only handles 8-bit audio at 44100 Hz, but we should be able to extend it if needed.

### DIFF
--- a/docs/man/futhark-literate.rst
+++ b/docs/man/futhark-literate.rst
@@ -32,7 +32,7 @@ programming techniques.
 **Warning:** Do not run untrusted programs.  See SAFETY below.
 
 Image directives and builtin functions shell out to ``convert`` (from
-ImageMagick).  Video generation uses ``ffmpeg``.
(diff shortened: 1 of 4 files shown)


Munksgaard:
I will write a literate example for the Futhark website at some point.

athas:
Does the test also verify that the tone is pleasant?

Munksgaard:
I am unsure how best to quantify the pleasentness of a sound. Perhaps we need a test harness with someone strapped into it, who will be exposed to the result of this test everytime something is pushed to the repository.

athas:
Two obvious next steps:

1. Add a codec option so you can get ffmpeg to compress the audio.  Wave is not really usable for Internet distribution.
2. Support chunking, like the video command, so the entire sound does not have to fit in memory at once.  (Maybe not important; I have no idea how much space sound consumes.)

Munksgaard:
Also 32 bit and 16 bit audio (http://trac.ffmpeg.org/wiki/audio%20types) and specifying a different frequency.

Issue #1799, opened 2022-12-15
Parallelise copy in multicore backend


athas:
This one should be really easy (ignoring locality considerations).

Issue #1768, opened 2022-11-19
Improve multicore allocator


athas:
Currently we use plain `malloc()/free()`, but this is extremely slow for large allocations, as these will be freshly memory-mapped.

PR #1817, opened 2022-12-25
Improvements to the multicore backend

athas:
Starting with #1799, but I will also address #1768.


athas:
This also helps the ISPC backend, for that matter.

athas:
I'm slightly worried about our GPU-oriented free list semantics being a bad fit on multicores, though.  Particularly the property of not removing unused stuff from the free list until allocations start failing.  Can get awfully antisocial on a large system...

Issue #1820, opened 2022-12-27
Request: add log1p in math prelude


nbos:
`log1p(x)` computes accurately `log(1 + x)` even when `x` is very small. 

It looks like OpenCL and CUDA have it as primitives, although I can't find reference to a 16-bit CUDA `log1p`.

I can give it a go if it turns out just to require appending definitions to the prelude.

athas:
Sure, have at it. The main challenge is getting it done for all the backends, but if you add a test to tests/primitive, CI will tell you the ones you missed. For the 16 bit version, I would just use 32 bit arithmetic - we do that with many others. Let me know if you need more pointers.

nbos:
How bad is it if results vary by 1 or 2 ULPs between backends? E.g. Haskell's `log1p 2.0` gives `1.0986122886681096` to the interpreter while my implementation for ISPC gives `1.0986122886681098`. Do I just omit those cases in tests?

athas:
The testing tool uses a fairly generous tolerance when comparing.

PR #1821, opened 2022-12-28
Add log1p

nbos:
Works fine in interpreter, but when compiled with backend `c`, `multicore`, even `python`, it underflows to 0.0 for small (absolute) values, as if it were implemented as `log1p x = log (1 + x)`


nbos:
#1820

athas:
Nice work.  All that is missing is tests, right?

nbos:
The tests pass, but calling it manually it still looks like it underflows:
Compile `entry main (x: f32) : f32 = f32.log1p x` with any CPU backend and pipe `echo 1e-12 | ./test`; it returns `0.000000f32` instead of expected `1e-12f32`.

athas:
Isn't t that just because the text printer uses only six digits? Try -b and inspect the raw bytes.

nbos:
You're right it's the printer, even `id` turns `1e-12` into `0.000000f32`

athas:
Do you have a name for the changelog entry or will you just go by nbos?

nbos:
I'll go by nbos

athas:
Great, thanks for the contribution!

athas:
Thanks for the contribution!

PR #1828, opened 2023-01-03
Run on own action.

athas:


athas:
@WilliamDue This is the PR that switches the workflow to use self-hosted runners.

athas:
Our machines are much faster than GitHub's, so maybe we should use them for most of the jobs...

athas:
Also note: this is only safe because I have configured the repository to require approval before running workflows for *every* pull request by an outside contributor.  That sort of sucks, but I couldn't find any other way to do it.

Munksgaard:
As long as they don't clog up the machines too much so that we can't run our own experiments on them...

athas:
This shouldn't clog up anything more than Buildbot currently does.

.github/workflows/main.yml
```diff
@@ -551,36 +543,30 @@ jobs:
...
         cd futhark-benchmarks && pwd && ./get-data.sh external-data.txt
-
-    - run: |
-        futhark test -C --no-terminal --backend=opencl  --pass-compiler-option=--Werror futhark-benchmarks
+    - name: Benchmark OpenCL
+      run: |
+        futhark bench futhark-benchmarks --no-terminal --backend=opencl --json futhark-opencl-A100-$GITHUB_SHA.json
+    - name: Benchmark CUDA
+      run: |
+        futhark bench futhark-benchmarks --no-terminal --backend=cuda --json futhark-cuda-A100-$GITHUB_SHA.json
```
Munksgaard, line 569: Now that we're actually doing this, we could just push directly to https://futhark-lang.org/benchmark-results/, right?

Munksgaard:
I think it will? Maybe not a lot, but currently buildbot doesn't test PRs, right?

.github/workflows/main.yml
athas: Yes, that is the next step.

athas:
We don't get that many PRs.  Also, Buildbot currently recompiles a lot.  This PR uses GitHub workers to actually compile Futhark, and then just copies the binary to our machines.

One thing we should perhaps do is only run the GPU jobs when the PR has been reviewed, to cut down on unnecessary work on drafts and such.

Munksgaard:
> One thing we should perhaps do is only run the GPU jobs when the PR has been reviewed, to cut down on unnecessary work on drafts and such.

Yeah, that makes sense.

PR #1831, opened 2023-01-05
Remove Analysis.LastUse and use ArrayShortCircuiting.LastUse instead

Munksgaard:
This seems to be safe. I'm annoyed that I don't remember why this didn't work in the past, or perhaps we only ever tried to do it the other way around.


athas:
It would be delightful if we could have just a single of these analyses.  It would also simplify the narrative in case anyone ever decides to write a dissertation about the subject.  No idea why I'm thinking of that.

athas (approved):

PR #1839, opened 2023-01-13
futhark pkg: use 'git' directly.

### LINKED ISSUE #1743: Idempotency of futhark pkg upgrade.
The timestamp generated in the `futhark.pkg` file by `futhark pkg upgrade` seems to be the time of the call, and not the time of the referenced commitment.

Is this an oversight, or intentional?

### LINKED ISSUE #1835: More flexibility in futhark pkg paths
Currently, futhark pkg places very specific constraints on the urls of packages which it adds. Currently, only packages which are:

- on github or gitlab
- public

are permitted. Additionally, the urls themselves are quite strict. At my institution, we have our own instance of gitlab hosted locally. In our set up, repos are grouped which results in urls that may contain more than just the user/organization name. For example:

https://hostname.com/organization/collection/futhark-package.git

A complimentary limitation is the requirement that packages be hosted publicly. When I tried to move a package to a github url but kept private, I still couldn't use futhark pkg.

Removing either one of these constraints help our use case quite a bit.

Thanks!

athas:
Fixes #1835.

Fixes #1743.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * You can now mix consumption and higher-order terms in slightly more
   cases (#1836).
 
+* `futhark pkg` now invokes Git directly rather than scraping
+  GitHub/GitLab.  This means package paths can now refer to any Git
+  repository, as long as `git clone` works.  In particular, you can
+  use private and self-hosted repositories.
+
 ### Removed
 
 ### Changed
--- a/docs/man/futhark-pkg.rst
+++ b/docs/man/futhark-pkg.rst
@@ -53,8 +53,8 @@ afterwards.
 Most commands take a ``-v``/``--verbose`` option that makes
 ``futhark pkg`` write running diagnostics to stderr.
 
-Network requests (exclusively HTTP GETs) are done via ``curl``, which
-must be available on the ``PATH``.
+Packages must correspond to Git repositories, and all interactions are
+done by invoking ``git``.
 
 COMMANDS
 ========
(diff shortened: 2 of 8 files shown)


athas:
@poeface can you check this out and see whether it works for your purposes?  If you cannot compile the compiler yourself, here'll be a prebuilt tarball once CI is done.

poeface:
@athas, Thanks so much for your help on this! I was able to compile and install this version of futhark. Once I set up an ssh key with our university gitlab instance and added the url conversion instructions to my .gitconfig which you recommended in the issue thread, I was able to run `futhark pkg add` with the repo. That ran just fine! However, `futhark pkg sync` failed with the following message:

     futhark: /tmp/futhark-pkg-4e7eb3b1d0bddae5/gitzipsuzieduzscearlzsutilitieszsfutharkzmmath/lib/git.psu.edu/cearl/utilities/futhark-math: getDirectoryContents:openDirStream: does not exist (No such file or directory)

I took a look at my /tmp folder during the `add` operation, and sure enough a futhark-pkg-##### directory IS being created. However, once the operation finishes, the directory disappears. Let me know if you need any more information or need me to try anything else.

athas:
The directory is supposed to be deleted at the end.  However, it appears that some files are missing even before it is deleted.  Could you run `futhrak pkg add -v` and provide the output?

athas:
Also, what is the contents of `futhark.pkg` of the package you're trying to add?  I suspect the package path might be wrong.

poeface:
You were right! I had configured the target package wrong. When I went back and redid the steps you provide in the documentation, add and sync worked like a charm! Let me know if you have more tests you'd like me to run on my end, but otherwise this update meets our needs.

athas:
I'll improve the handling/detection of incorrect package paths, but I think this is otherwise operational.

athas:
ACtually, what did your invalid `futhark.pkg` contain?  I cannot figure out how to reproduce the crash you observed.

athas:
Ah, I think I got it.  The problem was that the `lib` directory didn't contain the proper directory, right?

Issue #1408, opened 2021-07-21
Make it possible to run program from stdin or args


Munksgaard:
In python, you can write `python -c 'print(42)'` to start an interpreter on the input. Similarly, it would be nice to be able to write `futhark run -c 'iota 42'` or `echo "iota 42" | futhark run -` to run the interpreter on some simple input.

athas:
This should be pretty easy - I'd look at the implementation of `futhark repl` to see how to type-check and run arbitrary expressions.

Munksgaard:
Actually, upon further consideration, we should probably not allow the `echo "iota 42" | futhark run -` form, since standard input is usually used for input to the program being run, right?

athas:
Yes.  Also, using `-` for stdin is a relic of (never-existent?) primordial times; we have a word for stdin: `/dev/stdin`.

athas:
But anyway, just `-c exp` is what we should support (too bad `-e` is already taken for something else).

razetime:
[from chat on gitter] since `futhark run` completely changes  behaviour when `-c` is provided, `futhark eval` is a possible subcommand this functionality can come under.

athas:
For this hypothetical `futhark eval`, I suggest that it *optionally* takes a file to load before evaluating expressions, and then any number of expressions.  Usage examples:

```
$ futhark eval '2+2'
4
$ futhark eval '2+2' '1+1'
4
2
$ futhark eval -f foo.fut 'some_function_from_foo 2'
1.41421356237
```

I don't like that `futhark eval` differs from `futhark run` and `futhark eval` (and pretty much every other command) in taking the file as a named option instead of a positional one.  Perhaps the file should be positional and the expressions passed as options?

Munksgaard:
Would it make sense to have a `--backend` argument to `futhark eval`?

Munksgaard:
> ```
> 
> ```
> 
> 
>     
>   
> 
> I don't like that `futhark eval` differs from `futhark run` and `futhark eval` (and pretty much every other command) in taking the file as a named option instead of a positional one.  Perhaps the file should be positional and the expressions passed as options?

I think in this case it is reasonable enough to take the file as a named option. Surely, passing expressions is the norm, while passing files is the exception?

athas:
> Would it make sense to have a --backend argument to futhark eval?

It uses the interpreter.

> Surely, passing expressions is the norm, while passing files is the exception?

[...]

PR #1844, opened 2023-01-17
Add futhark eval command

razetime:
For issue #1408.

### DIFF
--- a/docs/man/futhark.rst
+++ b/docs/man/futhark.rst
@@ -66,11 +66,11 @@ A Futhark compiler development command, intentionally undocumented and
 intended for use in developing the Futhark compiler, not for
 programmers writing in Futhark.
 
-futhark eval options... <exprs...>
+futhark eval [-f FILE] [-w] <exprs...>
 ----------------------------------
 
 Evaluates expressions given as command-line arguments. Optionally 
-allows a file import.
+allows a file import using ``-f``.
 
 futhark hash PROGRAM
 --------------------
--- a/docs/man/futhark.rst
+++ b/docs/man/futhark.rst
@@ -66,6 +66,12 @@ A Futhark compiler development command, intentionally undocumented and
 intended for use in developing the Futhark compiler, not for
 programmers writing in Futhark.
 
+futhark eval options... <exprs...>
+----------------------------------
+
+Evaluates expressions given as command-line arguments. Optionally 
+allows a file import.
+
 futhark hash PROGRAM
 --------------------
 
--- a/docs/man/futhark.rst
+++ b/docs/man/futhark.rst
@@ -67,7 +67,7 @@ intended for use in developing the Futhark compiler, not for
 programmers writing in Futhark.
 
 futhark eval [-f FILE] [-w] <exprs...>
-----------------------------------
+--------------------------------------
 
 Evaluates expressions given as command-line arguments. Optionally 
 allows a file import using ``-f``.
--- a/src/Futhark/CLI/Eval.hs
+++ b/src/Futhark/CLI/Eval.hs
@@ -8,11 +8,9 @@ import Data.Maybe
 import Data.Text qualified as T
 import Futhark.Compiler
 import Futhark.MonadFreshNames
--- import Futhark.Util (toPOSIX)
 
 import Futhark.Pipeline
 import Futhark.Util.Options
--- import Language.Futhark.Interpreter (interpretExp, ExtOp)
 
 import Futhark.Util.Pretty
 import Language.Futhark
@@ -26,7 +24,6 @@ import System.IO
 import Prelude
 
 main :: String -> [String] -> IO ()
--- main _ _ = putStrLn "Hello World"
 main = mainWithOptions interpreterConfig options "options... <exprs...>" run
   where
     run [] _ = Nothing
@@ -119,14 +116,7 @@ newFutharkiState cfg maybe_file = runExceptT $ do
     badOnLeft T.prettyTypeError . snd $
       T.checkDec imports src T.initialEnv imp $
         mkOpen "/prelude/prelude"
-  -- (tenv2, d2, src'') <-
-  --   badOnLeft T.prettyTypeError . snd $
-  --     T.checkDec imports src' tenv1 imp $
-  --       mkOpen $
-  --         toPOSIX $
-  --           dropExtension file
   ienv2 <- badOnLeft I.prettyInterpreterError =<< runInterpreter' (I.interpretDec ienv1 d1)
-  -- ienv3 <- badOnLeft I.prettyInterpreterError =<< runInterpreter' (I.interpretDec ienv2 d2)
   pure (src', tenv1, ienv2)
   where
     badOnLeft :: (err -> err') -> Either err a -> ExceptT err' IO a
--- a/src/Futhark/CLI/Eval.hs
+++ b/src/Futhark/CLI/Eval.hs
@@ -106,13 +106,10 @@ newFutharkiState ::
   Maybe FilePath ->
   IO (Either (Doc AnsiStyle) (VNameSource, T.Env, I.Ctx))
 newFutharkiState cfg maybe_file = runExceptT $ do
-  files <- case maybe_file of
-    Just f -> pure [f]
-    Nothing -> pure []
   (ws, imports, src) <-
     badOnLeft prettyCompilerError
       =<< liftIO
-        ( runExceptT (readProgramFiles [] files)
+        ( runExceptT (readProgramFiles [] $ maybeToList maybe_file)
             `catch` \(err :: IOException) ->
               pure (externalErrorS (show err))
         )
--- a/src/Futhark/CLI/Eval.hs
+++ b/src/Futhark/CLI/Eval.hs
@@ -39,13 +39,7 @@ runExprs exprs cfg = do
       hPutStrLn stderr $ fromJust file <> ": file not found."
       exitWith $ ExitFailure 2
     Right s -> pure s
-  runExprs' exprs src env ctx
-
-runExprs' :: [String] -> VNameSource -> T.Env -> I.Ctx -> IO ()
-runExprs' [] _ _ _ = pure ()
-runExprs' (x : xs) src env ctx = do
-  runExpr src env ctx x
-  runExprs' xs src env ctx
+  mapM_ (runExpr src env ctx) exprs
 
 -- Use parseExp, checkExp, then interpretExp.
 runExpr :: VNameSource -> T.Env -> I.Ctx -> String -> IO ()
@@ -60,8 +54,7 @@ runExpr src env ctx str = do
       hPutDoc stderr $ I.prettyTypeError terr
       exitWith $ ExitFailure 1
     (_, Right (_, e)) -> pure e
-  let ext = I.interpretExp ctx fexp
-  pval <- runInterpreter' ext
+  pval <- runInterpreter' $ I.interpretExp ctx fexp
   case pval of
     Left err -> do
       hPutDoc stderr $ I.prettyInterpreterError err
--- a/src/Futhark/CLI/Eval.hs
+++ b/src/Futhark/CLI/Eval.hs
@@ -1,35 +1,29 @@
 module Futhark.CLI.Eval (main) where
-  
+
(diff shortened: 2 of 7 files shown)


src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,155 @@
...
+
+runExprs :: [String] -> InterpreterConfig -> IO ()
+runExprs exprs cfg = do
+  -- putStrLn "Running expressions"
+  let InterpreterConfig _ file = cfg
+  -- putStrLn "File has come up."
+  maybe_new_state <- newFutharkiState cfg file
+  -- putStrLn "New state made"
+  (src, env, ctx) <- case maybe_new_state of
+    Left _ -> error "Unable to load file."
```
athas, line 49: This should not be an `error` (which is used only for internal bugs), but `hPutStrLn stderr` or something similar.

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,155 @@
...
+
+
+-- Use parseExp, checkExp, then interpretExp.
+runExpr :: VNameSource -> T.Env -> I.Ctx -> String -> IO ()
+runExpr src env ctx str = do
+  uexp <- case parseExp "" (T.pack str) of 
+    Left _ -> error "Syntax Error"
+    Right e -> pure e
+  let (_, expdat) = T.checkExp [] src env uexp
+  fexp <- case expdat of
```
athas, line 65: Do this directly in the `case`: `case TC.checkExp [] src env uexp of (_, Left _) -> ...`

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,155 @@
...
+    Left _ -> error "Type Error"
+    Right (_,e) -> pure e
+  let ext = I.interpretExp ctx fexp
+  pval <- runInterpreter' ext
+  val <- case pval of
+    Left _ -> error ""
```
athas, line 71: Same as above; don't bind before pattern-matching and don't use `error`.

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,155 @@
...
+      "Do not print warnings."
+  ]
+
+
+newFutharkiState ::
+  InterpreterConfig ->
+  Maybe FilePath ->
+  IO (Either (Doc AnsiStyle) (VNameSource, T.Env, I.Ctx))
+newFutharkiState cfg maybe_file = runExceptT $ do
+  files <- case maybe_file of
```
athas, line 108: There is a function `maybeToList` for doing this.

athas (changes requested):
This is on the right track, and most of the problems are superficial.  You should definitely run the style checker locally.  I'd also like to see the command listed [here](https://github.com/diku-dk/futhark/blob/master/docs/man/futhark.rst) and perhaps a test added [here](https://github.com/diku-dk/futhark/tree/master/tests_adhoc).

src/Futhark/CLI/Eval.hs
razetime: I've added an `exitWith` as well.

src/Futhark/CLI/Eval.hs
razetime: done.

src/Futhark/CLI/Eval.hs
razetime: I am getting an error when I bind during pattern matching:
```
  val <- case runInterpreter' ext of
    Left _ -> error ""
    Right (Left _) ->  error ""
    Right (Right x) -> pure x
```
```
• No instance for (MonadIO (Either a0))
    arising from a use of ‘runInterpreter'’
• In the expression: runInterpreter' ext
  In a stmt of a 'do' block:
    val <- case runInterpreter' ext of
             Left _ -> error ""
             Right (Left _) -> error ""
             Right (Right x) -> pure x
  In the expression:
    do uexp <- case parseExp "" (T.pack str) of
                 Left _ -> error $ "Syntax Error:" <> str
                 Right e -> pure e
       fexp <- case T.checkExp [] src env uexp of
                 (_, Left _) -> error $ "Type Error:" <> str
                 (_, Right (_, e)) -> pure e
       let ext = I.interpretExp ctx fexp
       val <- case runInterpreter' ext of
                Left _ -> error ""
                Right (Left _) -> error ""
                Right (Right x) -> pure x
       ....
```

zfnmxt: I think he means that you shouldn't bind the result of the `case`. For example:
```
case pval of
    Left err -> do
      putDoc $ I.prettyInterpreterError err
      exitWith $ ExitFailure 0
    Right x -> putDoc $ I.prettyValue x <> hardline
```
and similar for the other cases. [Here's an example.](https://github.com/diku-dk/futhark/blob/9780afb82718ddd3c394e902f28e8a73e953044e/src/Futhark/CLI/REPL.hs#L252-L274)

docs/man/futhark.rst
```diff
@@ -66,6 +66,12 @@ A Futhark compiler development command, intentionally undocumented and
 intended for use in developing the Futhark compiler, not for
 programmers writing in Futhark.
 
+futhark eval options... <exprs...>
+----------------------------------
+
+Evaluates expressions given as command-line arguments. Optionally 
+allows a file import.
```
athas, line 73: Document the syntax for the file import.  There are few enough options that you can just spell them out (instead of `options...` put `[-f FILE] [-w]`, or whatever it is).

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,136 @@
...
+  fexp <- case T.checkExp [] src env uexp of
+    (_, Left _) -> error $ "Type Error:" <> str
+    (_, Right (_, e)) -> pure e
+  let ext = I.interpretExp ctx fexp
+  pval <- runInterpreter' ext
+  val <- case pval of
+    Left err -> do
+      putDoc $ I.prettyInterpreterError err
+      exitWith $ ExitFailure 0
+    Right x -> pure x
```
athas, line 64: No need to bind this in `val`; just do the `putDoc` immediately here.

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,136 @@
...
+    Right e -> pure e
+  fexp <- case T.checkExp [] src env uexp of
+    (_, Left _) -> error $ "Type Error:" <> str
+    (_, Right (_, e)) -> pure e
+  let ext = I.interpretExp ctx fexp
+  pval <- runInterpreter' ext
+  val <- case pval of
+    Left err -> do
+      putDoc $ I.prettyInterpreterError err
+      exitWith $ ExitFailure 0
```
athas, line 63: Errors should produce a nonzero exit code.  Also, errors go on stderr (`hPutDoc stderr`).

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,136 @@
...
+runExprs' [] _ _ _ = putStrLn ""
+runExprs' (x : xs) src env ctx = do
+  runExpr src env ctx x
+  runExprs' xs src env ctx
+
+-- Use parseExp, checkExp, then interpretExp.
+runExpr :: VNameSource -> T.Env -> I.Ctx -> String -> IO ()
+runExpr src env ctx str = do
+  uexp <- case parseExp "" (T.pack str) of
+    Left _ -> error $ "Syntax Error:" <> str
```
athas, line 53: Don't use `error`.

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,136 @@
...
+    Right e -> pure e
+  fexp <- case T.checkExp [] src env uexp of
+    (_, Left _) -> error $ "Type Error:" <> str
```
athas, line 56: Don't use `error`.

athas (changes requested):
See HACKING.md or STYLE.md for how to automatically format your code as needed.

src/Futhark/CLI/Eval.hs
razetime: i think this one is realted to binding *before* pattern matching, but i will avoid assigning the `case` expr as well.

docs/man/futhark.rst
razetime: added the individual options

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,155 @@
...
+      T.checkDec imports src T.initialEnv imp $
+        mkOpen "/prelude/prelude"
+  -- (tenv2, d2, src'') <-
+  --   badOnLeft T.prettyTypeError . snd $
+  --     T.checkDec imports src' tenv1 imp $
+  --       mkOpen $
+  --         toPOSIX $
+  --           dropExtension file
+  ienv2 <- badOnLeft I.prettyInterpreterError =<< runInterpreter' (I.interpretDec ienv1 d1)
+  -- ienv3 <- badOnLeft I.prettyInterpreterError =<< runInterpreter' (I.interpretDec ienv2 d2)
```
athas: Remove commented-out code.

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,155 @@
...
+  (tenv1, d1, src') <-
+    badOnLeft T.prettyTypeError . snd $
+      T.checkDec imports src T.initialEnv imp $
+        mkOpen "/prelude/prelude"
+  -- (tenv2, d2, src'') <-
+  --   badOnLeft T.prettyTypeError . snd $
+  --     T.checkDec imports src' tenv1 imp $
+  --       mkOpen $
+  --         toPOSIX $
+  --           dropExtension file
```
athas: Remove commented-out code.

docs/man/futhark.rst
```diff
@@ -66,6 +66,12 @@ A Futhark compiler development command, intentionally undocumented and
 intended for use in developing the Futhark compiler, not for
 programmers writing in Futhark.
 
+futhark eval [-f FILE] [-w] <exprs...>
+----------------------------------
```
athas, line 70: Add more dashes or Sphinx will complain.

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,150 @@
...
+      hPutStrLn stderr $ fromJust file <> ": file not found."
+      exitWith $ ExitFailure 2
+    Right s -> pure s
+  runExprs' exprs src env ctx
+
+runExprs' :: [String] -> VNameSource -> T.Env -> I.Ctx -> IO ()
+runExprs' [] _ _ _ = pure ()
+runExprs' (x : xs) src env ctx = do
+  runExpr src env ctx x
+  runExprs' xs src env ctx
```
athas, line 48: I think this entire thing can be replaced by `mapM_ (runExpr src env ctx) exprs`.

src/Futhark/CLI/Eval.hs
```diff
@@ -0,0 +1,150 @@
...
+      T.hPutStrLn stderr serr
+      exitWith $ ExitFailure 1
+    Right e -> pure e
+  fexp <- case T.checkExp [] src env uexp of
+    (_, Left terr) -> do
+      hPutDoc stderr $ I.prettyTypeError terr
+      exitWith $ ExitFailure 1
+    (_, Right (_, e)) -> pure e
+  let ext = I.interpretExp ctx fexp
+  pval <- runInterpreter' ext
```
athas, line 64: No need to bind `ext`, just do `pval <- runInterpreter' $ I.interpretExp ctx fexp`.

athas (changes requested):
This is looking pretty good now.

docs/man/futhark.rst
razetime: pleased the sphinx

src/Futhark/CLI/Eval.hs
razetime: yep, it works correctly.

athas (approved):

athas:
Oh, one last thing: please add an entry to the CHANGELOG.

athas:
Thanks for the contribution!

athas:
Feel free to write more useful Haskell.

PR #1857, opened 2023-01-23
Less aggressive inlining.

athas:
This commit is the beginning of reworking our inlining policy to be less aggressive.  Currently it inlines all functions called from SOACs that involve arrays in any way.  This is not quite enough: we also need to inline anything that contributes to irregularity.

Given our downstream compilation restrictions, inlining isn't just about performance - it is about whether we can compile at all.

On heston32, the impact of this is pretty promising.  Pretty much unchanged performance, but more than 2x faster compilation.


athas:
I've realised we will also need to augment the autotuner to be able to tune threshold parameters used in callees.  This will require us to expose the call graph of the program in some machine-readable manner.  I think the best place to do this is in the manifest.

Issue #748, opened 2019-05-02
Rewrite cmp-bench-json.py in Haskell


athas:
[cmp-bench-json.py](https://github.com/diku-dk/futhark/blob/master/tools/cmp-bench-json.py) is a small script we use for comparing the output of the `--json` flag to `futhark bench`.  It might make more sense to make it a proper documented `futhark` subcommand instead, as this script is not part of the proper Futhark releases.  Maybe we could also give it a nice (optional) searchable TUI interface, or sort by regressions, or things like that.

athas:
A good starting point would be to define a `FromJSON` instance for [DataResults](https://github.com/diku-dk/futhark/blob/master/src/Futhark/CLI/Bench.hs#L64) to match the existing [toJSON instance](https://github.com/diku-dk/futhark/blob/master/src/Futhark/CLI/Bench.hs#L66).

athas:
There are now already defined `FromJSON` and `ToJSON` instances, so this is even easier!

athas:
This should be a subcommand `futhark bench-compare`.  It is simple: read the two JSON files using the existing deserialisation functions and produce a summary, like the existing script does.

PR #1860, opened 2023-01-30
`cmp-bench-json.py` rewritten in Haskell (Issue #748)

WilliamDue:
In this pull request `cmp-bench-json.py` should be correctly rewritten in Haskell and available as a subcommand.
The usage is the same as `cmp-bench-json.py` so `futhark compare-bench <file> <file>`.
It is most likely not a perfect replication but I was not able to find a difference when comparing the output with `cmp-bench-json.py` output using `diff(1)` and the direct product of these [test files](https://github.com/diku-dk/futhark/files/10539929/tests.zip).

### DIFF
--- a/src/Futhark/CLI/Benchcmp.hs
+++ b/src/Futhark/CLI/Benchcmp.hs
@@ -10,11 +10,6 @@ import Data.Map qualified as M
 import Data.Text qualified as T
 import Data.Vector qualified as V
 import Futhark.Bench
-    ( RunResult(runMicroseconds),
-      Result(memoryMap, runResults),
-      DataResult(..),
-      BenchResult(..),
-      decodeBenchResults )
 import Futhark.Util (showText)
 import Futhark.Util.Options (mainWithOptions)
 import Statistics.Sample qualified as S
@@ -93,7 +88,7 @@ readFileSafely filepath =
 toDataResultsMap :: [DataResult] -> M.Map T.Text (Either T.Text Result)
 toDataResultsMap = M.fromList . fmap toTuple
   where
-    toTuple (DataResult dataset dataResults) = (T.pack dataset, dataResults)
+    toTuple (DataResult dataset dataResults) = (dataset, dataResults)
 
 -- | Converts BenchResults to a Map with the file path as a key.
 toBenchResultsMap ::


src/Futhark/CLI/CompareBench.hs
```diff
@@ -0,0 +1,322 @@
...
+    -- | If the speedup was significant.
+    significant :: Bool
+  }
+  deriving (Show)
+
+-- | Terminal colors used when printing the comparisons. Some of these are not
+-- colors ways of emphasising text.
+data Colors = Colors
+  { -- | The header color.
+    header :: String,
```
athas, line 34: I doubt it matters in practice for a program this small, but I/we have a general principle of never using `String`.  Use the type `Data.Text.Text` instead.  (The fact that the default String type in Haskell is really bad is an unfortunate historical accident).  The main difference is that you'll use `<>` for concatenation instead of `++`.

src/Futhark/CLI/CompareBench.hs
```diff
@@ -0,0 +1,322 @@
...
+    okgreen :: String,
+    -- | Warning color.
+    warning :: String,
+    -- | When something fails.
+    failing :: String,
+    -- | Default color.
+    endc :: String,
+    -- | Bold text.
+    bold :: String,
+    underline :: String -- Underline ^ text.
```
athas, line 47: Should also be a Haddock comment.

src/Futhark/CLI/CompareBench.hs
```diff
@@ -0,0 +1,322 @@
...
+ttyColors =
+  Colors
+    { header = "\ESC[95m",
+      okblue = "\ESC[94m",
+      okgreen = "\ESC[92m",
+      warning = "\ESC[93m",
+      failing = "\ESC[91m",
+      endc = "\ESC[0m",
+      bold = "\ESC[1m",
+      underline = "\ESC[4m"
```
athas, line 61: We can tolerate this, but I'm wondering whether using the proper prettyprinting library (https://github.com/diku-dk/futhark/blob/master/src/Futhark/Util/Pretty.hs) would be better, as we would then be able to strip out these codes when producing e.g. file output.

athas (changes requested):
Looks pretty good, actually!

src/Futhark/CLI/Main.hs
```diff
@@ -77,7 +78,8 @@ commands =
       ("literate", (Literate.main, "Process a literate Futhark program.")),
       ("lsp", (LSP.main, "Run LSP server.")),
       ("thanks", (Misc.mainThanks, "Express gratitude.")),
-      ("tokens", (Misc.mainTokens, "Print tokens from Futhark file."))
+      ("tokens", (Misc.mainTokens, "Print tokens from Futhark file.")),
+      ("compare-bench", (CompareBench.main, "Compare two Futhark benchmarks."))
```
athas, line 82: Should be `benchcmp` for consistency with `datacmp`.

athas (changes requested):
Also add to `docs/man/futhark.rst`.

docs/man/futhark.rst
```diff
@@ -24,6 +24,13 @@ documented below.
 COMMANDS
 ========
 
+futhark benchcmp FILE_A FILE_B
+-----------------------------
```
athas, line 28: Add another dash.

WilliamDue: Oh ups

athas:
Merge master into your branch. You're working from older code.

athas (approved):

PR #1862, opened 2023-01-31
Use Hendrix for CI.

athas:


Munksgaard:
I'm sure they're going to love it!

athas:
This works way more smoothly than I thought it would.

WilliamDue:
I am not sure if it is relevant here but collecting benchmarks results even if one benchmark fails should be achievable with ``${{ always() }}`` as described [here](https://github.com/diku-dk/futhark-ci/blob/slurm/benchmarks.md#collection-of-results) and it was tested [here](https://github.com/SelfHostedRunnerTest/futhark/actions/runs/4029547666).

athas:
Looks pretty decent.  The failure on the TITAN X is because that benchmark required more memory than is available on the machine.

WilliamDue:
Hello @athas can I ask if these are the only changes that were needed to be made?
 - Set a max time for all jobs. Currently it is set to 8 hours.
 - Use srun.
 - Remove the weird piping to sh.
 - Removal of gpu field by only using a slurm-options field.

athas:
Why did you change which commit is used for `futhark-benchmarks`?

WilliamDue:
> Why did you change which commit is used for `futhark-benchmarks`?

When I reverted I got an error doing the benchmarks and I was guessing I were using the wrong version of futhark-benchmarks.

This was the error [link](https://github.com/diku-dk/futhark/actions/runs/4215923948/jobs/7317474933).

athas:
Use the same version as on `master`.  No change should show up in the diff.

Issue #550, opened 2018-05-12
Warn on unused functions


Athas:
This is particularly tricky in the presence of modules, because we may not know that a function is truly unused until it is hidden by an outer ascription.  Maybe @melsman knows how ML compilers do this (if they do).  Otherwise, hey, research project.

razetime:
with some searching i found this one for OCaml: https://github.com/LexiFi/dead_code_analyzer

razetime:
will `futhark check-unused` be a good name for this warning?

athas:
That can always be bikeshedded later.  I actually envision this being part of a more general linter-style command that checks for all sorts of things.

athas:
Here are some notes on interesting cases.  Some involve multiple files, so pretend that the code blocks are actually multiple files, indicated by comments.

# Case 1

```
-- a.fut
def f x = x + 2
```

Is `f` unused?  It's not called from anywhere, but this file might be imported by something else.  Hence we should consider `f` used.

# Case 2

```
-- a.fut
def f x = x + 2

-- b.fut
import "a"
def g x = x + 2
```

If b.fut is the actual application or library "root file", then we should consider `g` used but `f` unused.

I think the basic principle of operation should be this: the tool is given one or more `.fut` files.  These we consider *root files*.  It then type-checks these files and all of the files they import, and report any function not reachable through a non-`local` top-level definition in a root file.  This means there is a risk of having unused top level functions in libraries, but hopefully these will be easy to spot, as they show up in generated documentation.

razetime:
some words from athas on modules:
The issue is that definitions change their names when module abstraction is used.  E.g:
```Futhark
module m : { val x : i32 } = { def x = 2 }
def main = m.x
```

The `VName` for the `def x` and the `m.x` will be different (specifically, the numeric tag will differ).

The module definition will contain a mapping between the "outer" name and the "inner" name.  You will have to make use of that mapping, somehow.

PR #1865, opened 2023-02-03
Warn on unused functions [WIP]

razetime:
Student project, issue #550.


athas:
Start by merging master into your branch.  You'll get a few conflicts and errors due to AST changes, but nothing major.

Then fix the various style issues and warnings exposed by CI.

After that I'll do a closer review of your code, but as I remember the fundamental structure is sound.

athas:
The merging is incomplete and the code still does not compile.

razetime:
yes, it will be rectified. I will post a reply once all the required changes are done.

razetime:
the code compiles now and i've used `ormolu -i` on the two new files. I will remove the `un` testing folder now.

athas:
I'll give the code a review a bit later, but here's a functionality request: `futhark unused` should only list unused definitions in the files actually passed on the command line.  Currently it will list e.g. most of the prelude as unused (unless you have a program that actually uses all of it).  Similarly, if you use a library, it'll list a lot of that library as unused.  That's not useful, as that is expected behaviour.

I'm also not a big fan of the output format.  It seems awfully redundant:

```
/prelude/soacs.fut: map1 -> /prelude/soacs.fut:58:1-59:10
```

A better format might be the following:

```
/prelude/soacs.fut:58: map1
```

This excludes the column number, as it's unlikely that a single line of code will define multiple functions of the same name on the same line.  Coloured output can be used to highlight the function name.  Maybe the function should come first - not sure.

razetime:
I have a doubt: https://github.com/diku-dk/futhark/issues/550#issuecomment-1406392165 for this example, if `b.fut` is given on the command line, we should only show results from `a.fut`, right?

I will change the output format right away.

athas:
No, if `b.fut` is given on the command line, you should only show results for `b.fut`.

athas:
Hm, wait, no, you're right, that would be less useful.  Maybe we need to think a little bit more about how to filter the results.

src/Futhark/CLI/Unused.hs
```diff
@@ -0,0 +1,45 @@
...
+              M.toList $
+                findUnused files imp
+
+data CheckConfig = CheckConfig Bool
+
+initialCheckConfig :: CheckConfig
+initialCheckConfig = CheckConfig False
+
+lineOnly :: [Char] -> [Char]
+lineOnly str = (elemIndices ':' str !! 1) `take` str
```
athas, line 45: Don't do this by making assumptions about what `locStr` output looks like and then splitting it apart.  Write a separate function for turning a `Loc` (or `SrcLoc`) into a string.

src/Futhark/CLI/Unused.hs
```diff
@@ -0,0 +1,45 @@
...
+printUnused :: [FilePath] -> IO ()
+printUnused files = do
+  res <- runFutharkM (readProgramFiles [] files) Verbose
+  case res of
+    Left err -> liftIO $ do
+      dumpError newFutharkConfig err
+      exitWith $ ExitFailure 2
+    Right (_, imp, _) -> do
+      putStrLn $
+        unlines $
```
athas, line 33: From what I remember, `unlines` will always add a final linebreak, and so will `putStrLn`.  So this will result in a blank line at the end.

src/Language/Futhark/Unused.hs
```diff
@@ -0,0 +1,207 @@
...
+-- ∘ Subtract the set produced in (3) from the set of all VNames. This gives you a set of VNames that are not reachable from the root files.
+
+-- Guidelines:
+-- (1) create a mapping for functions in *any* import to the directly usable functions in it.
+-- (2) create a mapping from normalized filename to top-level functions inside the filename.
+-- (3) Compute transitive closure for the map in (1)
+
+-- We are looking for VNames that represent functions.
+
+-- cabal run futhark -- unused un/b.fut
```
athas, line 25: No need for this comment.

src/Language/Futhark/Unused.hs
```diff
@@ -0,0 +1,207 @@
...
+
+getDecs :: FileModule -> [DecBase Info VName]
+getDecs (FileModule _ _env (Prog _doc decs) _) = decs
+
+findUnused :: [FilePath] -> [(ImportName, FileModule)] -> M.Map FilePath [(VName, SrcLoc)]
```
athas, line 30: You should add a comment for what this `[FilePath]` means.  If I recall correctly, it's the root reachable files or something like that.

src/Language/Futhark/Unused.hs
```diff
@@ -0,0 +1,207 @@
...
+      used1 = tClosure bf $ M.union bf $ M.unions $ map snd rf
+      used = S.unions $ map snd $ M.toList used1
+  M.fromList $
+    map
+      ( \(x, y) ->
+          ( x,
+            map (\a -> (a, locs M.! a)) $
+              filter (`M.member` locs) $
+                S.toList $
+                  S.fromList (map fst $ M.toList y) `S.difference` used
```
athas, line 46: This can be written without a round trip through Set.  Something like `filter (flip S.notMember used)`.

src/Language/Futhark/Unused.hs
```diff
@@ -0,0 +1,207 @@
...
+    If ex1 ex2 ex3 _ -> funcsInExp ex3 n $ funcsInExp ex2 n $ funcsInExp ex1 n vm
+    DoLoop _ _ ex1 loop_ex ex2 _ -> funcsInExp (fromLoop loop_ex) n $ funcsInExp ex2 n $ funcsInExp ex1 n vm
+    BinOp (QualName _ vn, _) _ (ex1, _) (ex2, _) _ -> funcsInExp ex2 n $ funcsInExp ex1 n $ M.adjust (S.insert vn) n vm
+    LetWith _ _ sl ex1 ex2 _ -> foldl' (\x y -> funcsInExp y n x) vm $ fromSlice sl <> [ex1, ex2]
+    Index ex sl _ -> foldl' (\x y -> funcsInExp y n x) vm $ ex : fromSlice sl
+    Match ex cases _ -> foldl' (\x y -> funcsInExp y n x) vm $ ex : map fromCase (NE.toList cases)
+funcsInExp _ _ vm = vm
+
+-- funcs inside a module expression. also tracks functions that have different VNames inside and outside the module.
+-- ModVar omitted
```
athas, line 115: Add a case for `ModVar`.  Wildcard cases are dangerous because they mean we might not notice this function if we ever add more constructors to `ModExp`.

src/Language/Futhark/Unused.hs
```diff
@@ -0,0 +1,207 @@
...
+    LetPat _ _ ex1 ex2 _ -> funcsInExp ex2 n (funcsInExp ex1 n vm)
+    LetFun vn (_, _, _, _, ex1) ex2 _ ->
+      funcsInExp ex2 n (funcsInExp ex1 vn (M.insert vn S.empty vm)) -- Important case! function defn
+    If ex1 ex2 ex3 _ -> funcsInExp ex3 n $ funcsInExp ex2 n $ funcsInExp ex1 n vm
+    DoLoop _ _ ex1 loop_ex ex2 _ -> funcsInExp (fromLoop loop_ex) n $ funcsInExp ex2 n $ funcsInExp ex1 n vm
+    BinOp (QualName _ vn, _) _ (ex1, _) (ex2, _) _ -> funcsInExp ex2 n $ funcsInExp ex1 n $ M.adjust (S.insert vn) n vm
+    LetWith _ _ sl ex1 ex2 _ -> foldl' (\x y -> funcsInExp y n x) vm $ fromSlice sl <> [ex1, ex2]
+    Index ex sl _ -> foldl' (\x y -> funcsInExp y n x) vm $ ex : fromSlice sl
+    Match ex cases _ -> foldl' (\x y -> funcsInExp y n x) vm $ ex : map fromCase (NE.toList cases)
+funcsInExp _ _ vm = vm
```
athas, line 112: Avoid wildcard patterns.  Match all expressions.

src/Language/Futhark/Unused.hs
```diff
@@ -0,0 +1,207 @@
...
+fromDimInd :: DimIndexBase f vn -> [ExpBase f vn]
+fromDimInd (DimFix x) = [x]
+fromDimInd (DimSlice m1 m2 m3) = catMaybes [m1, m2, m3]
+
+fromCase :: CaseBase f vn -> ExpBase f vn
+fromCase (CasePat _ x _) = x
+
+getFieldExp :: FieldBase Info VName -> ExpBase Info VName
+getFieldExp (RecordFieldExplicit _ ex _) = ex
+getFieldExp _ = error "placeholder" -- TODO: provide an appropriate error here.
```
athas, line 202: What's going on here?

src/Language/Futhark/Unused.hs
```diff
@@ -0,0 +1,207 @@
...
+-- possible future optimization:  remove VNames that aren't referenced in any top level import
+tClosure :: VMap -> VMap -> VMap
+tClosure bf af =
+  let bf' = tStep af bf
+   in if bf == bf'
+        then bf
+        else tClosure bf' af
+
+tStep :: VMap -> VMap -> VMap
+tStep af = M.map (\x -> S.union x $ S.unions $ S.foldl (<>) [] $ S.map (\y -> maybeToList $ y `M.lookup` af) x)
```
athas, line 64: This can be written more elegantly.  The last part certainly as something like:

    mapMaybe (`M.lookup` af)

athas (changes requested):
Apart from the code changes, this also need tests.  Take a look in `tests_adhoc` and see if you can come up with a way of testing it.

src/Language/Futhark/Unused.hs
razetime: removed the round trip and also removed an extra filtering pass in the process

src/Futhark/CLI/Unused.hs
razetime: used location data instead. my bad.

src/Language/Futhark/Unused.hs
razetime: added to this section. will change all the rest of the functions to do the same.

src/Language/Futhark/Unused.hs
razetime: it's used when reading record literals. to avoid code duplication. It's used on L78 and L139 current commit

athas: Using `error` is an absolute last resort.  Duplicate code is better.

razetime: I just realized it was an obvious use for `Maybe`. Fixed.

src/Language/Futhark/Unused.hs
razetime: i have rewritten this with mapMaybe.

athas:
Is anyone still interested in making this work?  Otherwise I'll close the PR for now.

razetime:
I will add some tests today.

razetime:
The tests get pretty big since the prelude is present in the output. how do we distinguish between prelude and other imports?

athas:
That's a good question.  I think that is actually an open question in the design that we never found a good answer for: how do we distinguish between unused functions that we care about, and unused functions that we do not?  The prelude is one case, but we also don't want to warn about unused functions in other third party code.  I don't offhand have a good answer.

razetime:
does it make sense to exclude imports that are not directly named in the file?

athas:
That would take care of the prelude, but not third party libraries.  It's a starting point, but it might be better to just exclude the prelude and then we can figure out the rest later.  The main problem is that this branch currently does not even compile.

razetime:
Closing this PR for now. will reopen/create a new issue when i make tests ready and have ideas on what output to show

PR #1873, opened 2023-02-11
Make consumption an effect on functions, rather than types.

athas:
This is a breaking change, because until now we allowed functions like

    def f (a: *[]i32, b: []i32) = ...

where we could then pass in a tuple where in an application `f (x,y)` the value `x` would be consumed, but not `y`.  However, this became increasingly difficult to support as the language grew (and frankly, it was always buggy).  With this commit, the syntax above is still permitted, but it is interpreted as

    def f ((a,b): *([]i32, []i32)) = ...

i.e. the single tuple argument is consumed *as a whole*.  Long term we can also consider amending the syntax or warning about cases where it is misleading, but that is less urgent.

I've wanted to make this simplification for a long time, but I always hit various snags.  Today I managed to make it work, and the next step will be cleaning up the notion of "uniqueness" in return types as well (it should be the more general notion of "aliases").


athas:
@coancea I think you use this style now and then, so you should note that it might stop working if/when this is merged.

Issue #1861, opened 2023-01-31
Improved plotting tools


athas:
I would like a program that produces plots of the data in the JSON files produced by `futhark bench`.  This program should probably be written in Python. (I have a preference for Gnuplot due to its simplicity, but I don't think young people nowadays like it).

For each benchmark/workload combination in a JSON file, plots of the following should be produced:

* A plot of per-run runtime, similar to the ones in Figure 3 in [this thesis](https://futhark-lang.org/student-projects/aleksander-bsc-thesis.pdf).
* A plot of the cumulative runtime, similar to Figure 2, right side.
* A plot of time densities, similar to Figure 2, left side.
* Maybe also lag plots, such as on Figure 8.

As a start, simply producing the plots as a bunch of images is enough.  However, after that it might be very nice to have it also produce an HTML page that embeds the plots as well as other interesting statistical information (like what Criterion can produce).

WilliamDue:
As far as I have understood these are the changes that are needed to be made.
- Display 95% confidence interval on the page.
- Display +/- % of mean on the page.
- Display memory usage on the page.
- Display the runtime density plot in a manner such that you can see the distribution clearly.
- Change the naming of iteration and index to runtime.

PR #1877, opened 2023-02-13
Plotting tool.

WilliamDue:
The script that generates the plots mentioned in issue #1861 there might still be some details missing.


tools/plot-bench.py
```diff
@@ -0,0 +1,305 @@
+import argparse
```
athas, line 1: Missing a shebang here.

tools/plot-bench.py
```diff
@@ -0,0 +1,305 @@
...
+
+    make_plot_jobs_and_directories(programs, plot_jobs, data, plots, filetype)
+
+    for plot, data in plot_jobs.items():
+        for data_chunk in chunks(data, max(len(data) // 4, 1)):
+            plotters[plot]().plot(data_chunk)
+
+
+if __name__ == '__main__':
+    main()
```
athas, line 305: Missing final newline.

WilliamDue:
To see usage look under tools where `plot-bench.py`  can be found. It can be used by `python3 plot-bench.py data.json` use the --help flag for further details.

tools/plot-bench.py
```diff
@@ -0,0 +1,305 @@
...
+        'text.usetex': False,
+        'axes.prop_cycle': matplotlib.cycler(color=['#5f021f'])
+    })
+
+    args = get_args()
+    data = json.load(open(args.filename, 'r'))
+    plots = format_arg_list(args.plots)
+    programs = format_arg_list(args.programs)
+    filetype = args.filetype
+    quality = args.quality
```
athas, line 283: Realistically, nobody is going to set this.  It is *your* responsibility as the programmer to pick what is right, rather than pushing it on to your users.  Since graphs compress losslessly very well as PNG files, and the sizes don't matter, I would prioritize visual quality.

tools/plot-bench.py
```diff
@@ -0,0 +1,305 @@
...
+        plots = set(plot_jobs.keys())
+
+    if programs is None:
+        programs = set(data.keys())
+
+    make_plot_jobs_and_directories(programs, plot_jobs, data, plots, filetype)
+
+    for plot, data in plot_jobs.items():
+        for data_chunk in chunks(data, max(len(data) // 4, 1)):
+            plotters[plot]().plot(data_chunk)
```
athas, line 301: I would like the program to print the name (to stdout) of every plot file that it creates.

athas (changes requested):
This is a good start.  I think the next step is to also generate an HTML file that shows all the images.

athas:
Structure of the (single) HTML file I want to see generated:

* A `<nav>` with an `<ul>` containing an `<li>` for every benchmark, with a link to the `<section>` in the file with that benchmark.
* For each benchmark, a `<section>` with ``s pertinent graphs in some readable size (possibly these can be clicked to open a larger graph file, but I think all of these are perfectly legible at the default size).  Over time, this section can also be extended to contain more information (averages, standard deviations, min/max runtimes, etc).
* Embedded CSS in `<head>` (no separate `style.css`  or whatnot).  I want it to be extremely easy to copy the result around.

tools/plot-bench.py
```diff
@@ -0,0 +1,305 @@
...
+    filetype: str
+    ) -> None:
+    """Makes dictionary with plot jobs where plot_jobs are the jobs."""
+
+    for program in programs:
+        datasets = data.get(program).get('datasets')
+        for dataset_path, dataset in datasets.items():
+            runtimes = dataset.get('runtimes')
+            directory = os.path.dirname(program)
+            program_basename = pathlib.Path(program).stem
```
athas, line 237: This is wrong, as it cuts off the entry point name in things like `"futhark-benchmarks/babelstream/babelstream.fut:f32_add"`.

athas (changes requested):

tools/plot-bench.py
WilliamDue: I have removed quality now and now it just uses png dpi=200

tools/plot-bench.py
WilliamDue: This should be correct now but resulted in more rendering time.

WilliamDue:
I believe I have fixed the errors mentioned but I do not know if the html file is ideal.

athas:
This looks really nice!  I am very good at CSS so I will see if I can prettify it a bit.

athas:
More requests: I want the initial list to have a tree structure of the form

* benchmark1
  * dataset1
  * dataset2
  * ...
* benchmark2
* ...

I also want the sections to be grouped similarly, with headlines (`<h2>`) for benchmarks, and `<h3>`s for datasets.

(I might decide to merge this PR and turn this into a separate issue for further work.)

athas:
Also retain the original directory structure when producing the list of benchmarks.

Issue #1889, opened 2023-03-07
plot-bench.py: general readability improvements


athas:
The initial list should be structured not just by benchmark and dataset, but also group benchmarks into sublists based on their containing directory (e.g. everything in `futhark-benchmarks/micro/` should be indenteded one more level).

The dataset subsection headers should also contain the name of the benchmark somewhere, e.g. before or after `large.in` here:

![image](https://user-images.githubusercontent.com/55833/223435910-9cc719d7-8cb0-4a0d-a3d8-db4ec668835b.png)

This is useful for benchmarks that have many datasets and you want to link directly to them.

All headers should be clickable and link to themselves.

The ancillary data needs to be formatted better:

![image](https://user-images.githubusercontent.com/55833/223436203-2b73a52a-d8c0-4a54-a7d1-211c012ab39f.png)

* We do not need 20 significant digits of precision for the mean error.
* The memory usage should be rounded to an appropriate unit (e.g. GiB) - with the exact number perhaps pushed off to the side.
* The 95% confidence interval should also be rounded to an appropriate unit (microseconds, miliseconds, or seconds as appropriate).

Issue #1888, opened 2023-03-07
plot-bench.py: the ancillary data subsection is difficult to read


athas:
Instead of using a table or list, it's a `<div>` with some CSS set that makes it slightly table-like (?).  It does not work for me, and it is difficult to read because the numbers are not horizontally aligned:

![image](https://user-images.githubusercontent.com/55833/223435164-62982101-9f1c-4dd4-9e49-398727937a11.png)

Surely this can be made more readable.  Also, the Python code should be structured such that it is easy to add new data points in the future (currently it's a hardcoded HTML fragment).

Issue #1887, opened 2023-03-07
plot-bench.py: put graph files in a subdirectory


athas:
Currently the graph files are put in a subdirectory corresponding to the directory structure of the original benchmarks (e.g. `futhark-benchmarks`).  However, I would prefer if this was wrapped in a subdirectory named after the JSON file.  E.g. for `futhark-cuda-A100-c787f63f37c77e986ac53269b84e134648fcf111.json` the files would be in `futhark-cuda-A100-c787f63f37c77e986ac53269b84e134648fcf111-plots/futhark-benchmarks/...`.

Incidentally, instead of `index.html` perhaps we should use the name of the JSON file with `.json` replaced with `.html`.  That, along with the above, would make it possible to plot multiple JSON files in the same directory, which I expect to become a frequent use case.

Issue #1886, opened 2023-03-07
plot-bench.py: histogram improvements


athas:
![image](https://user-images.githubusercontent.com/55833/223433392-ba28b03a-8ccf-49eb-a76a-09689f7b746f.png)

* I think the y axis label should be "density", not "probability".

* The x axis having a lower bound of 0 is not good, as seen here:

![image](https://user-images.githubusercontent.com/55833/223433723-df3e4965-22c6-4546-abff-18b3d6ebf787.png)

It is completely impossible to see what the actual distribution is.

FluxusMagna:
I think labeling/scaling the y-axis is a bit problematic since the integral of the function should be equal to 0. Using the time scale it is of course possible to set it by some numerical integration of the data, but I'm not sure of how great value the scale would be.

In some ways a cumulative distribution function  would make more sense to plot, since it always goes between 0 and 1.

athas:
I don't understand why re-labelling the y-axis would be a problem (and the scale is fine).

A cumulative distribution graph might be useful, though.

FluxusMagna:
Since it's the integrand, the unit would be 1/s, and the amplitude would depend hugely on the overall runtime variance, without any consideration of the relative deviation. It would not be completely useless, but for comparison between processes of different average runtime, the number itself will not be very useful for comparing the two. Perhaps I'm just not seeing the actual purpose of this type of plot. 

Say for example that we have the same process running on different hardware. The runtime distribution looks the same, except stretched out for the slower case. In this type of plot the scaling on both the x-axis and y-axis need to change, but in a cumulative plot, only x-axis scaling would change, making comparison easier.

Another reason to prefer a cumulative plot is that we are pretty poor judges of the area under a spike in a plot like this(which is what the quote is about).  Spikes that are visually similar can differ a fair bit in terms of the number of values they represent. In this sense a bar-plot would be a better way to represent the bins.

athas:
The purpose of these plots is mostly to see whether something fishy is going on, e.g. two peaks where the average is in the middle, which probably suggests that some bistable phenomena like frequency stepping or overheating cycles is happening.  It is mostly for internal use by compiler hackers when investigating performance on new hardware.

FluxusMagna:
I guess if it's for purely qualitative observation, the scale does indeed not matter. I've previously worked with data where this type of plotting can obscure important quantitative information, but that's probably not the case here then.

athas:
[...]

PR #1893, opened 2023-03-13
A possible solution to issue #1886, #1887, #1888 and #1889.

WilliamDue:
The following pull request contains.
 - The y-axis on the runtime density plot is renamed to "Density".
 - Leading and trailing zeroes are removed from the runtime density plots.
 - The generated directory is named based on the given json file. If the file given is `name.json` then the resulting folder is `name-plot`. The html file is then given the name `name.html`
 - A table is used now instead of the weird div I used before.
 - The code has a dictionary called `DESCRIPTORS`  with functions which allows for new datapoints to be added to the table.
 - The list of benchmarks and their datasets are now in a tree structure.
 - The name of the benchmark is now added before the dataset.
 - All headers link to themselves.
 - The ancillary data is now rounded to two significant digits (I have no clue if this is agood choice?).
 - Memory usage uses GiB as the unit now.
 - All time units are chosen based on if the unit is the first unit that can express the quantity with an integer part. The possible units are hours, minutes, seconds, microseconds and nanoseconds.
 - Did also some refactoring to the html generation and also to the plotting such that only one plot file is made with subplots making them better aligned.


athas (changes requested):
Mostly looks pretty good!

Don't use scientific notation for memory usage; use decimals.  "1.68e-03GiB@device" is a weird way of spelling `1.68MiB`.

The four plots are now one big image.  Where they always like that?  I prefer an image per plot (it's easier to take apart when you only want one of them).

WilliamDue:
Yes I changed it to one image and now I have changed it back. I Also use some more sensible units for memory usage now.

Issue #1659, opened 2022-04-29
Allow arbitrary expressions in sizes


athas:
I had a conversation with Martin today where we concluded that it may be possible to allow arbitrary expressions to be used as sizes as long as

* they have type `i64`.
* they consume nothing.

This would allow us to give types to functions such as `concat`:

```Futhark
val concat [n] [m] 't : [n]t -> [m]t -> [n+m]t
```

We have discussed this many times, but were reluctant to add it because in general it needs us to compute the inverse of arbitrary functions:

```Futhark
def length [n] (x: [f n]t) = n
```

Given some `x`, how do we find `n`?  We say that an array `x` *witnesses* `n` if we can extract the value of `n` from the run-time shape of `x`.  The trick we believe may work is to require that whenever you bind a size parameter, *you must have at least one array using that as the size of a dimension*.  This array will then *witness* the actual size, but you can still use it in "non-witness" ways in other arrays.  One motivating example is triangular matrices:

```Futhark
type triangular [n] = {data: [n*n/2]t, size: [0][n]()}
```

Note the `size` field above - this is required already, for the same reason.  We just hadn't realised that this allows us to safely make the type system even more expressive.

## What this does not handle

This does not address the question of how we can make equality of expressions more flexible.  We will not be able to infer that `x++y` and `y++x` have the same shape.  This is left for future work; perhaps infinitely far in the future.

catvayor:
Why expect i64 instead of u64 as sizes should be non-negative ?

athas:
Unsigned integers should not be used to encode "this value cannot be negative", but only as bit vectors.

PR #1897, opened 2023-03-17
Arbitrary size expressions

catvayor:
Implementation of #1659 (internalization broke)


catvayor:
when removing constraints on unflatten, only 6 errors in full test:

1.  tests/types/badsquare-lam.fut
typing fails for a bad reason: 
> Function body does not have expected type.
  Expected: i64 -> ?[d₀].[d₀][d₀]i64
  Actual:   (n: i64) -> *[(n + 1)][(n + 2)]i64
  Cannot unify size variable "(n + 1)" with "d₀" (scope violation).
  This is because "n" is rigidly bound in a deeper scope.
This notably implies that a code such as
```futhark
type square [n] 't = [n][n]t
def ext_square : i64 -> square [] i64 = \n -> tabulate_2d (n+1) (n+1) (\i j -> i + j)
```
doesn't type for the same reason.

2. tests/shapes/funshape6.fut
should not type by causality checking, but now pass.
To be honest, I don't understand why it should fail.

3. tests/shapes/funshape3.fut
Same thing.

4. tests/shapes/funshape5.fut
Should fail by causality checks, this time it fails because the entry point is polymorphic.
Like the two points before about the expected error.

5. tests/shapes/error5.fut
Not matching types error was expected, now it's a message I added
> Cannot apply "ap" to "(\(xs: [n₂]t₄) ->
                          xs ++ xs)" (invalid type).
  Expected: a₀ -> a₀
  Actual:   (xs: [n₂]t₄) -> *[n₂ + n₂]t₄
  Cannot unify size variable "n₂ + n₂" with "n₂"
  This is because "n₂ + n₂" depend of "n₂".

I think it's correct, but the way I added the error may be the cause of fail 1.

6. tests/issue1435.fut
should not type with error 'artificial size', now pass
I don't know what it means.

prelude/array.fut
```diff
@@ -62,10 +62,10 @@ def reverse [n] 't (x: [n]t): [n]t = x[::-1]
 -- **Work:** O(n).
 --
 -- **Span:** O(1).
-def (++) [n] [m] 't (xs: [n]t) (ys: [m]t): *[]t = intrinsics.concat xs ys
+def (++) [n] [m] 't (xs: [n]t) (ys: [m]t): *[n+m]t = intrinsics.concat xs ys :> [n+m]t
 
 -- | An old-fashioned way of saying `++`.
-def concat [n] [m] 't (xs: [n]t) (ys: [m]t): *[]t = xs ++ ys
+def concat [n] [m] 't (xs: [n]t) (ys: [m]t): *[n+m]t = xs ++ ys :> [n+m]t
```
catvayor, line 68: This constrain should be useless

athas:
> Cannot unify size variable "n₂ + n₂" with "n₂"
> This is because "n₂ + n₂" depend of "n₂".

This is the [occurs check](https://en.wikipedia.org/wiki/Occurs_check), isn't it?

athas:
> tests/issue1435.fut

This is fine; it should type-check now.

catvayor:
> > Cannot unify size variable "n₂ + n₂" with "n₂"
> > This is because "n₂ + n₂" depend of "n₂".
> 
> This is the [occurs check](https://en.wikipedia.org/wiki/Occurs_check), isn't it?

I had to add a test in `linkVarToDim` else the compiler had problems. I think this check was not done on size variable as they were trivial before

athas:
Right, that makes sense.  But the error message should clearly contain "occurs check" somewhere.  It's a standard term of art.

melsman:
... Or a solution could be to unify n₂ with the 0-size...

catvayor:
The error on `tests/types/badsquare-lam.fut` doesn't seem to come from my modifications so I don't know if it's to be corrected. This behavior doesn't seem normal...

```
$ cat iota.fut
def ext_square : i64 -> []i64 = iota
$ futhark check iota.fut
$ cat lam-iota.fut
def ext_square : i64 -> []i64 = \n -> iota n
$ futhark check lam-iota.fut
Error at lam-iota.fut:1:33-44:
Function body does not have expected type.
Expected: i64 -> ?[d₀].[d₀]i64
Actual:   (n: i64) -> *[n]i64

Cannot unify size variable "n" with "d₀" (scope violation).
This is because "n" is rigidly bound in a deeper scope.
When matching return types.
If you find this error message confusing, uninformative, or wrong, please open an issue:
  https://github.com/diku-dk/futhark/issues
$ futhark -V
Futhark 0.24.1
git: HEAD @ d49961c (Sun Mar 12 10:58:33 2023 +0100)
```

athas:
Here is another interesting case:

```Futhark
iota (let x = y in x)
iota (let x = y in x)
```

These expressions will currently be considered to have different types, because the `x`s are not the same, yet they are obviously alpha-equivalent.  I don't know whether we care about this quirk.  Implementing alpha-equivalence checking is probably not difficult.

catvayor:
`test -C tests` now pass all tests, full test still encounter errors, mainly errors like
```
./tests/distribution/distribution3: When loading tuning from [...]: Unknown name 'default_threshold' on line 1.
```
some tests says `'main.suff_outer_par_0'` to be unknown, and finally `reshape1` don't produce the expected error message, but it looks right

athas:
Pass `--no-tuning` to `futhark test` when not using a GPU backend.

athas:
The next step, type-checking wise, is resolving problems with unification of size expressions.  For example, this does not type check:

```Futhark
def f [n][m] (_: [n+m]i32 -> bool) = true

def g [n][m] (_: [n+m]i32) = true

def main = f g
```

(Actually, we might not want `f`to type check for other reasons, but here it is `f g` that fails.)

This also shows up during module ascription, which is where I first noticed it:

```Futhark
module meta: {
  val plus_comm [a][b]'t : [a+b]t -> [b+a]t
} = {
  def plus_comm [a][b]'t (xs: [a+b]t): [b+a]t = xs :> [b+a]t
}
```

We probably need to do unification of expressions, rather than just straight-up equality.

athas:
Remarkably, compilation now works.  This must mean we have insufficient tests.

The failure is because we don't allow those fancy identifiers in the core language.  That can be fixed either by not producing them, or by modifying the core language parser to allow them.  There is no grand vision for the core language syntax, so do whatever appears convenient.

catvayor:
Current unknown size in expression elimination is maybe a bit to hard currently, for exemple
```futhark
map (filter (>0) xs ++) mat
```
is refused because `(filter (>0) xs ++)` has type `[n]t -> ?[d]. [d]t`.
it may be solved by doing this elimination latter, let's say just when we affect or return it, but it would be at the risk that those non-standard sizes appear in latter stage of compilation.

athas:
I'd be OK with not allowing that form for now, out of caution.  However, I am mystified that this is also rejected:

```Futhark
  let xs' = filter (>0) xs
  in map (xs' ++) mat
```

Interestingly, it is allowed if we explicitly bind the size of the filter:

```Futhark
  let [n] (xs': [n]i32) = filter (>0) xs
  in map (xs' ++) mat
```

catvayor:
I need to check this, but I think it's because if we don't bind explicitly the size of the filter it remains as an `UnknowableSize` which are freed

catvayor:
It is, my problem is more how to manage this in a match for example, because the same problem will arise, and I do not remember a way of bounding sizes in a match case, this has the advantage of being able to cover all cases where the problem can arise

athas:
You cannot bind sizes in a match case, but why would you need to?  Sizes can only be introduced by the value being matched on, which could be `let`-bound beforehand.

catvayor:
For something like
```futhark
match filter ...
  case xs -> body
```
This is obviously stupid, but something equivalent could appear for function returning sum types with existential sizes, then we could have the same strange problem of refusing `map (++ xs)` in `body`

athas:
Could that not be solved by normalising to the following?

```
let tmp = filter ...
match tmp
  case xs -> body
```

catvayor:
I guess I could do this yes, I also need to check for loops, but they seems pretty restrictive for this so should be fine

catvayor:
```futhark
def f (p: {a:i64,b:bool}) : [p.a]i64 = iota p.a
```
fails with `Unknown name "p.a"` at typing, targeting the size expression. Probably scope of return type not well calculated.

src/Language/Futhark/TypeChecker/Types.hs
catvayor, line 1: this make make `checkExpForSize = checkSizeExpM`, probably need to rename one of them to the other

src/Futhark/CLI/Dev.hs
```diff
@@ -74,11 +75,13 @@ data FutharkPipeline
     Pipeline [UntypedPass]
   | -- | Partially evaluate away the module language.
     Defunctorise
-  | -- | Defunctorise and monomorphise.
+  | -- | Defunctorise, monomorphise.
```
catvayor, line 78: 'and' was good

athas:
We need to consider the interaction with uniqueness types.  [Here's a primer for the uninitiated.](https://futhark-lang.org/blog/2022-06-13-uniqueness-types.html)

In the following examples I will use this contrived function that consumes its argument:

```Futhark
def consume (xs: *[]i64): i64 = xs[0]
```

The underlying problem is that an expression `consume A` must only be evaluated *at most once*, as *A* becomes inaccessible after being consumed.  I don't think size types guarantee that.  For example, this should not work:

```Futhark
def f [n] (ns: [n]i64) (xs: [consume ns]f32) = xs[0]
```

...and it doesn't, but the error message makes me feel that's by accident.  The error message should be something like "`ns` cannot be consumed here" or "free variables cannot be consumed in size expressions".  We already have some of the machinery for doing that (e.g. because we also don't want lambdas to consume free variables).

This one is harmless:

```Futhark
def f (xs: [consume (iota 10)]f32) = xs[0]
```

Although consumption is happening, it's not of a free variable.

A tricky situation occurs here:

```Futhark
def f [n] (ns: *[n]i64) = iota (consume ns)
```

Due to the substitution of `consume ns` into the return type of f, the function ends up having the return type `[(consume ns)]i64`, which seems dangerous, as this expression cannot possibly be evaluated again (`ns` has already been consumed).  Here are some thoughts on resolving it:

1. After inferring the type of a function, check whether any "forbidden" sizes have snuck in (meaning ones that consume).
2. After inferring the type of *anything*, check for forbidden sizes.  This means that `iota (consume ns)` *will be illegal* anywhere.  This is perhaps too restrictive.
3. Replace such forbidden sizes with unknown sizes, just like we used to do with e.g. `iota (x+y)`.  Probably issue a warning when this happens.

More generally, we also need to set down some rules for *how often* the expressions in a type can be evaluated.  I had an interesting talk with @melsman about this last week.  *At most once* is perhaps not realistic.  A constant number of times is fine, but if we e.g. end up substituting an expression into a loop, then that might change the asymptotics of the program and that is not fine.

src/Futhark/Internalise/Exps.hs
```diff
@@ -343,7 +343,7 @@ internaliseAppExp desc (E.AppRes et ext) e@E.Apply {} =
       -- application.  One caveat is that we need to replace any
       -- existential sizes, too (with zeroes, because they don't
       -- matter).
-      let subst = zip ext $ repeat $ E.SizeSubst $ E.ConstSize 0
+      let subst = zip ext $ repeat $ E.ExpSubst $ E.IntLit 0 (E.Info <$> E.Scalar $ E.Prim $ E.Signed E.Int64) mempty
```
athas, line 346: Use `sizeFromInteger`.

src/Futhark/Internalise/Monomorphise.hs
```diff
@@ -243,20 +243,62 @@ transformFName loc fname t
           )
           size_args
 
+transformTypeSizes :: TypeBase Size as -> MonoM (TypeBase Size as)
+transformTypeSizes = bitraverse onDim pure
+  where
+    onDim (SizeExpr e) = SizeExpr <$> transformExp e
+    onDim (AnySize v) = pure $ AnySize v
+
+transformTypeExp :: TypeExp Info VName -> MonoM (TypeExp Info VName)
```
athas, line 252: I am sad that we need to do this.  TypeExps are supposed to be a non-elaborated syntactical construct that we should never need to transform.  I think the only reason we keep them around is to produce error messages for size coercions that refer to the original (non-elaborated) type entered by the programmer.  However, fixing that is a matter for a different PR.

src/Futhark/Internalise/NormaliseSize.hs
```diff
@@ -0,0 +1,536 @@
...
+    )
+
+runNormaliser :: NormaliseM a -> VNameSource -> (a, VNameSource)
+runNormaliser (NormaliseM m) =
+  runState (runReaderT m mempty)
+
+addBind :: VName -> NormaliseM a -> NormaliseM a
+addBind = local . S.insert
+
+newtype InnerNormaliseM a
```
athas, line 45: Could you put a comment indicating what this is used for?

src/Futhark/Internalise/NormaliseSize.hs
```diff
@@ -0,0 +1,536 @@
+-- | This Normalisation module converts a well-typed, polymorphic,
+-- module-free Futhark program into an equivalent program without
+-- arbitrary size expression.
```
athas, line 3: Please put a comment indicating the overall strategy and important terms (e.g. "intros").

src/Language/Futhark/Syntax.hs
```diff
@@ -219,18 +224,56 @@ data AttrInfo vn
...
+    SizeExpr (ExpBase Info VName)
   | -- | No known size.  If @Nothing@, then this is a name distinct
     -- from any other.  The type checker should _never_ produce these
     -- - they are a (hopefully temporary) thing introduced by
     -- defunctorisation and monomorphisation.
     AnySize (Maybe VName)
-  deriving (Eq, Ord, Show)
+  deriving (Show)
+
+expSizeEq :: ExpBase NoInfo VName -> ExpBase NoInfo VName -> Bool
```
athas, line 238: Do we still need this?

athas (commented):
I'll need to spend a lot more time reading this, but here are a few remarks to start.

catvayor:
I think the best to do is taking the old behavior for consuming size expressions because if doing so, there are only 2 way of making such an expression appear:
- call a function that will just substitute one of its arguments as size -> replaced by an existential
- coercion -> there is no reason to do this as the size could not appear somewhere else, so forbid it should not be harmful

We will still need to check for non-consuming size where they could be obtain through `[] : [0][consume A]t`, so at least functions and pattern

src/Language/Futhark/Syntax.hs
catvayor: This are not only used in unification but also in normalization and unscopeType, they currently hold a `M.map Size VName`, but they probably could use bare cleaned exp

catvayor:
Warning message could be better, and it is probably also the case for error message, but consuming sizes are forbidden now, if my predicate is right, I used one that seems to have the right behavior seen somewhere else in the code, but I'm not completely sure it is right.

athas:
I'm investigating (and fixing) various size-expression-related things related to entry points.  One annoying part is that I don't see how we can make this work:

```Futhark
entry main (x: i64) (_: [x+1]i32) = x
```

Or rather, it works but the constraint is lost, so you can pass it whatever you want without complaints.  Fixing this will take some clever thinking.

src/Language/Futhark/Syntax.hs
```diff
@@ -1317,3 +1340,269 @@ instance Pretty BinOp where
...
+      TypeArgExpSize $ bareCleanSizeExp size
+    bareCleanTypeArgExp (TypeArgExpType tya) =
+      TypeArgExpType $ bareCleanTypeExp tya
+bareCleanTypeExp (TEArrow arg tya tyr loc) = TEArrow arg (bareCleanTypeExp tya) (bareCleanTypeExp tyr) loc
+bareCleanTypeExp (TESum cs loc) = TESum (map (second $ map bareCleanTypeExp) cs) loc
+bareCleanTypeExp (TEDim names ty loc) = TEDim names (bareCleanTypeExp ty) loc
+
+-- | Remove all annotations, parentesis, assertion and attribute from an expression, but retain the
+-- name/scope information.
+bareCleanExp :: ExpBase Info VName -> ExpBase NoInfo VName
```
athas, line 1530: Is this only used in the normalisation pass?  If so, it should be defined there.

src/Language/Futhark/Syntax.hs
```diff
@@ -1317,3 +1340,269 @@ instance Pretty BinOp where
...
+      TypeArgExpSize $ bareSizeExp size
+    bareTypeArgExp (TypeArgExpType tya) =
+      TypeArgExpType $ bareTypeExp tya
+bareTypeExp (TEArrow arg tya tyr loc) = TEArrow arg (bareTypeExp tya) (bareTypeExp tyr) loc
+bareTypeExp (TESum cs loc) = TESum (map (second $ map bareTypeExp) cs) loc
+bareTypeExp (TEDim names ty loc) = TEDim names (bareTypeExp ty) loc
+
+-- | Remove all annotations from an expression, but retain the
+-- name/scope information.
+bareExp :: ExpBase Info VName -> ExpBase NoInfo VName
```
athas, line 1397: Put this back in Traversals.

src/Language/Futhark/Syntax.hs
catvayor: It is currently used for SizeExpr comparison

athas: I would very much like to get rid of those instances (or rather, return them to the default).  If we need a bespoke way of comparing size expressions somewhere, then we should define an appropriate function rather than repurposing `==`.  Defining an `Eq` or `Ord` instance that is not straight structural equality is almost always a bad idea.

I just tried removing those instances entirely and the type checker at least still seems to work.

catvayor: It does work without them, but I think we need to check the normalised code to be sure that the same size is calculated twice, one as `n+1`, and as second time as `(n+1)` or with some other parentheses

athas: That would only be a constant number of re-evaluations, so no great tragedy.  And in any case - that equality definition should if necessary be part of the normalisation module.

src/Futhark/Internalise/NormaliseSize.hs
```diff
@@ -0,0 +1,570 @@
...
+instance Pretty ReplacedExp where
+  pretty (ReplacedExp e) = pretty e
+
+instance Eq ReplacedExp where
+  ReplacedExp e1 == ReplacedExp e2
+    | Just es <- similarExps e1 e2 =
+        all (uncurry (==) . bimap ReplacedExp ReplacedExp) es
+  _ == _ = False
+
+instance Ord ReplacedExp where
```
athas, line 41: I am somewhat uncomfortable with this instance, since it's not obvious to me that it is actually correct (e.g. transitive), which may break at some point.  Could we replace the `M.Map ReplacedExp VName` values with `[(ReplacedExp,VName)]` and use `Eq`-based list operations instead?  That of course involves far more linear searches, but I suspect these mappings are usually small, so there might not be a performance loss.

athas:
How does Idris handle the equivalent of `iota <| 10`?

catvayor:
With some difficulties for the type of `<|`, we have to handle explicitly the type constructor
```idris
foo : {b: a -> Type} -> (x:a) -> ((y:a) -> b y) -> b x
foo x f = f x
```

athas:
I don't really expect we'll be able to handle it in Futhark any better than we do now.  Doing so requires more complexity than I think we can justify.  I'm just happy it doesn't crash the type checker!

catvayor:
I think we should limit the size expression assertions on entry points, typing guaranty that arguments given by futhark function will validate the assert, so doing it in every function should not be required

athas:
That was my intention, but I forgot to do so.

catvayor:
It's the current behavior,

```futhark
def main n =
  loop t = 0 for _ in [0,1,2] do
    let xs = iota n ++ iota n
    in t + reduce (+) 0 xs
```
is currently normalized as
```futhark
entry main (n: i64) : i64 =
  let (d<{(+) n n}>: i64) =
    (+) n n
  in loop (t: i64) = 0 for (_: i32) in [0, 1, 2] do
       let (xs: [d<{(+) n n}>]i64) =
         (++) (iota n) (iota n)
       in (+) t (reduce (+) 0 xs)
```
which calculate the size `d<{(+) n n}>` only once inside the main function. However, `++` will also calculate it in its body, which occurs at every iteration

athas:
That part is OK.  But I still see unfortunate behaviour for a contrived program such as this:

```Futhark
def contrived [n] (i: i64) (_: [n]i64 -> [n](i64,i64)) = i + n

def main (xs: []i64) =
  let ys = iota (i64.sum xs)
  in loop acc = 0 for i < 10 do
       acc + contrived i (\p -> zip ys p)
```

That is monomorphise to this:

```
def contrived (n: i64) ((i: i64)) ((_: [n]i64 -> [n](i64, i64))) : i64 =
  (+) i n

def zip [n] ((as: [n]i64)) ((bs: [n]i64)) : *[n](i64, i64) =
  zip as bs

entry main [d₀] ((xs: [d₀]i64)) : i64 =
  let (ys: [(sum xs)]i64) =
    iota (sum xs)
  in loop (acc: i64) = 0 for i < 10 do
       (+) acc (contrived (sum xs) i (\(p: [(sum xs)]i64) ->
                                       zip ys p))
```

Note the `sum xs` inside the loop body.  This is unfortunate.  I wonder how we can fix that.

catvayor:
this could not happen for something like 
```futhark
def contrived [n] (i: i64) (_: [2*n]i64) = i + n

def main (n: i64) =
  let ys = iota (2*(n+1))
  in loop acc = 0 for i < 10 do
       acc + contrived i ys
```
monomorphised as
```futhark
def contrived (n: i64) ((i: i64)) ((_: [(*) 2 n]i64)) : i64 =
  (+) i n

entry main ((n: i64)) : i64 =
  let (ys: [((*) 2 ((+) n 1))]i64) =
    iota ((*) 2 ((+) n 1))
  in loop (acc: i64) = 0 for i < 10 do
       (+) acc (contrived ((+) n 1) i ys)
```
however

athas:
Right, it's somewhat exotic.  It only happens for functions that use sizes non-constructively, where monomorphisation inserts explicit arguments.  I agree that it can be solved by storing the result of the outermost instance of the size expression - but then normalisation has to happen before monomorphisation!  We keep moving them around!

catvayor:
I think I have a plan to merge them, I'll work in another branch

catvayor:
The size expression replacement was done to avoid complexity impact of size expressions, but it also avoids post-comsumption observations in programs involving 
```futhark
let t = iota (sum xs)
let n = consume xs
let t' = t ++ iota n
```
Which is currently normalized like
```futhark
let d<{sum xs}> = sum xs
let t = iota d<{sum xs}>
let n = consume xs
let d<{sum xs + n}> = d<{sum xs}> + n
let t' = t ++ iota n
```
Which would be invalid without this replacement in expressions. I feel the need to somehow prove that things will always go well regarding post-consumption observations to be sure we won't have problems.

catvayor:
If a size involving xs arise after a consumption of xs, then it could not be from an argument to a function involving xs, else the typing would have refused it.
So it's the result of a polymorphic function which have received an array with a size involving xs, the new size will then just make this size appear in its expression, removing occurrences of xs in it.
By repeating this, the sub-expressions with xs must have been calculated as sizes before the consumption of xs, and so the normalization won't introduce post-consuming observation of xs.

catvayor:
It's possible to make it work I think, but it would require tricky work in monomorphisation to not move the consumption before other size evaluations

catvayor:
It would notably require to store in sizes their diet, because it looks non-trivial to determinate this after the typing

also, I think we may have problems with `(n:i64) -> ([n+1]t, [n+2]t)`

athas:
I don't think the ordering should be problematic, as you could follow the same evaluation order as the type checker does.

What would be the problem with a function `f : (n:i64) -> ([n+1]t, [n+2]t)`?  You mean if you do things like `f (consume X)`?  Then to me that should normalise to

```Futhark
let tmp = consume X
let (r0: [tmp+1]t, r1: [tmp+2]t) = f tmp
```

...but I'm guessing we don't do this kind of full expression normalisation yet?

We can leave this for later (after merging the PR and working some more on the formalisation).  This interaction with uniqueness types is fiddly and of course it's important that the system is sound, but nobody is going to actually code like this, so it's not very important that it is very smooth.

catvayor:
> I don't think the ordering should be problematic, as you could follow the same evaluation order as the type checker does.

I guess so, but currently evaluations just appear as soon as possible, in the optimal order regarding sub-expression, which can't work with a consumption at play. Determinate if we need to place it at some place is complicated.

> What would be the problem with a function `f : (n:i64) -> ([n+1]t, [n+2]t)`? You mean if you do things like `f (consume X)`?

Currently the normalization only operate on sizes, and so is not able to detect case like this one, so the consumption would occurs twice as neither `n+1` nor `n+2` are sub-expressions of the other.

athas:
It's pretty cool that this works and has the expected type:

```
def f (r: {x:i64}) = iota r.x
```

I know of people who'll be happy with that.  Of course, we should still (*later, in another PR*) simplify applications such as `f {x=10}`.

athas:
What are we missing before we can merge this?  I'm going through it all now, and it seems pretty solid.  Have I forgotten something we have yet to do?

catvayor:
> Also, currently the expressions are evaluated as soon as possible, which can be sad for
> 
> ```futhark
>   if cond
>   then -- (body with complicated sizes, but fully calculable before the if)
>   else -- (simple body)
> ```
> 
> Or problematic with assertions as the expression could be evaluated before an assert even if it should be after, or worse, move an assert in the precedent example.

This questions maybe ? And the 'Causality checks' ?
Apart of these, I think it's all good

athas:
I don't particularly care about that.  I am worried that it will cause confusing errors, though.  Do we still have the machinery that made the type of `iota (consume X)` an unknown size variable instead of `consume x`?  If so, we could use it for size expressions that have internal bindings (and ban them in type expressions).

We can describe this as a "temporary" limitation, but I expect it'll be a while before it is lifted, as it's not a terribly useful programming pattern.

Alternatively we can just handle alpha-equivalence in expression unification.  That would be more elegant, but I need to check how much code it would need.

catvayor:
> I don't particularly care about that. I am worried that it will cause confusing errors, though. Do we still have the machinery that made the type of `iota (consume X)` an unknown size variable instead of `consume x`?

Yes, if it happens it produce a warning, but it's workaround at multiple places.

> If so, we could use it for size expressions that have internal bindings (and ban them in type expressions).
> We can describe this as a "temporary" limitation, but I expect it'll be a while before it is lifted, as it's not a terribly useful programming pattern.

I don't think it's needed, in the worst case those types don't unite and we say that alpha equivalence isn't supported currently

> Alternatively we can just handle alpha-equivalence in expression unification. That would be more elegant, but I need to check how much code it would need.

I'm working on doing it in `similarExps`, I don't think it's too much work

catvayor:
I think I have something given that `applySubst` can be called from `similarExps`

athas:
It is pretty cool that this works:

```Futhark
def window [n][k] 't (A: [n+k-1]t) : [n][k]t =
  tabulate n (\j -> take k (drop j A))

def dot x y = f32.sum (map2 (*) x y)

def convolution v A = map (dot v) (window A)

entry main [n] (xs: [n]f32) =
  convolution [1,2,3] (xs :> [(n-2) + 3 - 1]f32)
```

The type of `convolution` is correctly inferred as

```
val convolution [n₀] [k₁] : (v: [k₁]f32) -> (A: [n₀ + k₁ - 1]f32) -> *[n₀]f32
```

@melsman you might enjoy this.

Examples like this are strong evidence that non-constructive sizes are useful.

src/Language/Futhark/TypeChecker/Terms.hs
```diff
@@ -213,7 +213,7 @@ checkCoerce loc te e = do
   te_t_nonrigid <- makeNonExtFresh ext te_t
 
   onFailure (CheckingAscription te_t e_t) $
-    unify (mkUsage loc "size coercion") e_t te_t_nonrigid
+    unify (mkUsage loc "size coercion") te_t_nonrigid e_t
```
athas, line 216: Is this just for better error messages, or does it actually matter?  Unification is supposed to be commutative.

catvayor: It does matter, unification is not commutative when two free variables are to be unified: the first one is replaced by the second one, which can cause problems to `checkDoLoop` if a variable of the merge pattern is just coerced

athas: Do we have a test that shows this?  This could be a bug in unification.  When we have a choice of which free variable to pick, maybe we should pick the one that is bound at the outermost level.

catvayor: if reversing if commit, `tests/loops/pow2reduce.fut` and `tests/issue1424.fut` fails at typing.

For `tests/loops/pow2reduce.fut` (the one I studied to solve the problem):
- the merge type, which is the type of initial value with all sizes being freed, is `[loop₄]i32`
- `xs :> ...`: `[loop₄]i32` is unified with `[coerce₁₄]i32`
- when the pattern is `normTypeFully`, it takes the type `[coerce₁₄]i32`
- so it's not one of the variable introduced by the loop, and its variation is not detected

catvayor: I'll try to make this without this PR

catvayor: in master
```futhark
def main (xs : [10]i64) =
    loop xs for i < 10 do
        let n = 10 + i
        in (xs :> [n]i64) ++ [i]
```

current state, `futhark check` replies
```
Error at test.fut:2:10-11:
Initial loop values do not have expected type.
Expected: [k₈]i64
Actual:   [10]i64

Dimensions "k₈" and "10" do not match.

Note: "k₈" is unknown size returned by "++" at 4:12-32.
```

with this modification, it just accept

src/Language/Futhark/TypeChecker/Terms.hs
```diff
@@ -342,6 +564,7 @@ checkExp (AppExp (BinOp (op, oploc) NoInfo (e1, _) (e2, _) loc) NoInfo) = do
   -- existential sizes, because it must by necessity be a function.
   (_, p1_t, rt, p1_ext, _) <- checkApply loc (Just op', 0) ftype e1_arg
   (_, p2_t, rt', p2_ext, retext) <- checkApply loc (Just op', 1) rt e2_arg
+  (rt'', retext') <- unscopeUnknowable loc rt'
```
athas, line 567: Why is this necessary?  I can remove it and all tests still pass.

src/Language/Futhark/TypeChecker/Terms.hs
```diff
@@ -439,7 +662,8 @@ checkExp (AppExp (Apply fe args loc) NoInfo) = do
           Var v _ _ -> Just v
           _ -> Nothing
   ((_, exts, rt), args'') <- mapAccumLM (onArg fname) (0, [], t) args'
-  pure $ AppExp (Apply fe' args'' loc) $ Info $ AppRes rt exts
+  (rt', exts') <- unscopeUnknowable loc rt
```
athas, line 665: Similarly, apparently also unnecessary.

catvayor: It's for the rule for not allowing existential size that are not in scope to appear in size expressions.

I've made this for `tests/shapes/existential-argument.fut`, and more specifically for a case like
```futhark
def quad_eval 't (f : () -> []t) : []t =
    let t = f () ++ f ()
    in t ++ t
```
Which is monomorphised as
```futhark
def double_eval ((f: () -> ?[d₀].[d₀]i64)) : ?[d₁₀].[d₁₀]i64 =
  let [d<{(+) d₅ d₆}>] [d₅] [d₆] (t: [d<{(+) d₅ d₆}>]i64) =
    (++) (f ()) (f ())
  let (d<{(+) ((+) d₅ d₆) ((+) d₅ d₆)}>: i64) =
    (+) ((+) d₅ d₆) ((+) d₅ d₆)
  in (++) t t
```
Somehow these `d₅` and `d₆` does not crash the compiler in latter stages, but I'm more confident with this.

athas: Interesting!  I removed this code and now this function

```
def double_eval 't (f : () -> []t) : []t =
    f () ++ f ()
```

is type-checked as

```Futhark
def double_eval_6007 't_5993 ((f_5995: () -> ?[d₀_5994].[d₀_5994]t_5993): () -> []t_5993) : [d₅_6001 +_494 d₆_6003]t_5993 =
  f_5995 () ++_5618 f_5995 ()
```

It is clearly nonsensical to refer to free variables `d₅_6001` and `d₆_6003` in the return type.  However, the fix should not be to strip unknown sizes in application - those are fine *locally*.  But we should strip them when generalising the type of functions.

Interestingly, we do this correctly for functions such as

```Futhark
def double_eval 't (f : () -> []t) : []t =
    f ()
```

which get type-checked as

```Futhark
def double_eval_6000 't_5993 ((f_5995: () -> ?[d₀_5994].[d₀_5994]t_5993): () -> []t_5993) : ?[d₂_5998].[d₂_5998]t_5993 =
  f_5995 ()
```

athas: Note the explicit quantification of the existential `d₂_5998` in the return type of the last example above.  That's the thing we're missing.  Probably the implementation of generalisation only considers size expressions that are plain variables.

catvayor: yes but
```futhark
def double_eval n (f : () -> []i64) : []i64 =
    f () ++ iota n
```
is type-checked as
```futhark
def double_eval (n: i64) ((f: () -> ?[d₁].[d₁]i64): () -> []i64) : [d₆ + n]i64 =
  f () ++ iota n
```
which is an incorrect type

catvayor: and allow the following to type-check!
```futhark
def main (n:i64) =
    zip (double_eval n (\() -> iota n)) (double_eval n (\() -> iota 2))
```

athas: I think the culprit is `noSizeEscape` in `checkFunBody` which removes these unknown sizes.

catvayor: maybe just call it over all return type ? I'm not sure they are evaluated

athas: No, wait, that's not the case.  `noSizeEscape` is actually quite innocent.

catvayor: we should also free them in function arguments, or correct monomorphisation regarding these sizes:
```futhark
def foo [n][m]'t (_ : [n+m]t) : (i64,i64) =
  (n, m)

def double_eval n (f : () -> []i64) =
    foo (f () ++ iota n)
```
monomorphised as
```futhark
def foo [d<{(+) n m}>] (n: i64) (m: i64) ((_: [d<{(+) n m}>]i64)) : (i64, i64) =
  (n, m)

def double_eval (n: i64) ((f: () -> ?[d₁].[d₁]i64)) : (i64, i64) =
  foo 0i64 0i64 ((++) (f ()) (iota n))
```
which is obviously wrong.

If not correcting monomorphisation, don't we end up with same expressivity ?

athas: The problem is in `closeOverTypes`, specifically the way it uses `injectExt` to figure out where to put the existential.  The current logic is to put it at the first constructive use - but there might not *be* a constructive use!  In this case we should replace the entire dimension with a fresh unknown size, just like what `unscopeUnknowable` used to do (but it should be done at function level, not after any application).

athas: To motivate, the reason I do not like invoking `unscopeUnknowable` after every application is because it disallowed this:

```Futhark
def double_eval 't (f : () -> []t) =
  let x = f ()
  let y = f ()
  in zip (x ++ y) (x ++ y)
```

catvayor: no, sizes of x and y will be marked as bounded by the let-bindings

catvayor: that's what allow us to write
```futhark
  let x = filter ...
  in map (x++) ...
```

athas: Oh, I had not noticed that machinery.  Is *that* still necessary if we switch to only using `unscopeUnknowable` (or equivalent) on function bodies?

athas: I get that this is more similar to the paper, but I'm not sure of the gain - and I am uncomfortable with having size variables that are also bound in the environment.

athas: I have not pondered the interaction with monomorphisation - I'm still stuck in a type checking mindset.

catvayor: I don't think so, we have to check what happen for this program
```futhark
def double_eval [n] (f : () -> []i64) (ma:[][n]i64) =
    let t = f()
    in map (t ++) ma
```
as
```futhark
def double_eval [n] (f : () -> []i64) (ma:[][n]i64) =
    map (f () ++) ma
```
still rejected, I've done the scoping machinery to change the nature of the size variable introduced by f

catvayor: removing let machinery lead to refuse
```futhark
def double_eval [n] (f : () -> []i64) (ma:[][n]i64) =
    let t = f()
    in map (t ++) ma
```

catvayor: > I am uncomfortable with having size variables that are also bound in the environment.

Why ? we can't refers to them, and they are freed the same way as the variable introduced in a 'classic' way, I don't think they can harm

athas: I also tried removing the `let` machinery, and the program you posted above (with the binding of `t`) type checks fine for me.  The one without first binding `t = f()` fails with a causality violation, which matches old (and safe) behaviour.

athas: This is what I did:

```
diff --git a/src/Language/Futhark/TypeChecker/Terms.hs b/src/Language/Futhark/TypeChecker/Terms.hs
index e2c66cd9c..5a9a6a8e5 100644
--- a/src/Language/Futhark/TypeChecker/Terms.hs
+++ b/src/Language/Futhark/TypeChecker/Terms.hs
@@ -671,15 +671,11 @@ checkExp (AppExp (LetPat sizes pat e body loc) _) =
     constraints <- getConstraints
     incLevel . bindingSizes sizes $ \sizes' ->
       bindingPat sizes' pat (Ascribed t) $ \pat' -> do
-        let implicitSizes = filter (isUnknown constraints) $ M.keys $ unFV $ freeInPat pat'
-            implicitIdents = map (\vn -> Ident vn (Info $ Scalar $ Prim $ Signed Int64) mempty) implicitSizes
-        binding False implicitIdents $ do
-          mapM_ observe implicitIdents
-          body' <- checkExp body
-          (body_t, retext) <-
-            unscopePatType loc (sizesMap sizes' <> patNames pat' <> S.fromList implicitSizes) =<< expTypeFully body'
-
-          pure $ AppExp (LetPat sizes' pat' e' body' loc) (Info $ AppRes body_t retext)
+        body' <- checkExp body
+        (body_t, retext) <-
+          unscopePatType loc (sizesMap sizes' <> patNames pat') =<< expTypeFully body'
+
+        pure $ AppExp (LetPat sizes' pat' e' body' loc) (Info $ AppRes body_t retext)
   where
     sizesMap = foldMap (S.singleton . sizeName)
     isUnknown constraints vn
```

catvayor: I had done the same thing, I think that I had not fully remove unscopeUnknowable.

Then I don't think we still need it, it's also done in Match.

athas: `tests/operator/size-section2.fut` (tricky program!) monomorphises to an invalid program with these changes, but it might just depend on a subtle implementation detail that's no longer true.  I will investigate.

athas: Oh, it's because I forgot to actually bring back `unscopeKnowable` for function return types...

athas: Nope, that was indeed something I had forgotten, but it was unrelated.

athas:
Apologies for breaking your branch (again) @catvayor, but I wanted you to see the current state of things.  tests/operator/size-section2.fut is the only one that doesn't work.

catvayor:
It comes from the way things are typed, this program is of type `[d₄ + 10]i64`
```futhark
let xs = filter (<5) t
 in xs ++ t
```
But it sounds incorrect to me, as this perfectly equivalent program is typed as `[d₈]i64` (probably with d₈ in the ext of appRes)
```futhark
let [n] xs: [n]i64 = filter (<5) t
 in xs ++ t
```

I see a solution to make this two program have the same type by reestablishing a part of let-machinery

athas:
While perhaps counterintuitive, I think those are the right types for those programs.  It is OK for unknown sizes to be used in size expressions.  It would be weird to me if `let xs = filter (<5) t in xs ++ t` and `filter (<5> t ++ t` had different types.

Regarding the second expression you list, its type is a necessary consequence of turning *entire size expressions* into unknown sizes when unscoping.  (This is actually something I have been wondering if we should revisit.)

catvayor:
the problem is that this size is not accessible, monomorphisation need to introduce these sizes to calculate them and give to some function, we still have problems with
```futhark
def foo [n][m]'t (_: [n+m]t) = (n,m)

def double_eval n (f : () -> []i64) =
    foo (f () ++ iota n)
```
`double_eval` is monomorphised as 
```futhark
def double_eval (n: i64) ((f: () -> ?[d₁].[d₁]i64)) : (i64, i64) =
  foo 0i64 0i64 ((++) (f ()) (iota n))
```
because we don't have access to the size at the when the args are infered

athas:
I see.  Well, the *right* way to solve that could be to use full normalisation:

```Futhark

def double_eval n (f : () -> []i64) =
    let tmp1 = f ()
    let tmp2 = iota
    let tmp3 = tmp1 ++ tmp2
    in foo tmp3
```

We know (from the `AppRes`) which of these expressions introduce the unknown sizes, and then we can bind them explicitly there.

But we can do that after the merge, since what we have now does work fine in the vast majority of cases, although the sizes are not always as precise as they should be.  In particular, `let xs = filter (<5) t in xs ++ t` should have type `[k+n]i32` where `k` is unknown.

athas:
Here is a full program that I would like to type check, but currently does not:

```Futhark
def f [n] [m] 't (_: [n+m]t) = n * m

def main [n] (xs: [n]i32) =
  f (let xs' = filter (<3) xs
     in xs' ++ xs)
```

catvayor:
removing my last commit should make it type, but the code produced will have the same problem as `foo (f () ++ iota n)`.

athas:
I agree.  The problem arises in monomorphisation, but we know how to solve it.  I forgot to check whether it works in the interpreter, but I think it does, since it implicitly normalises the program.

athas:
I'm bored and want to procrastinate, so let's just merge this and see what happens.

PR #1900, opened 2023-03-19
Working on Python style checking.

WilliamDue:
I will in this draft pull request try to use Black for style checking and auto formatting.
A problem with this is Black may violate the 79 line width limit on lines when using  `black --line-length=79`  and `black --line-length=79 --check` will not report this violation.
This may also be an advantage since some lines may be troublesome to get to that small.
An alternative is use pycodestyle for style checking which will report this line length error but pycodestyle will also report E203 which is not [pep8 compliant](https://github.com/psf/black/issues/315).
This can be ignored using a setup.cfg file but there may be other problems likes this one.


athas (changes requested):
Would it be possible to use a `black` made available through Nix, as with the Haskell tools, so that we can reproduce the CI environment locally?

.github/workflows/main.yml
```diff
@@ -159,6 +159,20 @@ jobs:
...
       run: nix-shell --pure --run "make check"
 
+
+  python-test-style:
+    runs-on: ubuntu-22.04
+
+    steps:
+    - uses: actions/checkout@v3
+    
+    - uses: psf/black@stable
```
athas, line 169: Surely this should be different now?

WilliamDue: It should be using `nix-shell` now, currently I have included a pyproject.toml in the root folder, is this okay? It is used to have some default settings when using black in the folder.

athas:
Looks good to me now.  Why is this a draft PR?

WilliamDue:
> Looks good to me now. Why is this a draft PR?

I was not sure if my Github action would work and I realize I forgot to ask do you want type checking with mypy too?
Since it seems like libraries like Matplotlib does not support it and then either I would have to insert ignore type checking comments in the code or [resolve](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports) them using py.typed I think.

athas:
If you can get mypy working without too much difficulty, then that would be preferable.

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.black]
+line-length = 79
+exclude = "benchmark-performance-plot.py"
```
athas, line 3: Add missing linebreak.

athas (approved):

PR #1904, opened 2023-03-23
Work on arbitrary size expressions

athas:
Main improvement is that they are now handled semi-properly in the interpreter, but I also fixed some test programs.

### DIFF
--- a/src/Language/Futhark/Interpreter/Values.hs
+++ b/src/Language/Futhark/Interpreter/Values.hs
@@ -83,8 +83,11 @@ typeShape (Scalar (Record fs)) =
   ShapeRecord $ M.map typeShape fs
 typeShape (Scalar (Sum cs)) =
   ShapeSum $ M.map (map typeShape) cs
-typeShape _ =
-  ShapeLeaf
+typeShape t
+  | Just t' <- isAccType t =
+      typeShape t'
+  | otherwise =
+      ShapeLeaf
 
 structTypeShape :: StructType -> Shape (Maybe Int64)
 structTypeShape = fmap dim . typeShape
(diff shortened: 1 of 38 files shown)


src/Language/Futhark/Interpreter/Values.hs
```diff
@@ -76,32 +76,21 @@ emptyShape :: ValueShape -> Bool
...
+typeShape _ =
+  ShapeLeaf
 
-structTypeShape :: M.Map VName ValueShape -> StructType -> Shape (Maybe Int64)
-structTypeShape shapes = fmap dim . typeShape shapes'
+structTypeShape :: StructType -> Shape (Maybe Int64)
+structTypeShape = fmap dim . typeShape
   where
-    dim (SizeExpr (IntLit d _ _)) = Just $ fromIntegral d
+    dim (SizeExpr (Literal (SignedValue (Int64Value d)) _)) = Just $ fromIntegral d
```
catvayor, line 92: Current implementation of sizes tends to prefer `IntLit` to `Literal` for pattern matching, notably `sizeFromInteger` which replace the old `ConstSize` will make `SizeExpr (IntLit ...)`, because `1` is parsed as an `IntLit`.
The fact that you have some seems to come from an error I've made in `evalType.matchPtoA` at [line 576-577](https://github.com/diku-dk/futhark/blob/arbitrary-size-expressions/src/Language/Futhark/Interpreter.hs#L576-577), at first I used `Literal` not knowing where which one to choose and I forgot this occurence

athas: I will change this to using `IntLit` consistently.

athas:
The interpreter is now fully operational.

Issue #1903, opened 2023-03-23
Smuggling sizes through functions


athas:
This program type-checks and works:

```Futhark
module m1 = {
  def f [n] (_: [n]bool -> bool) = n
}

def contrived (_: [10]bool) = true

entry main1 = m1.f contrived
```

`m1.f` uses a size that is only made available through a function.  Functions are "black box" and do not carry their sizes at runtime (unlike arrays), but this works because type inference can connect the size parameter `n` to the type of `contrived` (that is, the instantiation `n=>10` is used).

This program type-checks but does not work:

```Futhark
module m2 : {
  type^ t
  val x : bool -> t
  val f : t -> i64
} = {
  type^ t = []bool -> bool
  def x b = \(_: [10]bool) : bool -> b
  def f [n] (_: [n]bool -> bool) = n
}

entry main2 = m2.f (m2.x true)
```

Here the module ascription hides the fact that `f` actually has a size parameter, meaning that type checking is not instantiating anything when checking the expression `m2.f (m2.x true)`.  This would work if the size `n` was also the size of an array parameter (we call this a "constructive use", and means the array acts as a kind of witness), because we can fish out the size at run-time from the array shape, but functions do not (and *should not*) carry around the sizes of their parameter types.

@melsman Any ideas on this one?  I can think of a few ideas:

* Ban functions such as `f` that use a size only in a "non-constructive" manner.  We actually have similar restrictions elsewhere, so this is probably fine.  It means that definitions such as `def iiota: [n]i64 = iota n` that use "return type size polymorphism" would no longer work, but I'm not sure we ever agreed such definitions were desirable, and I'm almost certain no code actually uses them.  I was always a bit nervous about this bit of flexibility, and I don't think it was allowed in the original conception of the size type system.

* *If* this problem only occurs due to module ascriptions (I'm almost but not completely sure of that), then add some additional rules.  We *do* still want to allow module ascription to hide size parameters, as otherwise abstract size-lifted types become useless, but perhaps we could require that if you hide a size parameter, that size parameter must be used constructively (i.e. as an actual array size).

* Come up with a more complicated dynamic semantics that can make this work.  The programs above do actually work with the compiler (but not interpreter), because defunctorisation will add back the missing size parameters.  However, the whole point of the interpreter is to keep us honest about the fundamental simplicity of the language semantics, so this is a bit of a "semantics smell" to me.

catvayor:
The problem with `m2` appears at monomorphisation:
[...]

PR #1906, opened 2023-03-26
Start on more principled unification of sizes.

athas:
Only a single case is handled now (binary operators).  This will have to be tediously extended.


athas:
Also, this does a slightly orthorgonal refactoring to use `Exp` directly instead of `Size` in more places.

src/Language/Futhark/TypeChecker/Unify.hs
```diff
@@ -559,55 +559,46 @@ unifyWith onDims usage = subunify False
...
+  bcs
+  bound
+  nonrigid
+  (AppExp (BinOp (op1, _) t1 (x1, _) (y1, _) _) _)
+  (AppExp (BinOp (op2, _) t2 (x2, _) (y2, _) _) _) = do
+    unifySizes usage bcs bound nonrigid (Var op1 t1 mempty) (Var op2 t2 mempty)
+    unifySizes usage bcs bound nonrigid x1 x2
+    unifySizes usage bcs bound nonrigid y1 y2
+unifySizes _ _ _ _ d1 d2
+  | SizeExpr d1 == SizeExpr d2 = pure ()
```
athas, line 577: This is the case we should eventually get rid of.

athas:
This crazy @melsman-style type-level programming now not only type checks, but also works in the interpreter:

```Futhark
module meta: {
  type eq[n][m]

  val coerce [n][m]'t : eq[n][m] -> [n]t -> [m]t
  val coerce_inner [n][m]'t [k] : eq[n][m] -> [k][n]t -> [k][m]t

  val refl [n] : eq[n][n]
  val comm [n][m] : eq[n][m] -> eq[m][n]
  val trans [n][m][k] : eq[n][m] -> eq[m][k] -> eq[n][k]

  val plus_comm [a][b] : eq[a+b] [b+a]
  val plus_assoc [a][b][c] : eq[(a+b)+c] [a+(b+c)]
  val mult_comm [a][b] : eq[a*b] [b*a]
  val mult_assoc [a][b][c] : eq[(a*b)*c] [a*(b*c)]
} = {
  type eq[n][m] = [0][n][m]()

  def coerce [n][m]'t (_: eq[n][m]) (a: [n]t) = a :> [m]t
  def coerce_inner [n][m]'t [k] (_: eq[n][m]) (a: [k][n]t) = a :> [k][m]t

  def refl = []
  def comm _ = []
  def trans _ _ = []

  def plus_comm = []
  def plus_assoc = []
  def mult_comm = []
  def mult_assoc = []
}

def main [n][m] (xs: [n]i32) (ys: [m]i32) =
  zip (xs ++ ys) (meta.coerce meta.plus_comm (ys ++ xs))
```

Next step is to make it work in the compiler as well.

catvayor:
this implies to solve #1903

athas:
@catvayor For handling complex expressions in momomorphisation, [perhaps this is useful as inspiration](https://github.com/diku-dk/futhark/blob/ee649e098a612eee6d61893277d9c5d105cfe56c/src/Language/Futhark/Interpreter.hs#L154-L201).

athas:
Also, feel free to merge this into your branch whenever you want.  It has already lived longer than I expected it to.

catvayor:
> Also, feel free to merge this into your branch whenever you want. It has already lived longer than I expected it to.

I was keeping it alive until it's goal was done

catvayor:
I tried to make a test of the type-madness program
```futhark
-- input { [1,2] [3,4] } output { [ (1,3), (2,4), (3,1), (4,2) ] }
```
but `futhark test` told me `unexpected '('`, is it voluntary to not allow tuple, or do I wrote it a bad way ?

athas:
Test blocks only support arrays of primitives.  Change the program to `unzip` the result.

PR #1908, opened 2023-03-30
Tests for more exotic cases in size expressions

athas:
Also includes fixes (I hope), because the problems were in obscure corners I'd be embarrassed to ask you to look too much at.

(Also includes a merge of master.)

### DIFF
--- a/src/Futhark/Internalise/NormaliseSize.hs
+++ b/src/Futhark/Internalise/NormaliseSize.hs
@@ -566,11 +566,7 @@ removeExpFromValBind valbind = do
     runWriterT (runReaderT (arrowArgRetType args rety') (scope, mempty))
   let newParams =
         funArg
-          `S.union` S.fromList
-            ( if null params' && isNothing (valBindEntryPoint valbind)
-                then filter (`elem` M.elems extNaming) $ retDims rety''
-                else argsParams
-            )
+          `S.union` S.fromList argsParams
       rety''' = arrowCleanRetType newParams rety''
       typeParams' =
         valBindTypeParams valbind


tests/modules/ascription-error7.fut
catvayor, line 1: What I understood of our discussion about constructive use of size is that they are needed for hidden size variables, why require it from parameters ?

athas:
I modified the loop test to be more interesting, so now internalisation doesn't work.

catvayor:
currently I think it's normalized as `def sum [k][n] (xs: [n]i32) : i32 = ...` and then the `[k]` is removed at monomorphisation because unused in the function type

athas:
Monomorphisation can't do better than that: since `k` is not used, it has no idea what it should be.

I will investigate something radical: maybe we don't need to do normalisation, or at least not this aggressively.

athas:
This PR now also fixes some omissions in (ultimately) monomorphisation.  Monomorphisation will now turn `sum` into this:

```Futhark
def sum (k: i64) ((xs: [2 ** k]i32)) : i32 =
  head (loop [loop₄] (xs: [loop₄]i32) = xs for (i: i64) in reverse (iota k) do
         step 0i64 (xs :> [2 ** (i + 1)]i32))
```

Note how the size parameter is now an explicit normal parameter.  Similarly, monomorphisation also fixes up the uses of `sum` to pass corresponding arguments, based on how it was instantiated.  You may find some inspiration in how that is done (ultimately, it uses the same expression unification machinery as the type checker itself and the interpreter).

I do think monomorphisation is the right place to do this, because that is where we already did it.  It is not clear to me whether full size normalisation should be a pass by itself or rolled into monomorphisation, but I think *this part* should definitely be in monomorphisation, and perhaps size normalisation should run *after* monomorphisation.  However, I admit I have not thought deeply about how existentials in return types are handled here, while it sounds like you have.

catvayor:
I don't really know if normalization is should be a separated pass, but monomorphisation should modify size expressions to do the transformation it does on expression (most importantly binop removal), else the normalization pass will reproduce them, giving errors on latter stages.

In the other hand, to apply the size parameter deduction, it has to keep the expression, or maybe we can give the map from names to expression resulting of normalization to monomorphisation and then do our size deduction on this.

catvayor:
canceling the checks, somehow a test seem to place the compiler into a loop (`tests/issue1068.fut` at defunctionalisation)

catvayor:
I'm adding trace step by step, now I have the problematic expression a `g (f {0=0, 1=1})` that apparently appear somewhere

athas:
Good catch!  Let me take a look.

catvayor:
If I caught the idea of what does record replacements, I do not know how it could work before, but it seems required now.
Also, maybe we should catch them for `ForIn` case as it also make a pattern ?

athas:
It used to work partially because there was a bug where I forgot to use `transformPat` for do-loops, which was fine without size expressions.  And it *sort of* continued to work even in that case, since the whole record-replacement stuff is more of an optimisation than strictly necessary.

The loop pattern is in principle only in scope for the `While` case, although I guess it doesn't matter much.  I have pushed your suggested fix.  Time to move on to new bugs.  In particular, I noticed that this does not work:

```Futhark
def f (p: {a:i64,b:bool}) : [p.a]i64 = iota p.a
```

It fails with a type error, but this seems like the sort of stuff we should want to allow.

athas:
(But not on this branch; let's close up this mess before we get into a new fight.)

catvayor:
yes, I will try to find the remaining cases where binOp are escaping monomorphisation through sizes then go on this problem

athas:
The real reason we have this trouble is that we don't have a dedicated pass *just* after type checking that normalises/desugars expressions (e.g. removing BinOps and operator sections).  Instead it's kind of jammed into monomorphisation.  That worked OK for a while, but now the language is growing more complicated and it's too easy to miss cases.  We'll continue doing it this way for now, but this gets put on The List for a cleanup.

athas:
The last problems (e.g. tests/shapes/toplevel0.fut) seem to be caused by normalisation, which does not correctly handle non-function top-level bindings with expression sizes (it appears to throw away the existentials).  This is a somewhat subtle aspect, so no great surprise.

catvayor:
I currently testing the removal of this, before it was required to be handled by monomorphisation, now it should be required to not do anything special due to the new work of monomorphisation

catvayor:
Should be good now, I let the checks run then I'll merge

athas:
You don't have to wait for the benchmark tests.  I think the cluster that runs them is overloaded.

catvayor:
I remember too see fail in them because they are different from testing, even if it was just because of unflatten, I prefered to wait at least one

PR #1910, opened 2023-04-04
An entry point test and fix.

athas:


catvayor:
> I'm investigating (and fixing) various size-expression-related things related to entry points. One annoying part is that I don't see how we can make this work:
> 
> ```futhark
> entry main (x: i64) (_: [x+1]i32) = x
> ```
> 
> Or rather, it works but the constraint is lost, so you can pass it whatever you want without complaints. Fixing this will take some clever thinking.

How are arguments of entry points supposed to be checked ? I mean, if they are called from the futhark program, type checking should constrain things so that the function have what it asks (for the type at least), but how it currently works for
```futhark
entry main (x: i64) (_: [x]i32) = x
```
?

athas:
Ultimately the code generator produces a check, but it can only check things that can be expressed in the core language.  The solution is probably to insert an `Assert` expression at some point, but I'm still pondering whether we need to extend the core IR instead.  Entry points are always a bit of a sore spot, because they are at the intersection of Futhark and the outside world.

The nuclear option would be to ban non-constructive size expressions in entry point parameters, but I think that is going too far.

athas:
In particular, for the program above, the core IR becomes this:

```Futhark

fun
  main_5999 (d<{(+) x 1}>_6004 : i64,
             x_6005 : i64,
             nameless_6006 : [d<{(+) x 1}>_6004]i32)
  : {i64} = {
  {x_6005}
}

entry("main",
      {x: i64,
       _: []i32},
      {i64})
  entry_main (d<{(+) x 1}>_6007 : i64,
              x_6008 : i64,
              nameless_6009 : [d<{(+) x 1}>_6007]i32)
  : {i64} = {
  let {entry_result_6010 : i64} =
    apply main_5999(d<{(+) x 1}>_6007, x_6008, nameless_6009)
    : {i64}
  in {entry_result_6010}
}
```

Note how there are no constraints expressed.  We could insert an `Assert` when we normalise those functions that are also entry points.

catvayor:
I don't know at which point it would be, but maybe we can reuse the `entryAscribed` of parameters to introduce asserts

athas:
I think this works now.  Turns out the mapping you construct while normalising the parameter types is exactly what we need.

I am not satisfied with the quality of the error messages this will produce, but at least errors will be caught, and it was done in a very local way.  It can be improved another day.

athas:
The errors are spurious; feel free to merge when you wish.

PR #1911, opened 2023-04-04
Simplify normalisation pass.

athas:
In particular, it was excessively monadic.  There was no need for the outer monad.  Also, the ArrowArgM did not need to have an inner monad.

You may consider whether ArrowArgM needs to exist at all.  I suspect those functions would be simpler if they were written in a conventional functional style.

Similarly, Writer looks really seductive at first, but it's slow and when things get complicated, code using it is difficult to read (e.g. 'pass').  I used to use it a lot, but now I always use State instead.

It's pretty common for people who are experienced functional programmers, but new to Haskell, to over-monadify their code.

### DIFF
--- a/src/Futhark/Internalise/NormaliseSize.hs
+++ b/src/Futhark/Internalise/NormaliseSize.hs
@@ -426,9 +426,8 @@ onExp e =
           pure $ sizeFromName (qualName vn) (srclocOf e)
 
 -- | Monad for arrowArg
--- Reader (scope, dimtoPush)
 -- Writer (arrow arguments, names that can be existentialy bound)
-type ArrowArgM a = ReaderT (S.Set VName, [VName]) (Writer (S.Set VName, S.Set VName)) a
+type ArrowArgM a = Writer (S.Set VName, S.Set VName) a
 
 -- | arrowArg takes a type (or return type) and returns it
 -- with the existentials bound moved at the right of arrows.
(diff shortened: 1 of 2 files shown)


src/Futhark/Internalise/NormaliseSize.hs
```diff
@@ -502,57 +483,59 @@ arrowCleanType paramed (Scalar ty) =
...
+  let scope' = scope `S.union` args `S.union` new_params
+  (body', exp_naming'') <-
+    runNormaliseM (expFree $ valBindBody valbind) scope' exp_naming'
+  let new_names = exp_naming'' `M.difference` exp_naming'
+  body'' <-
+    expReplace exp_naming'
+      <$> foldrM insertDimCalculus body' (M.toList $ canCalculate scope' new_names)
+
+  pure
+    ( S.insert (valBindName valbind) scope,
```
catvayor, line 525: shouldn't it be prev_scope ? scope also has the size parameters of the current binding

athas: Yes.  Fixed.

athas:
Feel free to merge this whenever you are satisfied with it.

catvayor:
You don't have any other remark on it ?

athas:
I can always make another PR later.  No reason to keep these things fractured.

PR #1914, opened 2023-04-11
Merge Normalisation and Monomorphisation

catvayor:
They are both very linked and it's hard to say that one step should be done before the other, so let's just merge them.

The modifications I'm planning to do:
- make a transformExp (resp transformType, ...) function, doing `Monomorphise.transformExp` and `NormaliseSize.expFree` at the same time
- include the work of `normaliseValBind` inside `monomorphiseBinding`

To solve our problem coming from size parameters not being used constructively (#1897), the objective is to insert an `NormaliseSize.onExp` operation during size argument inference of `Monomorphise.transformFName`.


catvayor:
now, the program
```futhark
def contrived [n] (i: i64) (_: [n]i64 -> [n](i64,i64)) = i + n

def main (xs: []i64) =
  let ys = iota (i64.sum xs)
  in loop acc = 0 for i < 10 do
       acc + contrived i (\p -> zip ys p)
```
is monomorphised & normalised as 
```futhark
entry main [d₀] ((xs: [d₀]i64)) : i64 =
  let (d<{(sum xs)}>: i64) = (sum xs)
  let [sum] (ys: [d<{(sum xs)}>]i64) =
    iota (sum xs)
  in loop (acc: i64) = 0 for i < 10 do
       (+) acc (contrived d<{(sum xs)}> i (\(p: [d<{(sum xs)}>]i64) ->
                                            zip ys p))
```
still not perfect, but the complexity impact is removed

athas:
By "not perfect" you mean the fact that the `sum` is still computed twice?

catvayor:
yes, this can be improved by doing some expReplace in insertDimCalculus

catvayor:
Now it's monomorphised as
```futhark
entry main [d₀] ((xs: [d₀]i64)) : i64 =
  let (d<{(sum xs)}>: i64) = (sum xs)
  let (ys: [d<{(sum xs)}>]i64) =
    iota d<{(sum xs)}>
  in loop (acc: i64) = 0 for i < 10 do
       (+) acc (contrived d<{(sum xs)}> i (\(p: [d<{(sum xs)}>]i64) ->
                                            zip ys p))
```
and I think it's difficult to make it better for this one.

When a size arise inside in another one like in 
```futhark
def contrived [n] (i: i64) (_: [n+1]i64 -> [n+1](i64,i64)) = i + n

def main (xs: []i64) =
  let ys = iota (i64.sum xs + 1)
  in loop acc = 0 for i < 10 do
       acc + contrived i (\p -> zip ys p)
```
it can be monomorphised in a pretty way
```futhark
entry main [d₀] ((xs: [d₀]i64)) : i64 =
  let (d<{sum xs}>: i64) =
    sum xs
  let (d<{((+) (sum xs) 1)}>: i64) = ((+) d<{sum xs}> 1)
  let (ys: [d<{((+) (sum xs) 1)}>]i64) =
    iota d<{((+) (sum xs) 1)}>
  in loop (acc: i64) = 0 for i < 10 do
       (+) acc (contrived d<{sum xs}> i (\(p: [d<{((+) (sum xs) 1)}>]i64) ->
                                          zip ys p))
```
but this is not guarantied, as seen in
```futhark
def contrived [n] (i: i64) (_: [n+1]i64 -> [n+1](i64,i64)) = i + n

def main (xs: []i64) =
  let ys = iota (i64.sum xs)
  in loop acc = 0 for i < 10 do
       acc + contrived i (\p -> zip (ys ++ [0]) p)
```
being monomorphised in
```futhark
entry main [d₀] ((xs: [d₀]i64)) : i64 =
  let (d<{(+) (sum xs) 1}>: i64) =
    (+) (sum xs) 1
  let (d<{(sum xs)}>: i64) = (sum xs)
  let (ys: [d<{(sum xs)}>]i64) =
    iota d<{(sum xs)}>
  in loop (acc: i64) = 0 for i < 10 do
       (+) acc (contrived d<{(sum xs)}> i (\(p: [d<{(+) (sum xs) 1}>]i64) ->
                                            zip ((++) ys [0]) p))
```
If such unfortunate case arise and are not caught during optimisation, correct this would require much more work, but not hard work I think.

I was also thinking about the multiple calculations arising from
```futhark
def f0 xs = xs ++ xs
def f1 xs = f0 xs
def f2 xs = f1 xs
...
```
and I think that, thanks to the guaranty that the `d+d` being calculated is calculated in the caller, we can force it to be an 'explicit size parameter' and avoid multiple evaluation at the price of an extra argument, but is it really interesting ?

PR #1919, opened 2023-04-18
implement alpha equivalence

catvayor:


catvayor:
> I'm also still on the fence about whether it's useful, since decaying to an unknown size is so easy and obviously works.

you are probably right, I though it would be easy but I just realized that this need to be done also on LetFun & LetWith, which sounds like a lot of work for extremely specific cases.

athas:
Are we giving up on this?  I personally think the complexity outweighs the gain.

catvayor:
I think so yes, I will implement the let's being forbidden

PR #1921, opened 2023-04-19
Sum-of-products support for solving equalities and inequalities.

zfnmxt:
This PR adds functionality to convert expressions in the source AST and (probably) the IR AST to a sum-of-products (sop) representation, which are things that look like
```
5 + x + 4*y + 2*x*y*z*z + k*m
```
i.e., sums of products of variables scaled by integers. The utility of the representation is that it's normalized, i.e., there is only one way to write each term. For example, the terms `x + (y +z)` and `z + (x + y)` have the same sop representation. (This isn't quite true since you can tack on arbitrary amounts of zero terms. I.e., `x = x + 0*y`, but it's easy to get rid of superfluous terms like this.)

The idea is to traverse the expressions of a program and accumulate an environment of `VName`-to-sum-of-products mappings while simultaneously inferring sop-based inequalities on `VName`s. For example, from

```futhark
foo : [m]int64 = filter p (iota n)
```
we can infer that `0 <= m <= n`. By building up an environment of equalities and ranges, we can answer equality and inequality queries (via [Fourier-Motzkin elimination](https://en.wikipedia.org/wiki/Fourier%E2%80%93Motzkin_elimination)). 

This PR will build on work being done in #1897 by enabling the compiler to discover equivalences between syntactically-distinct size expressions. The inferred ranges can also be used to remove things like dynamic bound checks.

### DIFF
--- a/futhark.cabal
+++ b/futhark.cabal
@@ -274,6 +274,15 @@ library
       Futhark.Construct
       Futhark.Doc.Generator
       Futhark.Error
+      Futhark.SoP.AlgEnv
+      Futhark.SoP.AlgEnvProc
+      Futhark.SoP.EquivEnvProc
+      Futhark.SoP.FourierMotzkin
+      Futhark.SoP.PrimExp
+      Futhark.SoP.RangeEnvProc
+      Futhark.SoP.SoP
+      Futhark.SoP.ToFromSoP
+      Futhark.SoP.Util
       Futhark.FreshNames
       Futhark.IR
       Futhark.IR.Aliases
@@ -514,6 +523,7 @@ library
     , mwc-random
     , prettyprinter >= 1.7
     , prettyprinter-ansi-terminal >= 1.1
+    , multiset
 
 executable futhark
   import: common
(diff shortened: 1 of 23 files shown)


athas:
It'll be at least a week before I have the spare clock cycles to seriously ponder this, but there is one chore we'll eventually have to address: figure out where PrimExp should live.  In this PR, it's duplicated into a separate module under Futhark.SoP.  Previously it lived in the compiler backend.  It may be time to split Futhark into multiple internal packages, corresponding to at least frontend, backend, and primexp/arithmetic.  The goal is to ensure that the frontend does not depend on the backend.  The shared parts should be factored into a separate internal package.

rowanG077:
Not sure if this will help you but there is a haskell Type checker plugin that does exactly that: https://github.com/clash-lang/ghc-typelits-natnormalise

You might get some inspiration from it.

athas:
Yes, the existence of that one gave me confidence that this is viable to do in a type checker, but I haven't studied how it actually works.

athas:
Closing as superseded by refinement typing PR.

PR #1925, opened 2023-04-24
More expressive existentials in size expression

catvayor:
The Objective is to type check & compile correctly this program:
```futhark
def f [n] [m] 't (_: [n+m]t) = n * m

def main [n] (xs: [n]i32) =
  f (let xs' = filter (<3) xs
     in xs' ++ xs)
```

This will require full-normalization, to make it look like
```futhark
def f [n] [m] 't (_: [n+m]t) = n * m

def main [n] (xs: [n]i32) =
  let [d] (xs': [d]i32) = filter (<3) xs
  let (tmp: [d+n]i32) = xs' ++ xs
  f tmp
```

This will also be required for
```futhark
def f [n] [m] 't (_: [n+m]t) = n * m

def foo n (f : () -> []i64) =
    foo (f () ++ iota n)
```
to make it look like
```futhark
def f [n] [m] 't (_: [n+m]t) = n * m

def foo n (f : () -> []i64)  =
  let [d] (tmp_0: [d]i64) = f ()
  let (tmp_1: [n]i64) = iota n
  let (tmp_2: [d+n]i64) = tmp_0 ++ tmp_1
  f tmp_2
```


catvayor:
now
```futhark
def partition [n] 'a (p: a -> bool) (as: [n]a): ?[k].([k]a, [n-k]a) =
  let p' x = if p x then 0 else 1
  let (as', is) = intrinsics.partition 2 p' as
  let (q:i64) = is[0]
  in (take q as', drop q (as' :> [n]a))
```
types well, still have to coerce because of the unknown variable in the return type of `intrinsics.partition`, e.g. `?[k]([k]a, *[n]i64)`

athas:
Hm, that `intrinsics.partition` type is still wrong.  Let me try fixing it again.

athas:
Alright, now it's this:

```
> :t intrinsics.partition
partition 'a₀ [n₁] : i32 -> (a₀ -> i64) -> [n₁]a₀ -> ?[k].(*[n₁]a₀, *[k]i64)
```

catvayor:
I'm not sure the use of unscopeUnknowable on return type is right in checkBinding, the return type `?[k].([k]a, [n-k]a)` is silently replaced by `?[d₁₆][d₁₇].([d₁₆]a, [d₁₇]a)` (which is wrong, but I think I can correct this by using witnesses), but I think in case it's correct (for a return type `?[k].[n-k]a` for example) it should produce an error if the return type is explicitly precised.

athas:
If the user enters a return type annotation of the form `?[k].[n-k]a` (or some other nonconstructive use), that is caught during elaboration of the type expression itself.

You are right that `unscopeKnowable` currently throws away too much information.

catvayor:
fine then, I've modified unscopeUnknowable

athas:
One reason it fails is that the substitution in `instantiateDimsInReturnType` is not complete (only looks at Var expressions).

catvayor:
```futhark
def f [n] [m] 't (_: [n+m]t) = n * m

def main [n] (xs: [n]i32) =
  f (let xs' = filter (<3) xs
     in xs' ++ xs)
```
is now well compiled,
normalisation gives
```futhark
def f [n] [m] 't ((_: [n + m]t): [n + m]t) : i64 =
  let (e<{n * m}>: i64) =
    n * m
  in e<{n * m}>

entry main [n] ((xs: [n]i32): [n]i32) : i64 =
  let (e<{filter (< 3) xs}>: [d₇]i32) =
    filter (< 3) xs
  let (xs': [d₇]i32) = e<{filter (< 3) xs}>
  let (e<{xs' ++ xs}>: [d₇ + n]i32) =
    xs' ++ xs
  let (e<{f (e<{xs' ++ xs}>)}>: i64) =
    f (e<{xs' ++ xs}>)
  in e<{f (e<{xs' ++ xs}>)}>
```
and the complete result of monomorphisation is
```futhark
def filter [n] ((p: i32 -> bool)) ((as: [n]i32)) : ?[d₁₃].*[d₁₃]i32 =
  let [k₇] {0=(0: [n]i32), 1=(1: [k₇]i64)} =
    partition 1 (\(x: i32) ->
                  let (e<{p x}>: bool) =
                    p x
                  let (e<{if e<{p x}>
                      then 0
                      else 1}>: i64) =
                    if e<{p x}>
                    then 0
                    else 1
                  in e<{if e<{p x}>
                     then 0
                     else 1}>) as
  let ((as': [n]i32), (is: [k₇]i64)) = {0=0, 1=1}
  let (d<{is[0]}>: i64) =
    is[0]
  let (e<{is[0]}>: i64) = d<{is[0]}>
  let (e<{as'[:e<{is[0]}>]}>: [d<{is[0]}>]i32) =
    as'[:e<{is[0]}>]
  in e<{as'[:e<{is[0]}>]}>

def ++ [n] [m] ((xs: [n]i32)) ((ys: [m]i32)) : ?[d<{(+) n m}>].*[d<{(+) n m}>]i32 =
  let (d<{(+) n m}>: i64) =
    (+) n m
  let (e<{concat xs ys}>: [d<{(+) n m}>]i32) =
    concat xs ys
  in e<{concat xs ys}>

def f [d<{(+) n m}>] (n: i64) (m: i64) ((_: [d<{(+) n m}>]i32)) : i64 =
  let (e<{n * m}>: i64) =
    (*) n m
  in e<{n * m}>

entry main [n] ((xs: [n]i32)) : i64 =
  let [d₇] (e<{filter (< 3) xs}>: [d₇]i32) =
    filter (let (x: i32) = 3
            in \(x: i32) ->
                 (<) x x) xs
  let (d<{(+) d₇ n}>: i64) =
    (+) d₇ n
  let (xs': [d₇]i32) = e<{filter (< 3) xs}>
  let (e<{xs' ++ xs}>: [d<{(+) d₇ n}>]i32) =
    (++) xs' xs
  let (e<{f (e<{xs' ++ xs}>)}>: i64) =
    f d₇ n (e<{xs' ++ xs}>)
  in e<{f (e<{xs' ++ xs}>)}>
```
It's pretty verbose, I think we can do better but working if you're not using letWith, and I still have to decide if I should replace by names the constructs `Update`, `RecordUpdate`, `OpsectionLeft`, `OpSectionRight` and `Lambda`.

Also, what should we do with attributes ?

catvayor:
what does this means ? `SegMap(gpu-mem)  should have occurred 1 times, but occurred 2 times.`

athas:
It's a structure test - it means that the program is now optimised differently (at the GPU representation level).  That's unexpected for these kinds of changes.  It *could* happen if you duplicate code.  When does it happen?

catvayor:
In multiple test (many in fact), I think nearly all test with LetWith, that's the only thing I've changed, and all undefined error became these.

catvayor:
seeing a fail, it may also come from my complete removal of attributes, as I don't know what to to with them currently

catvayor:
currently, for LetWith I do
`let x[i] = e1 in e2` -> `*lets of e1 in*` `let x[i] = e1' in` `*lets of e2 in* e2'`

athas:
That sounds like it should work as well, but if we want to take the word "normalise" seriously, maybe we should also eliminate redundant syntax.  (E.g maybe the BinOp/section stuff we do in monomorphisation is also better placed in normalisation.)

catvayor:
I'll move them latter, I'll try to make everything compile well first

athas:
Yes, try that.  Attributes are a bit ad hoc in how they propagate.  I wonder if there's any theory about them.  Probably not.

catvayor:
6 optimization modifications left:
- 3 inline things
tests/noinline/noinline2.fut, tests/noinline/noinline1.fut, tests/noinline/noinline4.fut
all of them use #[noinline] attribute, maybe it's to much ? or normalisation broke it ?
- 3 assert things
tests/migration/assert0.fut, tests/migration/assert2.fut, tests/migration/assert1.fut
these seems to have removed their assert at optimisation, which is sad

athas:
> 3 inline things
>   tests/noinline/noinline2.fut, tests/noinline/noinline1.fut, tests/noinline/noinline4.fut
>  all of them use #[noinline] attribute, maybe it's to much ? or normalisation broke it ?

Oh right, applying `#[noinline]` to to operands may cause that... in `#[noinline] f (g x)`, only `f` should be noinlined.

> these seems to have removed their assert at optimisation, which is sad

Yes, very sad.

athas:
`&&` and `||` also need to be treated specially (possibly you can just turn them into `If` - up to you).

catvayor:
to clarify with attributes, some should be applied to every sub-expressions and some others only on outer-most expression. 
It sometimes has to be kept on outside of asserts (`#[unsafe]`, I suppose it should affect every sub-exp), but every time ?
And there is nothing that allow us to decide what to do with an attribute ? How am I suppose to deal with them ?

catvayor:
tests/issue1501.fut looks like after the internalization, the call
`stress_7447(straight_res_7825, 1_7845, 1_7848, 1_7851, 2_7839, 2_7840, 2_7841)`, with
- `1_7845, 1_7848, 1_7851` of type `[d₃_7842]i32`
- `2_7839, 2_7840, 2_7841` of type `[straight_res_7825]i32`
- signature
```
  stress_7447 (d_8244 : i64,
               0_8245 : [d₃_7854]i32,
               1_8246 : [d₃_7854]i32,
               2_8247 : [d₃_7854]i32,
               0_8248 : [d_8244]i32,
               1_8249 : [d_8244]i32,
               2_8250 : [d_8244]i32)
```
don't type check, I think it's because d₃_7854 is not d₃_7842 (I didn't touch to size substitution I think), but I don't know why these unknown var was changed differently.

looking at the normalized code, we that these d₃ comes from the existential size of the constant
`straight_6046 : ?[d₃_6044].([d₃_6044]f32, ........)`, this size is then used as a global constant in latter valbind.
I remember seeing an issue with global sizes in internalization, could it be the same problem re-appearing for some reasons ?

athas:
I haven't thought through everything yet, but here's some remarks.

The issue with attributes is frustrating and annoying.  The treatment of attributes in the rest of the compiler is also ad-hoc, and depends on the attribute in question.  Handling them in an ad-hoc manner (e.g. propagating the `incremental_flattening` ones but not the `inline` ones) may be simply following established practice.  Perhaps you can get some [inspiration from how they are currently handled in internalisation](https://github.com/diku-dk/futhark/blob/5c0097f3ec81b31f3d2eb50c845c4b98ad624602/src/Futhark/Internalise/Exps.hs#L744-L764)

Regarding tests/issue1501.fut, after monomorphisation it appears that a size `d₃_6044` is bound both as a top level existential (i.e. as a top level variable) but then also locally in the entry point, and as an existential in `stress_7447`.  This is wrong.

catvayor:
monomorphisation looks lost about if it should consider `d₃_6044` as known or not, and it's probably because I am too.
It's first bounded as an existential in the type of a top-level constant, should we consider it as known ? If not, what to do for foo in a program like
```futhark
let t = filter ...
let foo xs = zip t xs
```
xs should obviously have the same size as t, do we just write forall size as typing don't really have problems with it ?

athas:
```
 It's first bounded as an existential in the type of a top-level constant, should we consider it as known ?
```

Yes, that's the idea.  "Existentials" at the top level aren't *really* existential - you should consider them as completely normal bound variables.  This means you need to be careful when looking at `ValBind`s and making sure that *if* the `ValBind` has no parameters, to extract the existentials from their `valBindRetType` and bind them in the same symbol table that you bind the `valBindName`.  This is a *very* subtle thing, so I'm not surprised you're confused.  That is what the comment says, too:

```
    -- | If 'valBindParams' is null, then the 'retDims' are brought
    -- into scope at this point.
```

But it's probably the kind of cryptic comment that only makes sense if you already know what it's supposed to say.

catvayor:
ok, this is corrected, but it broke tricky, one of those sizes became unknown after internalization

catvayor:
in master, this sizes isn't named in main, so it does not make problems
before marking those sizes as known in this PR, one of the let bind was catching this sizes
nothing declare it again in main as it should be available, but the internaliser somehow deleted it, I think it's because it understood that in reality, the arrays was of sizes `three`:
in the IR:
```futhark
let {three_7376 : ({}, i64)} =
  3i64
let {tmp_7377 : ({}, [three_7376]i32)} =
  apply replicate_7311(three_7376, 1i32)
  : {*[three_7376]i32}
let {tmp_7378 : ({tmp_7377}, [three_7376]i32)} =
  tmp_7377
let {v_res_7379 : ({}, i64)} =
  apply dyn_lifted_lambda_7324(three_7376)
  : {i64}
-- ...
fun
  dyn_lifted_lambda_7324 (three_7436 : i64)
  : {i64} = {
  {three_7436}
}
```
however, it did take the time the change all d₆_6000

athas:
Right, this looks like an internaliser bug that just didn't matter before.  I will see if I can figure out the cause.

athas:
This is a snippet of `FUTHARK_COMPILER_DEBUGGING=2 futhark --mono tricky.fut` on master:

```
def v_6009 : ?[d₆_6008].([d₆_6008@(i64)]i32, i32 -> [d₆_6008@(i64)]i32) =
  #[d₆_6008]
  (let (three_5998: i64) = 3@(i64)
   in (replicate_6022@((n_5664: i64) -> (x_5665: i32) -> *[n_5664@(i64)]i32) three_5998@(i64) 1@(i32), \(i_6003: i32) ->
        replicate_6022@((n_5664: i64) -> (x_5665: i32) -> *[n_5664@(i64)]i32) three_5998@(i64) i_6003@(i32)))
```

And this is on your branch:

```
def v_6001 : ?[d₆_6000].([d₆_6000@(i64)]i32, i32 -> [d₆_6000@(i64)]i32) =
  #[d₆_6000]
  (let (three_5990: i64) = 3@(i64)
   #[d₆_6000]
   (let (tmp_7303: [three_5990@(i64)]i32) =
      replicate_7311@((n_5656: i64) -> (x_5657: i32) -> *[n_5656@(i64)]i32) three_5990@(i64) 1@(i32)
    in (tmp_7303@([three_5990@(i64)]i32), \(i_5995: i32) ->
         let (tmp_7304: [three_5990@(i64)]i32) =
           replicate_7311@((n_5656: i64) -> (x_5657: i32) -> *[n_5656@(i64)]i32) three_5990@(i64) i_5995@(i32)
         in tmp_7304@([three_5990@(i64)]i32))))
```

The `#[d₆_6000]` notation is some syntax I made up for the `appResExt` (in retrospect it is bad, because it's the same as for attributes).  It shouldn't show up twice.  I can't figure out why that might cause this problem, though.

athas:
Oh, that's not the problem: the problem is that the corresponding `appResType` for those does not actually refer to the sizes in the `appResExt`.  It's annoying to debug this because we don't prettyprint them...

athas:
I "improved" (changed) the printing of AppRes with `FUTHARK_COMPILER_DEBUGGING=2`.  Now we can see that after lambda lifting:

```
def v_6001 : ?[d₆_6000].([d₆_6000@(i64)]i32, i32 -> [d₆_6000@(i64)]i32) =
  (let (three_5990: i64) = 3@(i64)
   (let (tmp_7303: [three_5990@(i64)]i32) =
      replicate_7311@((n_5656: i64) -> (x_5657: i32) -> *[n_5656@(i64)]i32) three_5990@(i64) 1@(i32)
    in (tmp_7303@([three_5990@(i64)]i32), lifted_lambda_7313@((three_5990: i64) -> (i_5995: i32) -> *[three_5990@(i64)]i32) three_5990@(i64)))
   @(([d₆_6000@(i64)]i32, (i_5995: i32) -> *[d₆_6000@(i64)]i32), [d₆_6000]))
  @(([d₆_6000@(i64)]i32, (i_5995: i32) -> *[d₆_6000@(i64)]i32), [d₆_6000])
```

And after defunctionalisation:

```
def v_6001 : ?[d₆_6000].([d₆_6000@(i64)]i32, {three_5990: i64}) =
  (let (three_5990: i64) = 3@(i64)
   (let (tmp_7303: [three_5990@(i64)]i32) =
      replicate_7311@(i64 -> i32 -> *[n_7322@(i64)]i32) three_5990@(i64) 1@(i32)
    in (tmp_7303@([three_5990@(i64)]i32), dyn_lifted_lambda_7324@(i64 -> {three_5990: i64}) three_5990@(i64)))
   @(([three_5990@(i64)]i32, {three_5990: i64}), [d₆_6000]))
  @(([three_5990@(i64)]i32, {three_5990: i64}), [d₆_6000])
```

Note how it screwed up the types.  Darn.  This one is on me, although I may not have time to hack until the evening.

catvayor:
defunctionalisation should not see expression in sizes as it's after monomorphisation, maybe this bug ultimately comes from there ?

athas:
I fixed the bug in defunctionalisation.  tricky.fut works now.

athas:
I think the error in tests/higher-order-functions/shape-params6.fut is also due to defunctionalisation.  Will investigate.

athas:
Or maybe not.  Monomorphisation produces a function that looks like this:

```
def f_6619 [d<{(+_494) n_5985 m_5986}>_6614] (n_5985: i64) (m_5986: i64) ((f_5990: [d<{(+_494) n_5985 m_5986}>_6614@(i64)]i32 -> i64)) ((a_5991: [n_5985@(i64)]i32)) ((b_5992: [m_5986@(i64)]i32)) : i64 =
  let (tmp_6609: [d<{(+_494) n_5985 m_5986}>_6614@(i64)]i32) =
    (++_6618@((xs_5600: [n_5985@(i64)]i32) -> (ys_5601: [m_5986@(i64)]i32) -> *[d<{(+_494) n_5985 m_5986}>_6614@(i64)]i32)) a_5991@([n_5985@(i64)]i32) b_5992@([m_5986@(i64)]i32)
  in f_5990@([d<{(+_494) n_5985 m_5986}>_6614@(i64)]i32 -> i64) (tmp_6609@([d<{(+_494) n_5985 m_5986}>_6614@(i64)]i32))
```

The problem here is that `d<{(+_494) n_5985 m_5986}>_6614` is a size parameter that is only used nonconstructively in the parameters.  I can't explain why that leads to the issue we see, but it does seem off.

After normalisation the function looks like this:

```
def f_5999 [n_5985] [m_5986] ((f_5990: [n_5985@(i64) +_494 m_5986@(i64)]i32 -> i64): [n_5985@(i64) +_494 m_5986@(i64)]i32_483 -> i64_484) ((a_5991: [n_5985@(i64)]i32): [n_5985@(i64)]i32_483) ((b_5992: [m_5986@(i64)]i32): [m_5986@(i64)]i32_483) : i64 =
  let (tmp_6609: [n_5985@(i64) +_494 m_5986@(i64)]i32) =
    a_5991@([n_5985@(i64)]i32) ++_5610 b_5992@([m_5986@(i64)]i32)
  in f_5990@([n_5985@(i64) +_494 m_5986@(i64)]i32 -> i64) (tmp_6609@([n_5985@(i64) +_494 m_5986@(i64)]i32))
```

Which is correct.

catvayor:
found the culprit:
the partition of explicit/implicit params is done by looking if the size is in `mustBeExplicitInBinding bind_t'`
however, the sizes in bind_t' are not normalised, so `d<{(+_494) n_5985 m_5986}>_6614` isn't in it

athas:
Good catch.  I actually had another though (dangerous, avoid if possible, any consequences are not meant for this PR).

Now that we have full normalisation early, maybe we should also turn all size parameters into ordinary parameters.  And maybe we should not turn size expressions into variables until a bit later.

catvayor:
don't ask me why, but by cleaning the attributes propagation to solve assertion within unsafe, I solved the noinline's problems

catvayor:
If doing this, I think we should convert all the sizes at this moment, it would simplify questions about normalizing sizes but that's much more work on normalisation

athas:
> like map [n]'a'b : (a->b) -> [n]a -> [n]b to become map 'a'b : (n:i64) -> (a->b) -> [n]a -> [n]b at this stage ? isn't it the job of monomorphisation to infer this args ?

It is currently the job of monomorphisation, but monomorphisation doesn't actually do it for *all* of them (because previously the existential sizes might not be explicitly bound - they are now).  I don't have a strong vision for *where* it should be done, but it just seems sensible to get rid of as much implicit behaviour as possible (just like you're currently doing with full normalisation).

catvayor:
so the idea is to completely remove implicit size parameters ? I'll try to work on this when I'm done with desugaring

athas:
In some sense it could be seen as a form of monomorphisation: just like we remove type parameters, we could remove size parameters.

catvayor:
I was wondering, lambdas are sugar for LetFun aren't they ? it should be possible to replace every lambdas by some LetFun, is it a good idea ? I was initially wondering if LetFun were sugar for `let fun = \args -> body` but with the size parameter it looks not really possible...

athas:
That is done by the lambda lifting pass.  I don't have a strong opinion on whether it is a good idea to do it during normalisation.

catvayor:
++ case as other problems: the d₈ is consider as not in scope when transforming d₈+n, so an existential is introduced in the middle of the type of ++

catvayor:
this also cause the weird type of foo

catvayor:
Reordering some transformation will solve this second bug

catvayor:
the two last ++ really has different monotype:
- `(xs_5600: [?0]i64) -> (ys_5601: [?1]i64) -> *[?d<{(+_494) d₈_6005 n_5995}>_6623]i64`
- `(xs_5600: [?0]i64) -> (ys_5601: [?1]i64) -> *[?2]i64`
this comes from the fact that the environment making them appear known the size in advance, but not every time.
as we are not using sections, it can only comes from the `f () ++ iota n` in size, as the size of `f ()` isn't know, it's also the case for d+n

athas:
That is weird to me.  In `f () ++ iota n`, the size of the result *is* known, because full normalisation should first bind `f ()` to a name and make its size explicit.  That is, I'd expect the normalised expression to be something like:

```
let [k] (tmp0: [k]i64) = f ()
let (tmp1: [n]i64) = iota n
in (++) tmp0 tmp1
```

The result of the concatenation is `[k+n]i64`.

catvayor:
it is known for all desugared expression, that is all expression in function body, the problem is that expression in sizes are not desugared

catvayor:
avoid this would required that, after desugaring, the return of foo is
`[let [d] (tmp_0: [d]i64) = f() let (tmp_1: [n]i64) = iota n let (tmp_2: [d+n]i64) = (++) tmp_0 tmp_1 in foo tmp_2]i64`

catvayor:
In fact, we maybe can do this, remembering my naive work for alpha - equivalence (#1919) it was incorrect for typing, but maybe enough for monomorphisation

catvayor:
Currently, the full normalisation step only transform expression that are in the body, but not expressions inside sizes, which leads to the need of keeping some desugaring in monomorphisation step.
It can sounds harmless (and it probably is) as [sizes are evaluated a limited number of times](https://github.com/diku-dk/futhark/pull/1897#issuecomment-1503092166) but:
- It can cause some unneeded evaluation of some sub-expression
- It can break the assumption that there is only simple expression like in the output of normalisation

This two points can be seen in
```futhark
-- orig
def foo [n] [m] 't (_: [n+m]t) = n * m

def bar n (f : () -> []i64) =
    iota (foo (f () ++ iota n))

-- normalised
def foo [n] [m] 't ((_: [n + m]t): [n + m]t) : i64 =
  (*) n m

def bar (n: i64) ((f: () -> ?[d₁].[d₁]i64): () -> []i64) : *[foo (f () ++ iota n)]i64 =
  let [d₈] (tmp: [d₈]i64) =
    f ()
  let (tmp: [n]i64) =
    iota n
  let (tmp: [d₈ + n]i64) =
    (++) tmp tmp
  let (tmp: i64) =
    foo tmp
  in iota tmp

-- monomorphised
def foo [d<{(+) n m}>] (n: i64) (m: i64) ((_: [d<{(+) n m}>]i64)) : i64 =
  (*) n m

def bar (n: i64) ((f: () -> ?[d₁].[d₁]i64)) : ?[d<{foo d₈ n ((++) (f ()) (iota n))}>].*[d<{foo d₈ n ((++) (f ()) (iota n))}>]i64 =
  let [d₈] (tmp: [d₈]i64) =
    f ()
  let (d<{foo d₈ n ((++) (f ()) (iota n))}>: i64) =
    foo d₈ n ((++) (f ()) (iota n))
  let (d<{(+) d₈ n}>: i64) =
    (+) d₈ n
  let (tmp: [n]i64) =
    iota n
  let (tmp: [d<{(+) d₈ n}>]i64) =
    (++) tmp tmp
  let (tmp: i64) =
    foo d₈ n tmp
  in iota tmp

entry main (n: i64) : ?[d₄].*[d₄]i64 =
  let f {}: *[2]i64 =
    iota 2
  in bar n f
```

The expression `foo (f () ++ iota n)` is evaluated twice: as a size with nested function call and all the long of tmp's, the last tmp being this expression.
I don't think it's very problematic as, the return size of `bar` is well understood by monomorphisation, but it could be better,
and the fact that the size is not calculated in main probably means that we can have problems when using the result of bar.

I would like to apply normalisation process to sizes, but it will imply some modification to monomorphisation for it to work well as normalisation will introduce let-bindings in it:
- Eq of Monomorphise.ReplacedExp will need to be done on the rebuilt expression (e.g. replace var of let-bindings by their original expression) to not have alpha-equivalence problems (still need some thinking for introduced lambdas).
- When introducing the size variable, the normalised form should be introduced, with all the let's flattened.
- Identify identical let-bindings to remove the duplicate, at least the one introduced by the size (does `f () ++ f ()` should evaluate `f ()` twice ? at first I would say no, but what if `(++) : *[n]t -> *[m]t -> *[n+m]t`).
- Detect that in fact the d₈ is bounded inside the expression for the type of `bar n f` to be calculated.

catvayor:
also, for 
```futhark
def partition [n] 'a (p: a -> bool) (as: [n]a): ?[k].([k]a, [n-k]a) =
  let p' x = if p x then 0 else 1
  let (as', is) = intrinsics.partition 2 p' as
  in (take is[0] as', drop is[0] as')
```
it would be cool if the typing detect that the return type is `([d]t, [n-d]t` without helping it with a let-bind

catvayor:
now the only reason the program
```futhark
map (filter (>0) xs ++) mat
```
is refused is causality checking, but I think we can allow it now, thanks to normalisation

athas:
Yes, that seems harmless.  Causality checking is just a bunch of hardcoded sequence points - as long as it matches the binding order in full normalisation, we'll be fine.  I will make the change.

PR #1928, opened 2023-05-01
Correctly record write sets for BasicOps not otherwise handled

### LINKED ISSUE #1927: Problem with short circuiting
Consider this program:

```Futhark
def main k =
  let src = iota k
  let dst = iota 5
  in dst with [1:4] = src
```

Short-circuiting will put the array `src` in the same memory as `dst`, but *before constructing dst*.  This means that the construction of `dst` will overwrite `src`.

Worse, if `k` is too large, we will write out-of-bounds, as short-circuiting does not ensure that the bounds check is done before `src` is constructed.

This is a pretty bad bug that can result in anything from invalid results to memory corruption.

Munksgaard:
Previously, some BasicOps, like Iota didn't have their write sets correctly computed, which caused our short-circuiting pass to incorrectly short-circuit some programs involving iota.

Fixes #1927


Munksgaard:
> This breaks a bunch of structure tests for short-circuiting. Were those always unsound?

No, my fix was wrong. Hopefully the new version will unbreak things.



> Also, does this address the Certificates aspect of #1927?

No, but I'm working on that too.

PR #1932, opened 2023-05-02
When short-circuiting, source should inherit certificates of dest

### LINKED ISSUE #1930: Memory short circuiting ignores certificates
This program causes memory errors because the short-circuiting moves the write to `dst` (or rather its memory block) before the assertion implementing the bounds check:

```Futhark
-- ==
-- input { 5000i64 [1i64] }
-- error:

def main [n] (k: i64) (dst: *[n]i64) =
  let src = iota k
  in dst with [1:4] = src
```

I think @Munksgaard volunteered to fix it.

Munksgaard:
This commit fixes the issue described in #1930 where short circuiting without also applying the certificates of the destination to the source can cause runtime errors and memory corruption.  In addition to propagating the certificates, we also need to check that the values used in the certificate are actually in scope.

Fixes #1930


Munksgaard:
@athas: Regarding your comment in #1930: Yes, I was able to reuse some of the machinery, but as you can see here, additional handling was needed to pass certificates around to the places where we need them (checking whether the values inside are in scope).

Munksgaard:
I think most of these tests are failing because asserts are not lifted in LiftAllocations. @athas, do you want to take a stab at that?

athas:
Sure.

athas:
Can this not be done simply by adding a case for `Assert` [here](https://github.com/diku-dk/futhark/blob/99956f6eb83e339114cef1f1ca80cd6942327a70/src/Futhark/Pass/LiftAllocations.hs#L87)?

athas:
Surely those `<<loop>>` errors are not due to not moving the Asserts around?  Those indicate infinite loops in Haskell.

athas:
I think I fixed that part.  @Munksgaard could you verify that it was just a typo and that I fixed it?

PR #1955, opened 2023-05-29
Add Foldable and Traversable instances

Topsii:
Adresses https://github.com/haskell/core-libraries-committee/issues/93


athas:
Isn't that proposal abandoned?  This change is harmless, but is there a reason for it?

PR #1969, opened 2023-06-19
Remove Rotate from core language.

athas:
Still supported in the source language through a fairly trivial definition.


athas:
I'm not sure what the performance impact of this will be.  We do have a small handful of rotate simplification rules, but I think they may also apply to `map`s with indexing.

PR #1977, opened 2023-07-03
Type-check aliasing in separate pass

### LINKED ISSUE #1872: Do alias/consumption checking in separate type checking pass
To simplify the type checker, we should put the alias/consumption checking in a pass after we have inferred the structure of types.  It's very difficult to track aliases before we know the types fully, and we've had plenty of related bugs.  This would also simplify the type checker, and in particular make it easier to move to a constraint-based model.

athas:
This significantly simplifies the type checker, and also fixes known
(and future) bugs related to tracking aliases of an incompletely typed
term.

This also allows us to provide better error messages, although it's
possible that they are currently slightly worse, as the aliasing
checking does not use the "context" mechanism of the term-level type
checker.

The AST itself is simplified because it no longer has to embed alias
information (which wasn't correct anyway, and was not actually used).

There is one user-visible change: lambdas that *must* return an
alias-free result will now require a return type annotation.  We may
loosen this in the future, but it is a very rare case in practice.

Closes #1872.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   multidimensional variants), as well as `split`, now have more
   precise types.
 
+* Local and anonymous (lambda) functions that *must* return unique
+  results (because they are passed to a higher order function that
+  requires this) must now have an explicit return type ascription that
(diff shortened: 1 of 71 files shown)


athas:
One piece of functionality is missing: all variables are now reported as unused, because the alias checking machinery did double duty as a usage checker.  That'll need its own mechanism now.

PR #1981, opened 2023-07-06
Remove the notion of an index function "base".

athas:
We don't actually need this anymore.  It's a relic from a time where we were much more afraid of constructing new index functions.


athas:
Unfortunately, this reveals a bug in short-circuiting that was previously covered up by the many redundant copies inserted by allocation (with this change, we insert far fewer).  The bug can be demonstrated by this program:

```Futhark
def tridagSeq [m] (y: *[m]f32) (beta: f32) =
    loop y
      for i < m do
        let i    = i + 1
        let y[i] = y[i] - beta*y[i-1]
        in  y

def main [m][n] (u: *[n][m]f32) (betas: []f32) =
  map (\beta -> map (\u_row  -> tridagSeq (copy u_row) beta) u) betas
```

The intent is to give each thread a copy of `u_row`.  Unfortunately, when compiled with the GPU backends, short circuiting will short-circuit the unique parameter `u` with the `u` copies, meaning all threads will store their copy in the same memory.  Not good.

@Munksgaard has offered to take a look, but I will probably also stare at this a bit.

athas:
Notes on remaining failures:

* tests/soacs/scan2.fut fails because of memory expansion in a branch, which does not allocate enough memory.
* tests/intragroup/expansion0.fut fails because of double buffering, which does not allocate enough memory.

A common factor is that on `master`, the arrays involved all have a base whose rank is different from the array rank (i.e. the arrays are indexing results), which causes the allocator to copy them into a zero-offset fresh array before returning them from a branch or passing them to a loop.  This is clearly the right choice in these circumstances, and I wonder if we could bake this logic into the allocator in some other way.  I would really like to get rid of `IxFun.base`.

athas:
I still would like to get rid of these bases, but it's probably not urgent.  It also turns out we use it as a rather crucial source of information for memory expansion.  We can get rid of it when we redesign memory expansion.

PR #1990, opened 2023-07-20
Add new simplification rule for invariant loop parameters.

athas:
This was suggested by Cosmin to address some of the code produced by AD.

The idea is that for a loop of the form

```
loop p = x ...
  ...stms...
  in res
```

we construct and simplify the body

```
  let p = x
  ...stms...
  in res
```

and if that simplifies to 'x', then we conclude that the loop parameter 'p' must be invariant to the loop and simply bind it (and the loop result) to 'x'.

Complication: for multi-parameter loops, we must also check that the *original* computation of 'res' does *only* depends on other invariant loop parameters.

Currently we do this only for loops that have a constant as one of their initial loop parameter values.

The main downside of this rule is that doing recursive simplification is quite expensive.  Especially after sequentialisation, pretty much every 'reduce' will have been turned into a loop that triggers this rule (although the rule itself will fail in most cases, after doing the simplification).  Therefore I'm a bit hesitant to enable it as is. Sure, the Futhark compiler is slow and it was never meant to be fast, but it is still quite easy for the compiler to become *uselessly slow* if we are not careful.  E.g. on OptionPricing, this rule itself makes compilation 10% slower (and does not actually optimise anything - this is purely the cost of failing checks).

### DIFF
--- a/src/Futhark/Optimise/Simplify/Engine.hs
+++ b/src/Futhark/Optimise/Simplify/Engine.hs
@@ -54,7 +54,6 @@ module Futhark.Optimise.Simplify.Engine
     bindLParams,
     simplifyBody,
     ST.SymbolTable,
-    hoistStms,
     blockIf,
     blockMigrated,
     enterLoop,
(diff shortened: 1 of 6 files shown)


athas:
@coancea You're very welcome to suggest ways to speed this up.

athas:
We can bring this back in the future if we need it.

PR #2006, opened 2023-08-09
Track data dependencies in SOACs

nhey:


nhey:
That works well. Do you have any clue as to why unused operations (adds) are eliminated for map and reduce tests, but not for the scan test? I'm printing the dependencies at various places and they appear to be similar for map and scan.

The tests are in `tests/dependence-analysis/{map0,scan0,reduce0}.fut`.

src/Futhark/IR/SOACS/SOAC.hs
```diff
@@ -589,6 +593,30 @@ instance CanBeAliased SOAC where
...
+    let depsOfScan (Scan lam nes, deps_in) = depsOf' lam nes deps_in
+        depsOfRed (Reduce _ lam nes, deps_in) = depsOf' lam nes deps_in
+        depsOf' lam nes deps_in =
+          let deps_nes = map (depsOf mempty) nes
+          in lambdaDependencies mempty lam (zipWith (<>) deps_nes deps_in)
+
+        deps_map = print "deps_map" $
+          lambdaDependencies mempty map_lam (map oneName arrs)
+
+        (deps_scans_in', deps_map') = splitAt (scanResults scans) deps_map
```
athas, line 606: I'd suggest something like `splitAt3 (scanResults scans) (redResults reds) deps_map`.

athas:
The operator in the scan test is more complicated than the one for the reduce test.  Are you sure it works for reductions?

athas:
Oh, but the actual reason is that the removal of dead reduction results is by the dedicated rule `removeDeadReduction`.  There is no similar rule for scans.  It would have to be added (and would be very similar).

nhey:
https://github.com/diku-dk/futhark/blob/a2de95177077391afe3c810de4e8afca2a37fd2f/src/Futhark/IR/SegOp.hs#L1004

I'd like to implement `opDependencies` for the above to make the CI progress to tests. Would it be correct to simply apply `opDependencies` to the `SegOp` kernel body? (I have no idea what I'm doing.)

athas:
To a first approximation, that should be fine.

nhey:
Should JVP and VJP be handled in `opDependencies`? Otherwise it's ready for review.

athas:
Yes, they should.  I'll review this PR this week.

athas (approved):

athas:
I will also take a look.  It's possible that your change is not a fault, but simply reveals a bug in some other simplification.

nhey:
Ok, I'll leave it for now. The dependencies are as I would expect for the standard and multicore pipelines. I couldn't produce any differences between main branch and this branch by adding tests to the program. The IR after the standard pipeline is different, but looks correct to me. (However, I guess the error occurs in one of the simplify passes in the multicure/ispc/wasm pipelines, so that may be irrelevant.)

athas:
I also cannot reproduce locally.  Hypothesis: the error only occurs after you merge `master` into this branch (which GitHub will do before running CI).

athas:
I found a fix for the real bug this uncovered, but the reason it happens is that the data dependency analysis is now smart(?) enough that we simplify away the kind of manual double buffering that Cosmin is fond of.  I guess he'll find out the hard way!

PR #2012, opened 2023-08-25
Add error checking for array dimensions when using interpreter fix #1682

### LINKED ISSUE #1682: Internal compiler error when supplying incorrct array size when using `futhark run`
I have the following program [here](https://github.com/Alexnortung/futhark-nn/blob/ef87201d5eb29cee1945c9c87322d2bc261f3c3f/layers/convolutional.fut#L255-L259). When running the following command I get an internal compiler error

```
echo "[[[1.0, 2.0], [3.0, 4.0]], [[5.0, 6.0], [7.0, 8.0]]]
[2.0, 3.0]
[[[10.0, 2.0],[-1.0, -2.0]],[[20.0, 4.0], [-3.0, -4.0]]]
" | futhark run layers/convolutional.fut -e conv_layer_channels
```

The error I get is

```
src/Language/Futhark/Interpreter.hs:1918:13-63: Non-exhaustive patterns in (ShapeDim n
                                                         (ShapeDim m shape),
                                                xs')
```

In the previous command I use a 3 dimensional array, but the program expects a 4 dimensional array, so the following command works

```
echo "[[[[1.0, 2.0], [3.0, 4.0]], [[5.0, 6.0], [7.0, 8.0]]]]
[2.0, 3.0]
[[[10.0, 2.0],[-1.0, -2.0]],[[20.0, 4.0], [-3.0, -4.0]]]
" | futhark run layers/convolutional.fut -e conv_layer_channels
```

dominicmkennedy:
This code checks for mismatches in the number of dimensions as well as checking that the length of each dimension is the same. If there is a mismatch a helpful error message is printed.

### DIFF
--- a/src/Language/Futhark/Interpreter.hs
+++ b/src/Language/Futhark/Interpreter.hs
@@ -2028,12 +2028,18 @@ interpretFunction ctx fname vs = do
     updateType _ t =
       Right t
 
-    -- FIXME: we don't check array sizes.
     checkInput :: ValueType -> StructType -> Either T.Text ()
     checkInput (Scalar (Prim vt)) (Scalar (Prim pt))
       | vt /= pt = badPrim vt pt
     checkInput (Array _ _ (Prim vt)) (Array _ _ (Prim pt))
       | vt /= pt = badPrim vt pt
+    checkInput vArr@(Array _ (F.Shape vd) _) pArr@(Array _ (F.Shape pd) _)
+      | length vd /= length pd = badDim vArr pArr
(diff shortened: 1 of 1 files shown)


athas:
Good fix, but CI complains about a minor style issue.  You can use `tools/style-check.sh` to run this locally.

(Don't worry about the CUDA test failure - that's a hardware error on the testing system.)

dominicmkennedy:
oops, my bad. I just made a new commit that fixes the style. Thanks for the feedback

Issue #1567, opened 2022-01-17
Add expression attribute for repl display


leonardschneider:
I like the Futhark repl, but I'm missing formatting while printing the values. I propose adding some basic formatting, which can come handy while debugging a program, by adding expression attributes that will be interpreted by the repl to display expressions.

Here are some examples I'd like to have.

```
> #[char] 97u8 -- display a character
'a'
> #[utf8] "Hello, world!" -- display a string in utf8 encoding
"Hello, world"
> #[raw] "Hello, world!" -- direct display for extensibility
Hello, world! 
```

athas:
I don't think an expression is the right way to do this.  What about a `:string` command that prints the result of the expression as a string literal (if it has an appropriate type)?

FluxusMagna:
That could probably also be useful in futhark literate.

athas:
Yeah, now that I think about it, maybe a general `:format` command that accepts a `printf`-style format string would be better.  The same machinery could be used for `futhark literate`.

leonardschneider:
You're right. I've played with a prototype and a command makes more sense.

athas:
Instead of `printf` style, I think the most elegant option would be something similar to [Pythons formatted string literals](https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals).  They are similar to `printf` strings, but instead of being positional, you put the name (or eventually, expression) of what you want to print:

```
> let str = "foo"
> :print "{str}"
[102u8, 111u8, 111u8]
> :print "{str:s}"
"foo"
```

With no specifier it will print in the default manner for the type, but with a specifier we can force e.g. arrays to be interpreted as UTF-8 strings.

PR #2020, opened 2023-09-16
Format command for REPL #1567

dominicmkennedy:
I have added the machinery to print strings with arbitrary expressions within the Futhark REPL. I have yet to add any options for formatting, my first thought is to add formatting code to `Util.Pretty`, but I am open to advice

### DIFF
--- a/src/Futhark/CLI/REPL.hs
+++ b/src/Futhark/CLI/REPL.hs
@@ -3,12 +3,12 @@
 -- | @futhark repl@
 module Futhark.CLI.REPL (main) where
 
-import Control.Arrow (ArrowChoice (left))
 import Control.Exception
 import Control.Monad
 import Control.Monad.Except
 import Control.Monad.Free.Church
 import Control.Monad.State
+import Data.Bifunctor (first)
 import Data.Char
 import Data.List (intersperse)
 import Data.List.NonEmpty qualified as NE
@@ -417,8 +417,8 @@ typeCommand = genTypeCommand parseExp T.checkExp $ \(ps, e) ->
 mtypeCommand :: Command
 mtypeCommand = genTypeCommand parseModExp T.checkModExp $ pretty . fst
 
-parseForPrint :: T.Text -> Either T.Text ([T.Text], [T.Text])
-parseForPrint printStr
+parseForFormat :: T.Text -> Either T.Text ([T.Text], [T.Text])
+parseForFormat printStr
   | not balanced =
       Left
         "Parse error on print string. Likely due to mismatched braces {}."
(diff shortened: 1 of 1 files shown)


src/Futhark/CLI/REPL.hs
```diff
@@ -405,6 +417,41 @@ typeCommand = genTypeCommand parseExp T.checkExp $ \(ps, e) ->
...
+printCommand :: Command
+printCommand input = do
+  case parseForPrint input of
+    Left err -> liftIO $ T.putStrLn err
+    Right (strs, exps) -> do
+      prompt <- getPrompt
+      case partitionEithers $ map (parseExp prompt) exps of
+        ((SyntaxError _ err) : _, _) -> liftIO $ T.putStr err
+        (_, uncheckedExps) -> do
+          vals <- mapM onExp uncheckedExps
```
athas, line 444: I have a hunch that continuing to evaluate later expressions, after an early one has failed, would be confusing.  I suggest stopping at the first error.

athas:
Neat, this is a good start for functionality we've wanted for a while.

* I think using Python-style formatting is the right call.  We don't need to implement [every feature](https://docs.python.org/3/library/string.html#formatspec) right now, but we should not go down a path that prevents us from adding features later.  Supporting brace-based interpolation, without anything else, is a pretty good way to begin, and useful on its own.

* Util.Pretty is supposed to extend (and wrap) the prettyprinting library.  I don't think it's a good fit for parsing format strings.  I don't have a strong vision for where else to do it.  Perhaps a new module is best, since we also want to use this functionality in `futhark literate`.  Maybe call it Futhark.Format?

* Parsing format strings should eventually be done in a slightly smarter way, so we can put the right source locations in the expressions.  But what you have will suffice for now.

src/Futhark/CLI/REPL.hs
dominicmkennedy: That sounds like a good idea, I added another commit resolving that, and also providing a *slightly* better error message if there is an error in an expression.

dominicmkennedy:
Putting all of the formatting into it's own module seems like a good idea. If it's alright with you I'm going to mark this PR as ready to merge, but I'd be willing to help out with a Futhark.Format module.

src/Futhark/CLI/REPL.hs
```diff
@@ -483,6 +533,15 @@ Only one source file can be loaded at a time.  Using the :load command a
 second time will replace the previously loaded file.  It will also replace
 any declarations entered at the REPL.
 
+|]
+      )
+    ),
+    ( "print",
```
athas, line 539: Call the command "format" instead.

src/Futhark/CLI/REPL.hs
```diff
@@ -3,6 +3,7 @@
 -- | @futhark repl@
 module Futhark.CLI.REPL (main) where
 
+import Control.Arrow (ArrowChoice (left))
```
athas, line 6: I think you can use `Data.Bifunctor.first` instead of this (and that is the common style in the compiler).

athas (changes requested):

src/Futhark/CLI/REPL.hs
dominicmkennedy: And done

PR #2029, opened 2023-10-08
Fix issue of failed run 'make install' in macOS

flyfish30:
The command 'make install' should call install command with option '-D', but in macOS, the install command hasn't option '-D'. So, I add a condition at install target in Makefile to detect operate system, if the operate system is macOS, then call install command with option '-d' to create installed directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@
 
 PREFIX?=$(HOME)/.local
 
+UNAME:=$(shell uname)
+
 # Disable all implicit rules.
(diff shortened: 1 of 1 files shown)


athas:
Would the macOS version also work on Linux?

flyfish30:
> Would the macOS version also work on Linux?

I build the macOS version and successfully run on macOS. In general, the macOS version of the executable program can't run on Linux.

PR #2031, opened 2023-10-12
Interchange histogram with inner maps during reverse AD

nhey:
Implementation of the ISWIM/IRWIM rewrite rule for reverse mode AD on histograms. Programs like the following enjoy significant speed-ups.
```
-- ==
-- entry: vecmul vecmin
-- compiled random input { [10000000]i64 [10000000][30]f32 [50000][30]f32 }

entry vecmul [n][d][bins] (is: [n]i64) (vs: [n][d]f32) (adj_out: [bins][d]f32) =
  vjp (hist (map2 (*)) (replicate d 1f32) bins is) vs adj_out

entry vecmin [n][d][bins] (is: [n]i64) (vs: [n][d]f32) (adj_out: [bins][d]f32) =
  vjp (hist (map2 f32.min) (replicate d f32.highest) bins is) vs adj_out
```


nhey:
Tests also fail locally; I must've run them with the wrong compiler before creating this PR.

PR #2033, opened 2023-10-18
futhark profile fixup

nhey:
Minor fixes related to multiple entry points/programs that I noticed while trying out the new profiling CLI.

### DIFF
--- a/docs/man/futhark-profile.rst
+++ b/docs/man/futhark-profile.rst
@@ -38,7 +38,7 @@ JSON file ``prog.json``, this will create a *top level directory*
 ``prog.prof`` that contains files with human-readable profiling
 information.  A set of files will be created for each benchmark
 dataset.  If the original invocation of ``futhark bench`` included
-multiple programs, then ``futhark prof`` will create subdirectories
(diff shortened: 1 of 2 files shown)


athas:
Fair enough with the typo, but what is the problem with colons in filenames?  Are subdirectories more convenient?  I'm fine if you think it's easier to navigate; I haven't yet used this with a program with many entry points.

nhey:
Removing them was just an easy fix to the fact that `dropExtension "prog.fut:entry"` drops `.fut:entry`. I can change it to preserve colons somehow.

athas:
No, maybe these subdirectories are for the best.

nhey:
On second thought, `dropExtension` shouldn't even make sense after applying `encode`, hm.

athas:
Right, that's bad.  But I do think subdirectories for entry points is better.  I have seen some programs with a dozen entry points, and if you have several data sets for each, then a single directory becomes very unwieldy.  Although you'll have to write some code that splits off the entry point at the end.

nhey:
The above should do it.

PR #2037, opened 2023-10-31
Fully parallelise multi-dimensional scatters.

### LINKED ISSUE #2035: Incomplete parallelisation of top-level multidimensional scatters
```Futhark
entry main (xss: *[][]i32) (is: []i64) (ys: [][]i32) =
  scatter xss is ys
```

It works when the `scatter` is nested.  I hope we can fix it without duplicating too much logic.

athas:
Closes #2035.


FluxusMagna:
Much appreciated!

PR #2041, opened 2023-11-03
single pass scan group virtualization

sortraev:
This PR adds group virtualization to SegScan.SinglePass. The generated single pass scan code now respects the suggested/requested num_groups/block_size, which in turn allows scanomaps with array construction in the map KernelBody, since mem expansion bases its expansion on this information.

Also
* status flags (for the lookback step) are now initialized inside the kernel
* some light code refactoring

### DIFF
--- a/src/Futhark/CodeGen/ImpGen/GPU/SegScan.hs
+++ b/src/Futhark/CodeGen/ImpGen/GPU/SegScan.hs
@@ -51,8 +51,8 @@ bodyHas f = any (f' . stmExp) . bodyStms
 canBeSinglePass :: [SegBinOp GPUMem] -> Maybe (SegBinOp GPUMem)
 canBeSinglePass scan_ops =
   if all ok scan_ops
-     then Just $ combineScanOps scan_ops
-     else Nothing
+    then Just $ combineScanOps scan_ops
+    else Nothing
   where
     ok op =
       segBinOpShape op == mempty
@@ -80,9 +80,9 @@ compileSegScan pat lvl space scan_ops map_kbody =
         SinglePass.compileSegScan pat lvl space scan_ops' map_kbody
       _ ->
         TwoPass.compileSegScan pat lvl space scan_ops map_kbody
+    emit $ Imp.DebugPrint "" Nothing
   where
     n = product $ map pe64 $ segSpaceDims space
-    targetSupportsSinglePass HIP  = True
     targetSupportsSinglePass CUDA = True
-    targetSupportsSinglePass _    = False
-
+    targetSupportsSinglePass HIP = True
+    targetSupportsSinglePass _ = False
--- a/src/Futhark/CodeGen/ImpGen/GPU/SegScan.hs
+++ b/src/Futhark/CodeGen/ImpGen/GPU/SegScan.hs
@@ -14,8 +14,8 @@ import Futhark.IR.GPUMem
 
 -- The single-pass scan does not support multiple operators, so jam
 -- them together here.
-combineScans :: [SegBinOp GPUMem] -> SegBinOp GPUMem
-combineScans ops =
+combineScanOps :: [SegBinOp GPUMem] -> SegBinOp GPUMem
+combineScanOps ops =
   SegBinOp
     { segBinOpComm = mconcat (map segBinOpComm ops),
       segBinOpLambda = lam',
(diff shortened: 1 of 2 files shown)


src/Futhark/CodeGen/ImpGen/GPU/SegScan.hs
```diff
@@ -82,17 +70,19 @@ compileSegScan ::
...
-  target <- hostTarget <$> askEnv
-  case target of
-    CUDA
-      | Just scan' <- canBeSinglePass scans kbody ->
-          SinglePass.compileSegScan pat lvl space scan' kbody
-    HIP
-      | Just scan' <- canBeSinglePass scans kbody ->
-          SinglePass.compileSegScan pat lvl space scan' kbody
-    _ -> TwoPass.compileSegScan pat lvl space scans kbody
-  emit $ Imp.DebugPrint "" Nothing
```
athas, line 96: But back this empty-line debug print.  It makes the output easier to read.

src/Futhark/CodeGen/ImpGen/GPU/SegScan/SinglePass.hs
```diff
@@ -219,431 +227,447 @@ compileSegScan ::
...
 
-  num_threads <-
-    dPrimVE "num_threads" $ num_groups' * group_size'
+  num_virtgroups <-
+    tvSize <$> dPrimV "num_virtgroups" (n `divUp` (group_size_e * chunk))
+  let num_virtgroups_e = pe64 num_virtgroups
+
+  -- TODO: what is num_threads, and should it be dependent on number of physical
+  --       or virtual groups?
+  num_threads <- dPrimVE "num_threads" $ num_virtgroups_e * group_size_e
```
athas, line 265: I think `num_threads` (as you have defined it here), is the *virtual* number of threads.  I'd suggest finding a different work for this (worker?), because `thread` usually means the physical concept.  Or at least call it `num_virt_threads`.

src/Futhark/CodeGen/ImpGen/GPU/SegScan/SinglePass.hs
```diff
@@ -219,431 +227,447 @@ compileSegScan ::
...
+          sAllocArray
+            "private"
+            ty
+            (Shape [intConst Int64 chunk])
+            (ScalarSpace [intConst Int64 chunk] ty)
+
+      sComment "Load and map" $
+        sFor "i" chunk $ \i -> do
+          -- The map's input index
+          phys_tid <-
```
athas, line 354: This is not a *physical* thread ID, is it?  It's a virtual thread ID.

athas (changes requested):
Please also add a comment somewhere in the module (it can be in the header) outlining the virtualisation strategy.

athas:
There is a style issue, but worse, it also appears some of the scan benchmarks fail with the CUDA backend.  (There's also a bunch of other things that fail because we are moving servers - you'll have to pick out the interesting failures from the wreckage.)

sortraev:
> There is a style issue, but worse, it also appears some of the scan benchmarks fail with the CUDA backend. (There's also a bunch of other things that fail because we are moving servers - you'll have to pick out the interesting failures from the wreckage.)

Yes, I see now that the benchmarks also fail when I run manually on the A100. I think the problem had to do with our change of the status flags initialization (we simply changed it to align with Cosmin's and our own prototype), but I'm not exactly sure why -- it was not a problem on our own 4090.

Anyway, I have pushed a commit which passes the benchmarks when I run them manually on the A100, so hopefully the bug has been squished (assuming this was the bug I was looking for; admittedly I don't see why it would be a bug). I will look into the other change requests and CI failures at a later time -- thanks for the comments

src/Futhark/CodeGen/ImpGen/GPU/SegScan/SinglePass.hs
```diff
@@ -314,7 +316,11 @@ compileSegScan pat lvl space scan_op map_kbody = do
                 (untyped (1 :: Imp.TExp Int32))
           sComment "Set dynamic id and reset status flag for this block" $ do
             copyDWIMFix sharedId [0] (tvSize dyn_id) []
-            copyDWIMFix statusFlags [tvExp dyn_id] (intConst Int8 statusX) []
+            -- TODO: find out why this flag initialization (rather than
+            -- initializing using sReplicate outside the current kernel) is
+            -- causing problems. it was never a problem with any of our
+            -- prototypes.
+            -- copyDWIMFix statusFlags [tvExp dyn_id] (intConst Int8 statusX) []
```
athas, line 323: There is no guarantee any of the previous blocks have had a chance to set their status flags by the time other blocks start reading them, is there?  My intuition is certainly that it must be *fully* done before the kernel starts.

sortraev: Well, this is embarassing. Yes, your intuition is of course correct. It seems that I (and, in my cowardly defense, the other prototype on which we based experiments) falsely applied the same logic onto the status flags as for the global ID counter, which _can_ be initialized once and efficiently reset inside the kernel.

The race is very unlikely, since it requires a thread group to both fetch its dynamic ID `j`, set `status_flags[j] = flag_X`, compute an intra-group scan, publish its aggregate to global mem, and start its lookback phase -- all in the time it takes between some other group fetching its dynamic ID `i < j` (this is guaranteed to happen before group `j` fetches theirs) and setting `status_flags[i] = flag_X` in the very next line of code. But yes, obviously possible!

PR #2052, opened 2023-11-29
Fix #2051.

### LINKED ISSUE #2051: Incorrect result with GPU backends
```Futhark
-- ==
-- compiled random input { [512][32]i32 } auto output

def main [n][m] (xss: [n][m]i32): []i32 =
  reduce (\(xs: []i32) ys ->
               loop zs = replicate m 0 for i < m do
                 let zs[i] = xs[i] + ys[i]
                 in zs
         )
         (replicate m 0)
         xss
```

Looks to be an error in code generation for *non-commutative non-segmented reductions with array operands*.

athas:


athas:
This does not fix any program that any real human actually wants to write.  I really wish we had a better way of handling these absurd operators before we got to code generation, as the code is gnarly and not obviously correct.

Anyway @sortraev, this fixes the issue you found.

PR #2055, opened 2023-12-03
Remove in-place lowering.

athas:
This is one of our oldest optimisations, and mainly served to clean up the code produced by sequentialisation, as well as a few other simple things that we would today call "memory short circuiting".

In-place lowering is a really complicated optimisation, so now that we have a propert short circuiting pass, we should seriously consider getting rid of it. Short circuiting is even more complicated, but it is also properly principled, and more general.

We still largely have the sequentialisation problems that in-place lowering was supposed to address (mostly nested maps), but at least on the GPU backends there seems to be no real impact from it anymore. I would still like to one day fix sequentialisation so that it produces better code for nested SOACs.

The big open question is how this impacts the performance of CPU code.


Munksgaard:
35973341_10160626751145523_5408291598729478144_n

PR #2057, opened 2023-12-06
Query register capacity for segred and segscan codegen.

athas:
This is very tedious code, and required adding the notion of "kernel constant expressions", as we have some expressions that _must_ be constant at kernel compilation time (which is at program runtime). We actually had this notion in the ImpCode representation, but now ImpGen provides some manual control as well.


athas:
Looks ilke the accurate querying is actually *detrimental* to performance, compared to the old hardcoded values.

athas:
The old numbers correspond to pretending we have twice as much local memory available as is actually the case.  Should we just multiply `CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK` by two?  What is the logic here?  @coancea, do you remember?

Issue #2061, opened 2023-12-14
Type error after short circuiting


athas:
This concise little program fails if passed to `--short-circuiting-gpu`: http://sigkill.dk/junk/issue1669.fut_gpu_mem

(It is derived from issue1669.fut, but you need to disable some simplification rules to make it look like this.)

The error message is:

```
...
Type error:
Return type
  {mem@[3i64]f64,
   [3i64]f64 @ mem_13191 ->
   {base: [27i64, 3i64];
    LMAD: {offset: mul_nw64 (i_13046) (3i64);
           strides: [1i64];
           shape: [3i64]}}}
cannot match returns of results
  {mem@[27i64][3i64]f64,
   [3i64]f64 @ mem_13191 ->
   {base: [27i64, 3i64];
    LMAD: {offset: mul_nw64 (i_13046) (3i64);
           strides: [1i64];
           shape: [3i64]}}}
Expected mem@[3i64]f64 but got mem@[27i64][3i64]f64
```

Probably something gets confused about scalar memory somewhere.  I didn't think we tried to short-circuit those anyway.

athas:
Uh, do we actually check for compatibility of memory spaces anywhere?

PR #2064, opened 2023-12-18
Fix 2061

Munksgaard:


Munksgaard:
Do you have a test .fut file for this @athas? Or can I somehow test that short-circuiting succeeds on the .fut_gpu_mem from #2061?

athas:
You can test it manually, which I assume you did, but I wasn't able to come up with an automated test for this one.

PR #2126, opened 2024-03-26
Efficient Sequentialization - Masters

CKuke:
This is a pull request to merge in our compiler extension in the form of an additional compiler pass. This pass performs efficient sequentialization on intra-block kernels with an outer map at block level.

It was developed as part of our masters thesis at the University of Copenhagen in Computer Science with Cosmin as our main supervisor.


src/Futhark/Optimise/IntraSeq.hs
```diff
@@ -0,0 +1,1225 @@
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-name-shadowing #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
```
athas, line 4: Get rid of these (except -Wno-orphans if you have a really good reason for it.)

athas (changes requested):

athas:
I will merge this into a new `intraseq` branch so @coancea can work on it.

PR #2127, opened 2024-04-02
Support for Unsafe UpdateAcc

sortraev:
Adds support for unsafe `UpdateAcc` ops, to omit boundary checks on the accumulator in cases where the indices are known (or assumed) to always be within bounds. The initial motivation was to use `WithAcc` to read block/register tiles from global to shared memory in the (WIP) tensor contraction tiling, wherein I would like to handle boundaries manually (if at all).

In most places, I simply took inspiration from the existing `Update` cases.

TODOs:
1) In those places in the compiler where `UpdateAcc` ops are passed along,
   I simply pass on the existing `Safety` information. I cannot discern
   whether there may be cases where an optimization or transformation
   enables or necessitates changing the `Safety`, so I hope someone will
   look into this before merging.
2) In those places in e.g. AD where new `UpdateAcc`s are created, I used
   `Safe` safety to preserve behavior, but it may or may not be beneficial
   to substitute `Unsafe` in certain places.
3) I am not familiar with Megaparsec, so I simply copy+pasted the
   current rule for `UpdateAcc` in `IR.Parse` (and `IR.Pretty`).

### DIFF
--- a/src/Futhark/IR/Parse.hs
+++ b/src/Futhark/IR/Parse.hs
@@ -317,10 +317,13 @@ pBasicOp =
       ArrayLit
         <$> brackets (pSubExp `sepBy` pComma)
         <*> (lexeme ":" *> "[]" *> pType),
+      -- TODO: should combine these rules in one.
+      keyword "update_acc_unsafe"
+        *> parens
+          (UpdateAcc Unsafe <$> pVName <* pComma <*> pSubExps <* pComma <*> pSubExps),
       keyword "update_acc"
         *> parens
-          (UpdateAcc <$> pVName <* pComma <*> pSubExps <* pComma <*> pSubExps),
-      --
+          (UpdateAcc Safe <$> pVName <* pComma <*> pSubExps <* pComma <*> pSubExps),
       pConvOp "sext" SExt pIntType pIntType,
       pConvOp "zext" ZExt pIntType pIntType,
       pConvOp "fpconv" FPConv pFloatType pFloatType,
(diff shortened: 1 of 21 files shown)


athas:
Regarding point (1), since you only ever create `Safe` `UpdateAcc`s, propagation should maintain the current behaviour, which is therefore correct.

Using `Unsafe` where appropriate in AD can be done later.

src/Futhark/IR/Parse.hs
```diff
@@ -317,10 +317,13 @@ pBasicOp =
       ArrayLit
         <$> brackets (pSubExp `sepBy` pComma)
         <*> (lexeme ":" *> "[]" *> pType),
+      -- TODO: should combine these rules in one.
```
athas, line 320: Either do so, or remove the TODO. I don't mind having two different keywords.

src/Futhark/Optimise/GenRedOpt.hs
```diff
@@ -105,7 +105,8 @@ genRed2Tile2d env kerstm@(Let pat_ker aux (Op (SegOp (SegMap seg_thd seg_space k
     -- some `code1`, followed by one accumulation, followed by some `code2`
     -- UpdateAcc VName [SubExp] [SubExp]
     (code1, Just accum_stmt, code2) <- matchCodeAccumCode kstms,
-    Let pat_accum _aux_acc (BasicOp (UpdateAcc acc_nm acc_inds acc_vals)) <- accum_stmt,
+    -- TODO: should Unsafe UpdateAccs be treated differently?
```
athas, line 108: Ask Cosmin, but I think not.

athas (changes requested):

src/Futhark/Optimise/GenRedOpt.hs
sortraev: I would also think not. @coancea do you have an opinion?

athas:
Is this still WIP or can it be merged?

sortraev:
As far as I'm concerned, it is ready to merge, yes!

PR #2129, opened 2024-04-15
Allow blank lines and comments in tuning files

sortraev:
Perhaps there is a good reason this was not already featured -- in which case, feel free to reject -- but I find this nice-to-have convenient, rather than having to delete or move unused tuning parameters.

Can import `isspace` to catch all whitespace characters.


athas:
I suppose this is harmless.

Issue #1403, opened 2021-07-19
Add a WebGPU backend.


athas:
This is a big one.  Current plan is to generate C that calls [native WebGPU](https://github.com/webgpu-native), which [Emscripten can translate to WASM+WebGPU](https://github.com/emscripten-core/emscripten/pull/10218).  Two major challenges exist:

* We must generate shaders in the WebGPU Shader Language (WGSL), which is *completely* unrelated to C.  We cannot just shim it like we did with CUDA, but must write an entirely new code generator.
* The WebGPU host API is also very dissimilar to CUDA/OpenCL, and reminds me more of Vulkan.  Significant effort must be expended here.

Before we can even think about starting the implementation, we need to gain more WebGPU expertise.  The best starting point is to write some WebGPU compute shader programs, initially in straight JavaScript, and then in C translated with Emscripten.

On top of these significant tasks, WebGPU is also not a ratified standard and browser support is really shaky.  This is a technically very challenging project in all regards.

athas:
Some notes.

* Apart from (obviously) running in a browser, there are also "native" WebGPU backends.  Specifically wgpu (used to implement WebGPU in Firefox) and Dawn (used for Chrome).
* These implement a shared header [webgpu.h](https://github.com/webgpu-native/webgpu-headers/blob/main/webgpu.h), which is also handled by Emscripten.
* Wgpu is written in Rust, but they have [https://github.com/gfx-rs/wgpu-native](wgpu-native), which presents a C API.  I have tried it and it works. 
* There is [this simple example](https://github.com/gfx-rs/wgpu-native/blob/master/examples/compute/main.c) of compute shaders.  It does not work with Emscripten, because it uses the wgpu-specific function `wgpuDevicePoll` to wait until various async operations complete.  Dawn has a similar function.  [Apparently browsers normally hook this into their own event loop or something](https://github.com/webgpu-native/webgpu-headers/issues/91), but maybe webgpu.h will need to be extended for our purposes.

Apart from the blocking/polling issue, the WebGPU C API looks reasonably simple (although more complex than OpenCL and CUDA).  I don't think it will present major issues.  I have yet to closely study WGSL (the language for writing shaders/kernels), but I did notice that it is missing many important scalar types (integers other than 32-bit for example).

athas:
[...]

PR #2140, opened 2024-05-02
Add WebGPU backend

spaarmann:
WIP WebGPU backend (#1403)

The JS glue generation is still very basic and barely enough to implement the server protocol on top of. There's also a slightly weird mixture of generating JS vs just accessing the embedded JSON manifest that I'm still hoping to clean up a bit.


rts/python/browser_test.py
athas, line 1: This should go in `tools/`.

rts/python/browser_test.py
```diff
@@ -0,0 +1,303 @@
+#!/usr/bin/env python
```
athas, line 1: `python3` surely?

rts/wgsl/scalar.wgsl
athas, line 1: For consistency, I suggest moving the 32-bit operations into a file `scalar32.wgsl`.

src/Futhark/Actions.hs
```diff
@@ -224,6 +248,7 @@ runCC cpath outpath cflags_def ldflags = do
             ++ cmdCFLAGS cflags_def
             ++
             -- The default LDFLAGS are always added.
+            -- The default LDFLAGS are always added.
```
athas, line 251: Duplicate line.

src/Futhark/Actions.hs
```diff
@@ -260,6 +285,8 @@ runISPC ispcpath outpath cpath ispcextension ispc_flags cflags_def ldflags = do
...
+            -- The default LDFLAGS are always added.
```
athas, line 289: Two duplicate lines.

src/Futhark/CodeGen/Backends/CWebGPU.hs
```diff
@@ -0,0 +1,382 @@
...
+       #ifdef USE_DAWN
+         #include <dawn/webgpu.h>
+       #else
+         #include <webgpu/webgpu.h>
+         #include <emscripten.h>
+         #include <emscripten/html5_webgpu.h>
+       #endif
+      |]
+
+-- | As server script. Speaks custom server protocol to local Python server
```
athas, line 377: I suggest not using the term "server protocol" to reference anything but the "Futhark server protocol" (which should be understood as "(Futhark server) protocol". The term "server protocol" on its own is either wrong or redundant. Just call a  "custom protocol" or "browser test protocol" if it must have a name.

tools/extract-shader.sh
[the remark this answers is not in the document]
spaarmann: No, this was for very early testing, I'll remove it.

athas:
I pushed a commit that enables some more browser crud. With that change, it works on my home machine. (Probably not on my laptop, which I think has a GPU that is blocklisted by Chromium.)

spaarmann:
> I pushed a commit that enables some more browser crud.
I guarded the extra arguments to not apply when using a remote driver as heuristic for my WSL setup since it seems passing them to Windows Chrome is not allowed. I assume this is fine for your home machine, but we might need a better chance if we ever use the remote mechanism for anything else.

I also revamped the JS interface to be much more usable. I ended up making a new JS implementation of the binary data protocol. There were some problems with using the one for the WASM backend, it uses some node-specific things and isn't exactly set up to deal with some of the API functions suddenly being asynchronous. Didn't want to mess with it too much without having looked much at the WASM backend codegen etc. and the format is simple enough.

teddyjfpender:
Do you guys need any help moving this forward? Happy to help :)

athas:
You are very welcome to help out. Currently matters are a bit unstructured, I'm afraid.

* @TudorPal has been working on improving the documentation and testing situation, and in particular has developed some examples that are available here: https://github.com/diku-dk/futhark-webgpu
* A very useful contribution would be to look at those examples and see whether they make sense for you.
* Then, I suggest trying to use the backend for something, and figure out where it falls short.

WebGPU has limitations that mean some things simply cannot work. Eventually we should have good error messages (or maybe fallbacks) for those, but we don't have a good list of the current limitations.

PR #2153, opened 2024-05-30
Implement forward- and backward mode AD in the interpreter

vox9:
I'm solely pushing this for code review - I'd need to revise the commits, and clean up the code before I do a real pull request :)


src/Language/Futhark/Interpreter.hs
```diff
@@ -1921,11 +1984,90 @@ initialCtx =
...
+              ValueRecord m -> foldl1 (++) (M.map getSeeds m)
+              _ -> error "Not implemented (d19h45iu782)" -- TODO: Arrays?
+        let s' = getSeeds s
+
+        -- Impregnate the values
+        let addSeeds i v = case v of
+              ValuePrim v' -> (i + 1, ValueSeed $ AD.VjpSeed depth $ AD.TapeId i $ AD.Primal $ fromJust $ getV v')
+              ValueSeed v' -> (i + 1, ValueSeed $ AD.VjpSeed depth $ AD.TapeId i $ AD.Seed v')
+              ValueRecord m -> second ValueRecord $ M.mapAccum addSeeds i m
+              _ -> error "Not implemented (gs3g3ss)" -- TODO: Arrays?
```
athas, line 2004: In order to improve the way this is written (and to handle arrays), I suggest defining a utility function for performing a monadic traversal across the "leaves" of an interpreter value. Maybe something with this type:

```Haskell
traverseValue :: Monad f => (PrimValue -> f (Value m)) -> (AD.ADSeed -> f (Value m)) -> Value m -> f (Value m)
```

src/Language/Futhark/Interpreter.hs
```diff
@@ -1921,11 +1984,90 @@ initialCtx =
...
+
+        -- Derive the seeds
+        let deriveSeeds v = case v of
+              ValuePrim v' -> ValuePrim $ putV $ AD.valueAsType (fromJust $ getV v') 0
+              ValueSeed s@(AD.JvpSeed d _ d') ->
+                if d == depth then -- TODO: Fix add op
+                  adToValue d'
+                else ValuePrim $ putV $ AD.valueAsType (AD.value $ AD.primal s) 0
+              ValueRecord m -> ValueRecord $ M.map deriveSeeds m
+              _ -> error "Not implemented (d19hasd782)" -- TODO: Arrays?
```
athas, line 2066: I have a hunch that some functions (like this one) should be factored into their own top-level definitions, to make it easier to understand.

src/Language/Futhark/Interpreter/AD.hs
```diff
@@ -0,0 +1,289 @@
...
+import Futhark.AD.Derivatives
+import Futhark.Analysis.PrimExp
+import Language.Futhark.Core
+
+import Debug.Trace qualified as DBG
+import Control.Monad (zipWithM)
+import Data.Maybe (fromJust)
+import Data.Foldable (foldlM)
+
+data ADSeed
```
athas, line 38: Add a comment remarking on what the meaning of "seed" is (both here and in your implementation in general).

src/Language/Futhark/Interpreter/AD.hs
```diff
@@ -0,0 +1,289 @@
...
+  = VjpSeed Int Tape
+  | JvpSeed Int ADValue ADValue
+
+instance Show ADSeed where
+  show :: ADSeed -> String
```
athas, line 43: Remove this type signature.

src/Language/Futhark/Interpreter/AD.hs
```diff
@@ -0,0 +1,289 @@
...
+
+depth :: ADSeed -> Int
+depth (VjpSeed d _) = d
+depth (JvpSeed d _ _) = d
+
+primal :: ADSeed -> ADValue
+primal (VjpSeed _ v) = tapeValue v
+primal (JvpSeed _ v _) = v
+
+valueAsType :: PrimValue -> Rational -> PrimValue
```
athas, line 81: You only ever call this function with a zero value for the `Rational`. There is a notion of `blankPrimValue` that you could use instead.

src/Language/Futhark/Interpreter/AD.hs
```diff
@@ -0,0 +1,289 @@
...
+opReturnType (OpConv op) = snd $ convOpType op
+opReturnType (OpFn   fn) = fromJust $ do
+                            (_, t, _) <- M.lookup fn primFuns
+                            Just t
+
+addFor :: Op -> Op
+addFor op = OpBin $ fromJust $ case opReturnType op of
+  IntType t -> Just $ Add t OverflowUndef
+  FloatType t -> Just $ FAdd t
+  _ -> Nothing
```
athas, line 114: Instead of using `fromJust` and getting a bad error message, just call `error` directly with an explanation of which `op` failed.

src/Language/Futhark/Interpreter/AD.hs
```diff
@@ -0,0 +1,289 @@
...
+
+multiplyFor :: Op -> Op
+multiplyFor op = OpBin $ fromJust $ case opReturnType op of
+  IntType t -> Just $ Mul t OverflowUndef
+  FloatType t -> Just $ FMul t
+  _ -> Nothing
+
+typeAsValue :: PrimType -> Rational -> PrimValue
```
athas, line 122: This is almost identical to `valueAsType`. Surely some code can be deduplicated?

src/Language/Futhark/Interpreter/AD.hs
```diff
@@ -0,0 +1,289 @@
...
+  | OpConv ConvOp
+
+instance Show Op where
+  show (OpBin  op) = show op
+  show (OpCmp  op) = show op
+  show (OpUn   op) = show op
+  show (OpFn   fn) = show fn
+  show (OpConv op) = show op
+
+value :: ADValue -> PrimValue
```
athas, line 69: Call this function `primalValue`. Everything is a "value" in the interpreter!

src/Language/Futhark/Interpreter/AD.hs
```diff
@@ -0,0 +1,289 @@
...
+  doOp (OpConv op) [x']
+runPrimExp (FunExp fn p _) m = do
+  p' <- mapM (`runPrimExp` m) p
+  doOp (OpFn fn) p'
+
+
+-- VJP
+
+data Tape
+  = TapeId Int ADValue
```
athas, line 220: You convinced me that the `Int` here is necessary (and it is obvious in retrospect), but it is *really* subtle. I suggest coming up with a name for it (too bad "seed" is taken - or is it?).

athas (changes requested):
This is actually pretty good and I am happy with how relatively little code it requires. There is however a pretty long list of style issues to address (and a little functionality things missing, too). I suggest `make check`.

athas:
Will this see any further changes?

athas:
I remember we had a quick chat about this, but I don't remember the outcome. I want to use this for my teaching soon, so unless I hear about changes soon, I'll make the necessary cosmetic fixes myself and merge it.

vox9:
(In case you haven't begun,) I will do it today. But from scratch. I've learned a lot from AP, after all!

athas:
What do you mean "from scratch"? Don't rewrite all the code.

vox9:
I'm not going to take a different approach, but I feel that the mess I made when mucking around the code last minute for the exam resulted in something very ugly. Essentially, I'm going through all the code I wrote, and reconsidering it. It should take no more than a day, maybe two, resulting in similar, yet cleaner code. That's fine, right?

athas:
Up to you, but I will start refactoring this PR on Monday if it's not ready then.

PR #2162, opened 2024-07-03
Add fast path for large arrays.

athas:
Currently only handles one dimensional arrays, but could be extended. Since this is only useful for generated code, I don't think wider applicability is important.

There is still work to do in the frontend, where some passes (e.g. monomorphisation) do not handle this particularly efficiently. We may want to also add ArrayVal to the frontend language, and desugar to it during full normalisation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 * New prelude function: `rep`, an implicit form of `replicate`.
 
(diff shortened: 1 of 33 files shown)


athas:
@WilliamDue With this change, `futhark cuda --library json.fut` takes 14s on my machine (vs 73s before). I think I can still make it a bit faster. Also, this is an OK demonstration of how many places you have to modify the compiler when adding a fundamental new language construct.

athas:
Now down to 9.5s.

athas:
Now 6.1s. Last bottleneck seems to be lexing.

PR #2168, opened 2024-07-22
Add support for indexing/zipping/projecting opaque arrays.

### LINKED ISSUE #2082: C ABI records
This is an enhancement proposal, aiming at better ergonomics and/or speed.

Passing arrays or records or tuples into entry points can currently be done efficiently by marshalling them transposed. To send an argument like

```
foo: [N](u8, u16)
```

the caller needs to call tuple_new() N times, potentially(?) allocating each time, and crossing the FFI barrier, making optimizations harder (if LTO can eliminate the calls; otherwise impossible).

To get back that speed, the arguments must be like this:

foo0: [N]u8
foo1: [N]u16

In this case, the array needs to be created only twice, but no extra allocations happen.

The issue with this approach is that this kind of output is rarely idiomatic in today's programming languages. To get this output, all data, if it exists in the form [N](u8, u16), must be processed again by the CPU (slow) in order to match this new shape.

Some languages without a stable internal data representation get around this problem by having another kind of record type. In Rust, a regular struct can be turned into one with a defined field order by specifying the C representation:

```
#[repr(C)]
struct Foo {
   a: u8,
   b: u16,
}
```

Such structs can be accessed on the C side as:

```
struct Foo {
  uint8_t a;
  uint16_t b;
}
```

and also as array members.

Futhark already supports primitive types laid out consecutively in memory. The purpose of the C ABI record would be to have a representation that lets it get packed into memory by the caller. Then a value like `foo: [N](c_abi(u8, u16))` would be created from an array directly:

```
struct futhark_i32_1d *futhark_new_c_abi_tuple_1d(struct futhark_context *ctx, const c_abi_tuple *data, int64_t dim0);
```

As a result, the caller has to do nearly no extra work.

### Details

This is faster than the old way only if element accesses on the futhark side are as fast as for normal records/tuples.

While the operations this gets rid of are all O(elements), this should have impact where bandwidth to the processing unit is low compared to the computations done by each entry call.

This scheme can be nested, but it makes sense to start with only primitive types at first.

The idea came to me when learning that small arrays are inefficient. I wanted an array of records, but records aren't easy to pass to Futhark. That pushed me to an array of small arrays, which turn out to be inefficient. Now I'm packing data into u32, which is both limited and annoying. It's still less annoying than splitting the data into multiple arrays.

athas:
Closes #2082.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,5 +1,5 @@
 packages: futhark.cabal
-index-state: 2024-05-22T04:42:44Z
+index-state: 2024-07-22T12:41:48Z
 
 package futhark
   ghc-options: -j -fwrite-ide-info -hiedir=.hie
--- a/docs/c-api.rst
+++ b/docs/c-api.rst
@@ -230,6 +230,11 @@ because C does not have a standard ``half`` type.  This integer
 contains the bitwise representation of the ``f16`` value in the IEEE
 754 binary16 format.
 
+.. _array-values:
+
+Arrays of Primitive Values
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 For each distinct array type of primitives (ignoring sizes), an opaque
 C struct is defined.  Arrays of ``f16`` are presented as containing
 ``uint16_t`` elements.  For types that do not map cleanly to C,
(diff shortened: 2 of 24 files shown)


athas:
This is fully documented and tested, but even if the tests work, I won't merge until I've implemented corresponding code in smlfut. This kind of interface is intended for bridges, not hand-written code, so I want to make sure it actually works for that purpose.

PR #2178, opened 2024-09-10
Futhark Formatter

WilliamDue:
This pull request contains the work in progress code formatter for Futhark.


athas:
If you want submodules, then put them under `Futhark.Fmt`. There is one rule for the `Futhark.CLI` hierarchy, and that is that the modules there corrspond exactly to CLI executables.

athas (changes requested):
The `Futhark.CLI.Fmt.AST` module does not, in fact, define an AST.

Although you should consider the formatting of the following file.

```Haskell
def x = 0x123
```

You will notice that the output is wrong. The solution is to keep around the original source bytestring and use the offsets in the AST location information to retrieve the original syntax for number literals.

athas:
I have pushed a change to `Futhark.CLI.Fmt` that makes it read input from stdin if no file is given.

WilliamDue:
> The `Futhark.CLI.Fmt.AST` module does not, in fact, define an AST.
> 
> Although you should consider the formatting of the following file.
> 
> ```haskell
> def x = 0x123
> ```
> 
> You will notice that the output is wrong. The solution is to keep around the original source bytestring and use the offsets in the AST location information to retrieve the original syntax for number literals.

We should have solved this problem [https://github.com/diku-dk/futhark/pull/2178/commits/80c79c636a6abcfe6445589c67d611af59bf8467](here). Also we will think of a better name for AST.

athas:
Since this is now pretty much functionally correct, I have marked it for review. Fix all the style issues please.

tests_fmt/build_exec_test.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+cabal build > /dev/null
```
athas, line 3: A test script should not compile the compiler. How would you use this in a CI setting?

tests_fmt/test.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/sh
```
athas, line 1: There is no need for this to be in a subdirectory. Put it in `tools` along with `testparser.sh`.

tests_fmt/build_exec_test.sh
WilliamDue: We have removed this script.

tests_fmt/test.sh
WilliamDue: We moved it to `tools` and called it `testformatter.sh`

WilliamDue:
We are decently happy with this version. Currently trailing comments works but it does not work perfectly see `tests/fmt/trailingComments0.fut`. We think we could do a hacky solution but it seems like we would have to maybe rethink some design choices a bit to make it work in a none hacky way.

Setup.hs
```diff
@@ -4,4 +4,5 @@ import Distribution.Simple
 
 main :: IO ()
 main = defaultMainWithHooks myHooks
-  where myHooks = simpleUserHooks
+  where
+    myHooks = simpleUserHooks
```
athas, line 8: Why did you change this?

src/Futhark/Fmt/Monad.hs
```diff
@@ -0,0 +1,438 @@
...
+data LastOutput = Line | Space | Text deriving (Show, Eq, Ord)
+
+-- | A data type to describe the layout the formatter is using currently.
+data Layout = MultiLine | SingleLine deriving (Show, Eq)
+
+-- | The format monad used to keep track of comments and layout. It is a a
+-- combincation of a reader and state monad. The comments and reading from the
+-- input file are the state monads job to deal with. While the reader monad
+-- deals with the propagating the current layout.
+type FmtM a = ReaderT Layout (StateT FmtState Identity) a
```
athas, line 153: Instead of `StateT FmtState Identity`, just say `State FmtState`.

src/Futhark/Fmt/Printer.hs
```diff
@@ -0,0 +1,532 @@
...
+  fmt (SizeExp d loc) = addComments loc $ brackets d
+  fmt (SizeExpAny loc) = addComments loc $ brackets nil
+
+instance Format UncheckedSpec where
+  fmt (TypeAbbrSpec tpsig) = fmt tpsig
+  fmt (TypeSpec l name ps doc loc) =
+    addComments loc $
+      doc <:> text "type" <+> l <:> sub
+    where
+      sub = sepFilter [True, not $ null ps] line [fmtName name, align (sep line ps)]
```
athas, line 431: This is a very convoluted way of writing
```
sub | null ps = fmtName name
    | otherwise = sep line [fmtName name, align (sep line ps)]

File: src/Futhark/Fmt/Printer.hs [hunk 6]
```diff
@@ -0,0 +1,532 @@
...
+  fmt (ValSpec name ps te _ doc loc) =
+    addComments loc $
+      doc
+        <:> text "val"
+        <+> sub
+        <:> text ":"
+        <+> te
+    where
+      sub = sepFilter [True, not $ null ps] line [fmtName name, align (sep line ps)]
```
athas, line 440: Similar to above. `sepFilter` is a strange function.

athas (changes requested):
The `tests/` directory is for tests that that are run. For testing whether programs are formatted correctly, I suggest you create a `tests_fmt/` directory in the repository root that contains unformatted programs, along with what they are expected to look like after formatting. See `tests_literate/` for a suite of tests that does something similar.

Setup.hs
WilliamDue: This happened when `tools/run-formatter.sh` was used for formatting.

WilliamDue: It was run on all of the the repository at some point.

athas (commented):
Please don't reformat files that are not relevant to your work.

athas:
What are all those Zone.Identifier files?

WilliamDue:
In technical terms, we have made whoopsie daisy.

WilliamDue:
> Please don't reformat files that are not relevant to your work.

I think the problem is gone now.

athas:
`tests_fmt/expected` would be more interesting if all the input files were not already formatted.

athas:
Can I merge this now?

thereseLyngby:
Test script in tests_fmt is not done yet, but should be in next hour or so. The test themselves remain "uninteresting", most of them cover issues, where the AST used to represent two different code snippets with the same term (like the pattern thing in loops).

athas:
So it is done now?

WilliamDue:
We have added it to CI now it should hopefully be "done" now. I think we have some edge case problems with correctly place comments dealing with `sep` but they should probably be squashed as they come. One of these are sumtypes, they are not located if they do not take an argument.

WilliamDue:
We think ideally the Fmt type should also be `Located`such that comments can be placed correctly and the location should be based on the expression that is formatted formats. And when concatenated with another Fmt the smallest location that is not NoLoc should be used for the new Fmt.

athas:
`chmod +x tools/testformatter.sh`

WilliamDue:
Actually a side note, there is probably quite alot of cases where `sep` places comments wrongly. This is because we had to change how sep worked because we wanted trailing comments. The thing with `Fmt` having a locaiton is probably not that good actually.

PR #2180, opened 2024-09-18
Fusion accs

coancea:
Adding fusion rules for (1) withaccs and (2) for map-flatten-scatter fusion. The latter transforms the scatter into a map-nest enclosed in an WithAcc.


src/Futhark/Optimise/Fusion/RulesWithAccs.hs
```diff
@@ -0,0 +1,652 @@
...
+
+substInSEs :: M.Map VName VName -> [SubExp] -> [SubExp]
+substInSEs vtab = map substInSE
+  where
+    substInSE (Var x)
+      | Just y <- M.lookup x vtab = Var y
+    substInSE z = z
+
+allEq :: (Eq a) => [a] -> [a] -> Bool
+allEq as bs =
```
athas, line 630: This is just `==`.

athas:
The MI100 failure is not your fault; just ignore it.

PR #2190, opened 2024-10-31
Use monadic definition of Fmt.

athas:
This moves around the uses of 'fmt' to be explicit in the printer itself rather than in the combinators. I propose this as a more explicit design.

The one exception is (<+/>), which I think is a symptom that something is wrong.

As long as Fmt must be an instance of Located, then there will be bugs.

### DIFF
--- a/src/Futhark/Fmt/Monad.hs
+++ b/src/Futhark/Fmt/Monad.hs
@@ -78,17 +78,20 @@ infixr 6 </>
 
 infixr 4 <|>
 
-type Fmt = P.Doc ()
+type Fmt = FmtM (P.Doc ())
+
+instance Located Fmt where
+  locOf _ = NoLoc -- FIXME: This is nonsense.
 
 -- | This function allows to inspect the layout of an expression @a@ and if it
 -- is singleline line then use format @s@ and if it is multiline format @m@.
 fmtByLayout ::
-  (Located a, Format s, Format m) => a -> s -> m -> FmtM Fmt
+  (Located a) => a -> Fmt -> Fmt -> Fmt
 fmtByLayout a s m =
   s
     <|> ( case lineLayout a of
-            Just SingleLine -> fmt s
-            _any -> fmt m
+            Just SingleLine -> s
+            _any -> m
         )
 
 -- | This function determines the Layout of @a@ and if it is singleline then it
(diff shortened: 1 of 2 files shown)


src/Futhark/Fmt/Monad.hs
```diff
@@ -314,143 +311,140 @@ hardline = do
...
-comment :: T.Text -> FmtM Fmt
+comment :: T.Text -> Fmt
 comment c = do
   modify (\s -> s {lastOutput = Just Line})
   pure $ P.pretty c <> P.line
 
 -- in order to handle trailing comments its VERY important to
 -- evaluate the seperator after each element in the list.
-sep :: (Format a, Format b) => a -> [b] -> FmtM Fmt
+sep :: Fmt -> [Fmt] -> Fmt
```
WilliamDue, line 325: We need to know the location to add comments inside, but it maybe prependComments should be exported and used in Printer.hs. This change currently ruin comments in some cases.

src/Futhark/Fmt/Monad.hs
```diff
@@ -314,143 +311,140 @@ hardline = do
...
 -- | Seperates element by a @s@ followed by a space in singleline layout and
 -- seperates by a line followed by a @s@ in multine layout.
-sepLine :: (Format a, Format b) => a -> [b] -> FmtM Fmt
+sepLine :: Fmt -> [Fmt] -> Fmt
 sepLine s = sep (s <:> space <|> hardline <:> s)
 
 -- | This is used for function arguments. It seperates multiline arguments by
 -- lines and singleline arguments by spaces.
-sepArgs :: (Format a) => [a] -> FmtM Fmt
+sepArgs :: [Fmt] -> Fmt
```
WilliamDue, line 340: sepArgs needs to know the Location of each argument in order to format this probably.

src/Futhark/Fmt/Monad.hs
```diff
@@ -314,143 +311,140 @@ hardline = do
...
+    else b
 
 -- | If in singleline layout seperate by spaces. In a multiline layout seperate
 -- by a single line if two neighbouring elements are singleline. Otherwise
 -- sepereate by two lines.
-sepDecs ::
-  (Format a) =>
-  [a] ->
-  FmtM Fmt
+sepDecs :: [Fmt] -> Fmt
```
WilliamDue, line 442: sepDecs also needs to know the location of each element in order to know if there should be inserted a single line or two lines between declarations.

WilliamDue (commented):
I think fmtByLayout can be used instead of (<+/>), I think it is fine idea but the problem is sorta sometime you wanna know the location of individual elements and then format them?

I guess maybe sepDecs and sepArgs could be rewritten such they only need to know the layout inside the enviroment maybe? But in this case I would still think that you would have to do something like `localLayout a $ fmt a` where a must be located.

WilliamDue:
I think something i can go along with is removing `Format` and then explicitly in sep have signature like.

```
sep :: Located a => (a -> Fmt) -> Fmt -> [a] -> Fmt
sepDecs :: Located a => (a -> Fmt) -> [a] -> Fmt
sepArgs :: Located a => (a -> Fmt) -> [a] -> Fmt
```

athas:
> I think fmtByLayout can be used instead of (<+/>), I think it is fine idea but the problem is sorta sometime you wanna know the location of individual elements and then format them?

This should not be done by low-level generic combinators. Move that kind of decision making outwards. Your concern with not duplicating code has led to a design that is more complicated and still not correct in all cases. E.g. for the `LetPat` case, we could locally define an operator `(<+/>)`:

```Haskell
where (<+/>) = if singleLine loc then (<+>) else (</>)
```

athas:
There is also no reasonable way to write correct prettyprinting of TERecord and TETuple without simply having a case for multi-line and a case for single-line. And this is *fine*. The code duplication is not that significant.

PR #2217, opened 2025-02-07
Scan Scatter Fusion

WilliamDue:
This pull request aims to extend Futharks fusion algebra such that it is able to fuse a single pass scan followed by a scatter into a single kernel. The scan scatter kernel should allow for the ability to keep the scans output in shared memory and use the output for the scatter. The goal with this fusion rule is it will improve the performance of common operations such as filters, partition and segmented reduce.

### DIFF
--- a/src/Futhark/IR/Parse.hs
+++ b/src/Futhark/IR/Parse.hs
@@ -710,6 +710,7 @@ pSOAC pr =
       keyword "redomap" *> pScrema pRedomapForm,
       keyword "scanomap" *> pScrema pScanomapForm,
       keyword "screma" *> pScrema pScremaForm,
+      pScanScatter,
       keyword "vjp" *> pVJP,
       keyword "jvp" *> pJVP,
       pScatter,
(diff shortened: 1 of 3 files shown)


athas:
All tests pass!

LiljeDue:
@athas Should we investigate if there are any benchmarks that have become slower and if not then merge the code and later optimize the problems with filter and partition? I would also prefer if we could keep the `scan-scatter-fusion` branch so I more easily can look at the history of it.

athas:
You can keep whatever you want, but yes, we need to do a systematic performance investigation. We also need to clean up the code and figure out why your hand-written filters are still faster.

athas:
I'm working through the code now and making minor adjustments. It's taking a while. The amount of work here is really impressive.

LiljeDue:
All checks passes now! @athas

athas:
Before we merge, do you want to implement the optimisation for when the postop is identity?

LiljeDue:
I did it for two-pass scan and also for single pass scan using the early write function is this not what we wanted?

athas:
Yes, then we are done.

LiljeDue:
Is it too late to not delete the branch?

athas:
You have it locally.

PR #2225, opened 2025-02-24
SegOp Argument Reordering.

WilliamDue:
This pull request reorders the argument order of the SegHist, SegScan, and SegRed operations such that the kernel body comes before the operations that are applied on the kernel body outputs.

### DIFF
--- a/src/Futhark/Optimise/MemoryBlockMerging.hs
+++ b/src/Futhark/Optimise/MemoryBlockMerging.hs
@@ -163,30 +163,18 @@ optimiseKernel graph segop0 = do
     SegMap lvl sp tps body ->
       SegMap lvl sp tps $
         body {kernelBodyStms = maxstms <> stms <> kernelBodyStms body}
-    SegRed lvl sp tps body binops ->
-      SegRed
-        lvl
-        sp
-        tps
-        ( body {kernelBodyStms = maxstms <> stms <> kernelBodyStms body}
-        )
-        binops
-    SegScan lvl sp tps body binops ->
-      SegScan
-        lvl
-        sp
-        tps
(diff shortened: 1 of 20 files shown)


src/Futhark/Optimise/MemoryBlockMerging.hs
```diff
@@ -148,15 +163,30 @@ optimiseKernel graph segop0 = do
...
+        )
+        binops
+    SegScan lvl sp tps body binops ->
+      SegScan
+        lvl
+        sp
+        tps
+        ( body {kernelBodyStms = maxstms <> stms <> kernelBodyStms body}
+        )
+        binops
```
athas, line 181: This looks very bad; make it fit on fewer lines by using a local binding `where body' = body {kernelBodyStms = maxstms <> stms <> kernelBodyStms body}`.

WilliamDue: I have prettified it now.

src/Futhark/Optimise/MemoryBlockMerging.hs
```diff
@@ -69,15 +69,30 @@ setAllocsSegOp ::
...
+    tps
+    ( body {kernelBodyStms = setAllocsStm m <$> kernelBodyStms body}
+    )
+    segbinops
+setAllocsSegOp m (SegHist lvl sp tps body seghistops) =
+  SegHist
+    lvl
+    sp
+    tps
+    ( body {kernelBodyStms = setAllocsStm m <$> kernelBodyStms body}
```
athas, line 93: Prettify this too.

WilliamDue: It is done now.

PR #2240, opened 2025-03-31
Improve the asymptotic efficiency of vjp in the interpreter

### LINKED ISSUE #2187: Improve asymptotic efficiency of reverse-mode AD in interpreter
#2186 added AD to the interpreter. For reverse-mode, this is done by associating each value with a computation graph (really, DAG) that represents how the value was constructed. This is equivalent to the "Wengert tape" notion of AD. To compute the derivative, this graph is then traversed. Currently, the computation graph is represented as a tree, meaning that we redundantly recompute parts of it, which results in potentially exponential overhead. To fix this, we must exploit sharing inherent in the graph structure. The easiest solution may be to associate every node in the tree with a unique number, which during the traversal would let us recognise that we are re-visiting a part of the graph. Since this number must be globally unique, it must be maintained in the interpreter state.

I am also open to other ways to represent the graph, but this is the simplest one I could come up with.

vox9:
Closes #2187 

Cleanup is needed, but this does the trick as far as the time complexity goes. It should run in expected linear time now. Any better, and you'd have to remove all the ugly map operations.

Here are some pretty graphs depicting the old and new execution time of this test program:
```
def f (x: i64, d: i64) =
    loop x = x for i < d do
        if x < 4611686018427387904i64 then x + x
        else x - x + 1

def main (d: i64) =
    vjp f (1i64, d) 1i64
```
Compare
New


vox9:
Does this mean I get 7.5 ECTS? ;)

vox9:
Oh, by the way, I forgot to check if any tests could be uncommented. You might want to do that before merging :)

athas:
Very nice! Let me take a look at this. If you're serious about the ECTS, we can probably figure something out by making a retroactive POCS.

src/Language/Futhark/Interpreter.hs
```diff
@@ -133,19 +134,30 @@ lookupImport :: ImportName -> EvalM (Maybe Env)
 lookupImport f = asks $ M.lookup f . snd
 
 putExtSize :: VName -> Value -> EvalM ()
-putExtSize v x = modify $ M.insert v x
+putExtSize v x = do
+  (exts, uid) <- get
+  put (M.insert v x exts, uid)
```
athas, line 139: You can write this as `modify $ first (M.insert v x)`.

src/Language/Futhark/Interpreter.hs
```diff
@@ -133,19 +134,30 @@ lookupImport :: ImportName -> EvalM (Maybe Env)
...
 
 getExts :: EvalM Exts
-getExts = get
+getExts = gets fst
+
+putCounter :: Int -> EvalM ()
+putCounter i = do
+  (exts, _) <- get
+  put (exts, i)
```
athas, line 147: `modify $ second $ const i`

src/Language/Futhark/Interpreter/AD.hs
```diff
@@ -259,58 +291,95 @@ newtype VJPValue = VJPValue Tape
...
+        s''' <- case op' of
+          OpConv op'' ->
+            -- In case of type conversion, simply convert the sensitivity
+            sequence [ doOp' (OpConv $ flipConvOp op'') [s''] ]
+          _ -> calculatePDs op' (map tapePrimal p') >>= mapM (mul s'')
+
+        -- Propagate the new sensitivities
+        foldlM (\(a, b) (c, d) -> derive c d a b) (ss', rs') $ zip p' s'''
+      else
+        error "TODO: TIS!"
```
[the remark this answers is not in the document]
vox9: Oh, my bad, I forgot to specify that. It's code that should be impossible to reach!

src/Language/Futhark/Interpreter.hs
```diff
@@ -100,7 +101,7 @@ newtype EvalM a
   = EvalM
       ( ReaderT
           (Stack, M.Map ImportName Env)
-          (StateT Exts (F ExtOp))
+          (StateT (Exts, Int) (F ExtOp))
```
athas, line 104: I know it's just an `Int`, but it is really easy to confuse it with all the other `Int`s floating around. I recommend defining a

```Haskell
newtype Counter = Counter Int
```

somewhere to indicate that it's special.

src/Language/Futhark/Interpreter/AD.hs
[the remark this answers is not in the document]
vox9: ``if r > 0 then`` = If there are more references left
``else if r == 0 then`` = If this is the last reference (in this case, it should derive the tape)
``else`` = The reference counter messed up and undercounted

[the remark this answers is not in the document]
athas: Can you put a comment to that effect?

src/Language/Futhark/Interpreter.hs
```diff
@@ -2030,9 +2051,7 @@ initialCtx =
 
         -- TODO: Perhaps this could be fully abstracted by AD?
         -- Making addFor private would be nice..
-        add x y =
-          fromRight (error "jvp: illtyped add") $
-            AD.doOp (AD.OpBin $ AD.addFor $ P.primValueType $ AD.primitive x) [x, y]
+        add x y = getCounter >>= either (error "TODO") (\(a, b) -> putCounter b >> pure a) . AD.doOp (AD.OpBin $ AD.addFor $ P.primValueType $ AD.primitive x) [x, y]
```
athas, line 2054: What's TODO here?

athas (changes requested):
Also remember to run the formatter.

src/Language/Futhark/Interpreter.hs
vox9: Sure. In that case, is there a neat way to rewrite ``getCounter = gets snd``?

src/Language/Futhark/Interpreter.hs
vox9: Nothing more than a neat error message, which forwards the error from ``doOp``.

src/Language/Futhark/Interpreter.hs
athas: That should still read like that. `getCounter` should definitely return a `Counter`. You can then have utility functions `incCounter` etc for actually manipulating it. The point is to avoid accidentally treating the counter like a number. It *is* represented as a number, but that is more of an implementation detail.

vox9:
> Very nice! Let me take a look at this. If you're serious about the ECTS, we can probably figure something out by making a retroactive POCS.

Sounds good! I'm going on a vacation now, so I'll consider it and get back to you afterwards if I decide to do so!

athas:
Good work @Vox9!

PR #2285, opened 2025-06-21
allow `~` character in package path

cwhaddon:
Allow the `~` character to be used in package paths so that sourcehut repositories (e.g., git.sr.ht/~user/project) can be used.


athas:
Could you also add an entry to the CHANGELOG?

PR #2298, opened 2025-07-03
An attempt at optimizing the `TySolve` module

jacobgummer:
These changes should provide a better asymptotic time complexity for solving type constraints by using path compression when traversing links of type variables to reduce the cost of repeated lookups. 
However, as described in our [BSc thesis](https://futhark-lang.org/student-projects/jakob-laust-bsc-thesis.pdf), our changes do not, in general, result in a better running time in practice. As such, the code needs some fine-tuning if it's ever supposed to replace the current version of the `TySolve` module.


athas:
Do you have the program you used to generate the non-synthetic benchmarks?

LaustDengsoe:
> Do you have the program you used to generate the non-synthetic benchmarks?

Yes, they aren't pretty, but I have pushed the two Python files to the `tools` folder in `automap-tysolve-optim`.
Let me know if I should place them elsewhere.

PR #2302, opened 2025-07-30
Constraint-based type checking without AUTOMAP.

athas:
I'm not sure when we will be able to finish AUTOMAP, but the new type checker
    should be doable. AUTOMAP can then be built on top of this.


athas:
The thing I need to review most carefully is in `Note [Size Inference]`, but otherwise this is surprisingly good now. It is actually faster than the type checker on `master`.

athas:
This will bring some language changes: one of them is that we will no longer generalise local functions. In fact, polymorphic local functions require *full* type annotations. I am wondering whether we should just ban polymorphic local functions entirely.

PR #2304, opened 2025-08-01
Rewrite Scatter in terms of accumulators.

athas:


athas:
This may work now, although I still have to restore the scatter-concat optimisation.

PR #2338, opened 2025-12-10
Listing of source locations for cost centres in `profile.summary`

### LINKED ISSUE #2337: Provenance is not shown in `profile.summary`
The summary file I used last looked like this:

Image

[summary.txt](https://github.com/user-attachments/files/24081475/summary.txt)

There is no source information rendered in the summary, it is only present in the `log` and in the `timeline`. These files, however, are very large.

How could the source information be added to `summary`?
I thought about an additional column, but it may make the ascii-table unreasonable wide.

Hyalunar:
closes #2337

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+* Per Cost-Centre Source Location Listings in the `.summary`-file output of `futhark profile`. (By VegOwOtenks)
+
 ### Removed
 
 ### Changed


Hyalunar:
Adds source information output to the file `profile.summary`.
The source section is separated with a header, the start looks like this:
image
[entire file: summary.txt](https://github.com/user-attachments/files/24088036/summary.txt)

This increases the size of `summary` file by **a lot**. In my example, using this program [I wrote for Advent of Code Day 09](https://git.jossco.de/vegowotenks/aoc2025/src/branch/trunk/09/Main.fut), which is 80 lines in length. This patch increased the size of the summary file by more than a factor of two!

Hyalunar:
Please consider the above implementation a proposal. I would be happy to discuss and/or experiment with different formats before settling on anything, if anyone has different ideas of what it should look like.

By the way, how do I use ormolu? It seems to produce something that resembles a git patch format. But I am failing to apply it correctly, so far.

athas:
`ormolu -i FILE` reformats the given file in place. See also `STYLE.md` for how to automatically run the formatter on all source files.

athas:
This looks pretty promising, although it's too late in the day for me to read code. I'm OK with making the summary file larger, but we could also put the location information in a different file. What do you think would be best?

Hyalunar:
> This looks pretty promising, although it's too late in the day for me to read code. I'm OK with making the summary file larger, but we could also put the location information in a different file. What do you think would be best?

I would prefer having it in the same file, but I might try again with some of your benchmarks, something bigger.
Sharing the file would help manual inspection a lot. In my workflow, I simply search for the word under my cursor after identifying an interesting name and it jumps to the next occurence, which is the source listing.

src/Futhark/CLI/Profile.hs
```diff
@@ -48,26 +63,44 @@ data EvSummary = EvSummary
...
               [ showText (sum (map (evCount . snd) rows)),
                 "events with a total runtime of",
                 T.pack $ printf "%.2fμs" total
               ]
+          costCentreSources =
+            let costCentreSourceBlocks =
+                  map (costCentreSourceLines . fst)
+                    . L.sortOn (fst . fst)
+                    $ rows
+                costCentreHeaderTitle = T.pack " Cost Centre Source Locations "
```
athas: You don't need `T.pack`; we sue OverloadedStrings so string literals are already `Text`s.

athas:
Do not worry about the failing tests; they are the GPU cluster being flaky.

I agree with keeping everything in the same summary file for now. Long-term, I hope we can include much more information per cost centre (grid size etc), and then we might even want a file per cost centre, but that is not meaningful today.

src/Futhark/CLI/Profile.hs
Hyalunar: Addressed in https://github.com/diku-dk/futhark/pull/2338/commits/09a5a9106101aab01397fe021fe621dfdc9aee88

athas:
I hadn't even noticed that your example is for the multicore backend. Cool that it works too.

Hyalunar:
> I hadn't even noticed that your example is for the multicore backend. Cool that it works too.

Very cool that futhark supports so many targets. Multicore is the only one I can (currently) run that produces profiling information.

athas:
Final thing, could you add an entry to the CHANGELOG under *Added*? Feel free to credit yourself under whatever name you prefer.

PR #2351, opened 2026-01-07
Table in HTML Cost Centre Overview

### LINKED ISSUE #2349: futhark profile: concise table of cost centres in HTML output
I would like to have a concise table of cost centres (one row per centre) in the HTML output, in addition to the current information. That is, this table would essentially be similar to the current plain text output.

Hyalunar:
closes #2349

### DIFF
--- a/src/Futhark/Profile/Html.hs
+++ b/src/Futhark/Profile/Html.hs
@@ -147,6 +147,7 @@ generateCCOverviewHtml costCentres = do
               $ H.a
                 ! A.href (fromString $ '#' : T.unpack name)
                 ! A.class_ "silent-anchor"
+                ! A.title "Click to jump to details"
               $ H.text name
             mapM_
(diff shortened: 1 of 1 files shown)


Hyalunar:
This example table is from the `LocVolCalib` benchmark, Large Dataset:
image

All the highlighted names are clickable and jump to the details in the same file.

Hyalunar:
Should I add to the Changelog for every PR or is this grouped with the other Entry, because it implements the same feature?

athas:
Let us keep it in the same entry as long as the changes are part of the same release. The changelog is for user-visible changes, not development notes.

LiljeDue:
@Hyalunar Thank you for this awesome work, it works very well and I have used it a lot already, it makes it really easy for me to find bottlenecks in my Futhark code!

PR #2360, opened 2026-02-01
Evaluate Shell-Style Comments using the LSP

### LINKED ISSUE #2358: Evaluate Shell-Style Comments using the LSP
I'm personally an user of the feature ever since I know HLS provides it, but it's seemingly niche along all language server implementations I know.

# Reference/Inspiration

E.g.: A code snippet in Haskell (you can test this with HLS in your favorite editor)
```haskell

-- >>> factorial 5

factorial :: Int -> Int
factorial 0 = 1
factorial n = n * factorial (n - 1)

```

Looks like this in my editor: (note the blue virtual text at the end of the comment)

Image

Using the editor action to `resolve the code lens`, HLS will evaluate the expression and insert the result in a comment on the line below.

Image

# Futhark

I don't think that it's possible to meaningfully embed FutharkScript in such code lenses, because the output of futhark script is non-textual most of the time. The directive syntax could serve as a precedent.

# Relevant Specification Bits

Thinks I looked up when I was curious about the implementation in Haskell and possibilities for Futhark.

- [FutharkScript Directives](https://futhark.readthedocs.io/en/stable/man/futhark-literate.html#directives)
- [LSP: Code Lenses](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLens)
- [LSP: Commands](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#command)
- [LSP: `textDocument/codeAction`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction)
- [LSP: `workspace/executeCommand`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_executeCommand)

Hyalunar:
Closes #2358

This is currently in this state:
- you must make your editor request the code lenses yourself
- you can trigger the code lens, it will verify the parameters
- if everything goes well, the lsp will crash
- evaluating the expression is not implemented yet

I would like to reuse some code from `futhark eval`, but I'm not sure how to do it, since `futhark eval` has no underlying support module and does not export anything except for `main`. I could call `main`, but it would write to `stdout`, which is inacceptable in this scenario.

### DIFF
--- a/src/Futhark/CLI/Eval.hs
+++ b/src/Futhark/CLI/Eval.hs
@@ -7,8 +7,7 @@ import Control.Monad
 import Data.Map qualified as M
 import Data.Text qualified as T
 import Futhark.Eval
-  ( EvalIO (runEvalIO),
-    InterpreterConfig (..),
+  ( InterpreterConfig (..),
     interpreterConfig,
     newFutharkiState,
     runExpr,
@@ -36,13 +35,13 @@ main = mainWithOptions interpreterConfig options "options... <exprs...>" run
 runExprs :: [String] -> InterpreterConfig -> IO ()
 runExprs exprs cfg = do
   let InterpreterConfig _ file = cfg
-  maybe_new_state <- runEvalIO $ newFutharkiState cfg file M.empty
+  maybe_new_state <- newFutharkiState cfg file M.empty
   interpreter_state <- case maybe_new_state of
     Left reason -> do
       hPutDocLn stderr reason
       exitWith $ ExitFailure 2
     Right s -> pure s
-  forM_ exprs $ \expr -> putDocLn =<< runEvalIO (runExpr interpreter_state $ T.pack expr)
+  forM_ exprs $ \expr -> putDocLn =<< runExpr interpreter_state (T.pack expr)
 
 options :: [FunOptDescr InterpreterConfig]
 options =
--- a/src/Futhark/Eval.hs
+++ b/src/Futhark/Eval.hs
@@ -1,4 +1,13 @@
-module Futhark.Eval (InterpreterConfig (..), runExpr, interpreterConfig, newFutharkiState, EvalIO (..), AbortEvaluation (..), TraceEvaluation (..), EvalRecordRef (), runEvalRecordRef) where
+module Futhark.Eval
+  ( InterpreterConfig (..),
+    runExpr,
+    interpreterConfig,
+    newFutharkiState,
+    Evaluation (..),
+    EvalRecordRef (),
+    runEvalRecordRef,
+  )
+where
 
 import Control.Exception (IOException, catch)
 import Control.Monad (foldM, when, (<=<))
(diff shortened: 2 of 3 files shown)


athas:
Perhaps you could move some of the logic from `Futhark.CLI.Eval` to a new module `Futhark.Eval`? Setting up the interpreter is somewhat intricate, and there is no reason to duplicate that logic.

athas:
What happens if the user evaluates an infinite loop, or something that uses up all your memory?

Hyalunar:
> What happens if the user evaluates an infinite loop, or something that uses up all your memory?

It loops infinitely many times or uses up all the memory. Feels like a halting problem to me. We could do some best-effort stuff like a memory or execution time limit.

Hyalunar:
> It loops infinitely many times or uses up all the memory. Feels like a halting problem to me. We could do some best-effort stuff like a memory or execution time limit.

HLS has exactly the same problem, they don't have any guardrails and it has forced me to kill the Language Server at least a few times.

Usually, implementing a time and memory limit should be hard but I think asynchronous exceptions and heap allocation limits could work like a charm here.

Leavingy the questions: what is a reasonable limit? What if the user hits it but wants to continue?

athas:
This is a good argument for doing the evaluation in a separate process.

Hyalunar:
My work on this has stalled a little bit because I'm struggling to figure out how to properly set up the interpreter.
It seems like `futhark eval` is set on reading the files from disk, which would need to be factored out because the LSP might have newer/better code than whatever is on disk (The code that is in the file might not even compile, but the LSP always has a last known good state).

athas:
The interpreter goes through the infrastructure in `Futhark.Compiler.Program`, which supports a [VFS](https://github.com/diku-dk/futhark/blob/976cf56142ac807d452b74efbaa27ea58c535312/src/Futhark/Compiler/Program.hs#L82-L87) structure for overriding disk contents.

Hyalunar:
The basic support is there, just not very fleshed out yet.
[asciicast](https://asciinema.org/a/r8Rv4lfYKKnQHzOR)

athas:
Cool!

@vox9, this might also be a useful application for your work.

Hyalunar:
The feature works as intended now. I would appreciate feedback on some improvements/design decisions I have in mind.

# Timeout
Currently, the evaluation is limited using [a memory limit on total allocations](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/System-Mem.html#v:enableAllocationLimit), this is documented to be portable.
Additionally, a timeout could be implemented using [asynchronous exceptions](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Control-Concurrent.html#v:throwTo) but this seems to be GHC-Specific. Would this be a problem?  
Also, the Language Server is Unresponsive while the evaluation is running. I have not yet found out how to circumvent this.

# Trace
When the evaluation times out, all the traces are thrown away. It's possible to retain and display them, but they might be **huge** (since the computation timed out), maybe it's better not to?

# Hitting the Limits
There is no way to use the feature when your expression times out. I'm open for ideas on how this could be avoided.

- configuration, but from where?
- re-running with increased limits (where could the required information be stored reliably?)
- disabling the limit for a single run with a special directive

athas:
> Additionally, a timeout could be implemented using [asynchronous exceptions](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Control-Concurrent.html#v:throwTo) but this seems to be GHC-Specific. Would this be a problem?

No, being GHC-specific is not a problem.

> When the evaluation times out, all the traces are thrown away. It's possible to retain and display them, but they might be huge (since the computation timed out), maybe it's better not to?

By traces you mean Futhark-level debugging traces? You could cap them to some arbitrary number (say, the last 100) and then also print how many have been discarded.

> There is no way to use the feature when your expression times out. I'm open for ideas on how this could be avoided.

This is a UI issue, but I don't really know whether editors expose a UI to configure things like this. The easiest solution is to allow a custom timeout to be specified next to the expression in the source code.

Hyalunar:
I would leave the feature at this point for this PR. Adding the support for configurable limits might be a big bikeshed (because there must be syntax for it) which could be discussed in an issue. 
Maybe it's not a problem at all and the limits just go along nicely (Heh, probably not).

src/Futhark/LSP/CommandType.hs
```diff
@@ -0,0 +1,5 @@
+module Futhark.LSP.CommandType (CommandType (..)) where
+
+data CommandType
+  = CodeLens
+  deriving (Show, Read, Enum, Bounded)
```
athas: I will admit I'm a little concerned with this use of `Read`, as it seems to be a very subtle connection between the name of a Haskell constructor and a name that is part of the LSP protocol. It works right now, but in a somewhat surprising way.

athas:
It's too bad we don't have any automatic testing for the LSP functionality. Some of this stuff is quite subtle. I'm not sure how to test it, but *surely* someone has figured out something.

Hyalunar:
> It's too bad we don't have any automatic testing for the LSP functionality. Some of this stuff is quite subtle. I'm not sure how to test it, but _surely_ someone has figured out something.

I thought the same, I faintly remember a package in the `lsp`-family that could help with this. I might come back to this.

Hyalunar:
Testing would be possible by accepting another dependency: [`lsp-test`](https://hackage.haskell.org/package/lsp-test)

athas:
It would be a dependency in the test suite, which is essentially free.

Issue #1160, opened 2020-10-13
Shorthand syntax for in-place record updates


jarble:
Futhark has a convenient shorthand syntax for [in-place updates](https://futhark-book.readthedocs.io/en/latest/language.html#in-place-updates), but this syntax doesn't work with records:
```
let record_example a b = 
    let var = {a,b} in    
    let var.a = var.a + 1
    in vars
```
There's another way to do this, but it's more verbose:
```
let record_example a b = 
    let var = {a,b} in    
    let var = var with a = var.a + 1
    in vars
```
Would it be possible to make this "shorthand" syntax work with records as well as arrays?

athas:
That may not be a bad idea.  I'm surprised nobody has proposed it before.  I don't think there are any ambiguities or great technical hurdles to overcome.

Ulrik-dk:
I'd like to second this proposal, and extend it with a request for the ability to update arrays in records. This is currently impossible -  the solution given here [#1560] is to create a new record. That quickly becomes unwieldy, and difficult to read, for large records.

athas:
That's a bit more intricate.  A good starting point would be to implement expressions such as `r with a.b.c[i] = x`.  For simplicity, we could start by requiring the index to be at the end.

Ulrik-dk:
That would be very appreciated (by me).

Ulrik-dk:
I have some records, the definitions of which are ~400 lines long, and that have arrays which need in-place updating, so de- and reconstructing these records is not feasable.

athas:
I think a record of that size will give you other problems.  I don't think anyone will be working on this feature in the short term, but I suggest writing lens-like accessor functions manually to make the composition of updates more sane.

athas:
[This tutorial for Haskell seems to get the point across](https://www.fpcomplete.com/haskell/tutorial/lens/) although it goes much further than you'd need to in Futhark.

PR #2362, opened 2026-02-04
Adding support for shortened in place record update

armadi1809:
Resolving #1160


athas:
I suggest that we merge this once the basic functionality works, and then do a second PR that unifies record and array updates into a single concept.

This will also need an update to the grammar in the language reference before it can be merged.

athas:
Could you also add some negative tests?

athas:
CI fails because you have not implemented the case in `Futhark.Fmt.Printer`.

armadi1809:
Updated the missing case in the formatter.

PR #2365, opened 2026-02-09
Update WebGPU emdawnwebgpu compatability, add many tests 🖖 

Jesssullivan:
#### Update WebGPU backend for emdawnwebgpu

Updates the WebGPU backend to work with current emscripten 4.x and the emdawnwebgpu port (Dawn C API).

Compiler changes:
- Change emscripten flag from `-sUSE_WEBGPU` to `--use-port=emdawnwebgpu` (Actions.hs)
- Remove `#include <emscripten/html5_webgpu.h>` (CWebGPU.hs, declarations moved to webgpu.h)

Runtime changes `rts/c/backends/webgpu.h`:
- Replace `char const *` with `WGPUStringView` for labels, messages, and shader code
- Update callback signatures to use `WGPUStringView` and dual userdata parameters
- Replace `CallbackInfo` struct pattern for async operations
- Flatten `WGPUSupportedLimits`/`WGPURequiredLimits` to `WGPULimits`
- Move uncaptured error callback to `WGPUDeviceDescriptor`
- Rename: `WGPUBufferMapAsyncStatus` → `WGPUMapAsyncStatus`, `WGPUMapModeFlags` → `WGPUMapMode`
- Add `wgpu_str()` helper for string view conversion

**Tests, and more tests!**
- Add 20 :WebGPU-specific test files (tests/webgpu/*.fut)
- Tag the 59 existing tests with `-- no_webgpu` (atomics, i64 operations, features not in WebGPU)

**Tested on:**
- Rocky Linux 10, Chrome 133
- macOS aarch64 macOS 15.3, Chrome 144

---

Ah, also mandelbrot demo notes for testing these compat modifications incoming shortly

### DIFF
--- a/src/Futhark/CodeGen/Backends/GenericPython/AST.hs
+++ b/src/Futhark/CodeGen/Backends/GenericPython/AST.hs
@@ -197,7 +197,7 @@ instance Pretty PyClassDef where
     "class"
       <+> pretty cname
       <> ":"
-        </> indent 2 (stack (map pretty body))
+        </> indent 2 (stack $ punctuate line $ map pretty body)
 
 instance Pretty PyExcept where
   pretty (Catch pyexp stms) =
@@ -207,4 +207,4 @@ instance Pretty PyExcept where
       </> indent 2 (vsep $ map pretty stms)
 
 instance Pretty PyProg where
-  pretty (PyProg stms) = vsep (map pretty stms)
+  pretty (PyProg stms) = stack $ punctuate line $ map pretty stms
--- a/src/Futhark/CodeGen/Backends/PyOpenCL.hs
+++ b/src/Futhark/CodeGen/Backends/PyOpenCL.hs
@@ -315,18 +315,17 @@ readOpenCLScalar mem i bt "device" = do
             ArgKeyword "dtype" (Var $ compilePrimType bt)
           ]
   stm $ Assign val' nparr
-  stm $
-    Exp $
-      Call
-        (Var "cl.enqueue_copy")
-        [ Arg $ Var "self.queue",
-          Arg val',
-          Arg mem,
-          ArgKeyword "src_offset" $ BinOp "*" (asLong i) (Integer $ Imp.primByteSize bt),
-          ArgKeyword "is_blocking" $ Var "synchronous"
-        ]
+  stm . Exp $
+    Call
+      (Var "cl.enqueue_copy")
+      [ Arg $ Var "self.queue",
+        Arg val',
+        Arg mem,
+        ArgKeyword "src_offset" $ BinOp "*" (asLong i) (Integer $ Imp.primByteSize bt),
+        ArgKeyword "is_blocking" $ Var "synchronous"
+      ]
   stm $ Exp $ simpleCall "sync" [Var "self"]
-  pure $ Index val' $ IdxExp $ Integer 0
+  pure $ fromStorage bt $ Index val' $ IdxExp $ Integer 0
 readOpenCLScalar _ _ _ space =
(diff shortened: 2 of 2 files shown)


Jesssullivan:
re. discussion at [futhark-webgpu #1](https://github.com/diku-dk/futhark-webgpu/issues/1)

athas:
I don't expect CI to succeed since it also fails for the `webgpu` branch.

athas:
Oh boy, a lot of these failing tests are actually tests that did not run in the first place (and haven't in who knows how long) because they were missing the `==` syntax, which this PR adds. I think I'll fix that in `master` before merging this, because it becomes a bit too chaotic if the non-WebGPU tests start failing.

athas:
Where do the expected results in `tests/webgpu` come from? A lot of them seem wrong.

Jesssullivan:
> Where do the expected results in `tests/webgpu` come from? A lot of them seem wrong.

Aha, I see!  I've been incorrectly shifting with ushr (`map2 (i64.>>>)` generating false expected results.

athas:
Futhark doesn't have `>>>` (although it did many years ago). Instead, logical shifts are done by using unsigned types. Most of the other errors seem related to Futhark's `%` operator, which is a proper modulo operation, and not a remainder operation as in C (that is still available in Futhark as `%%`). These operations differ in how they handle negative operands.

athas:
@TudorPal Could you check whether this works for you on a new version of Emscripten?

TudorPal:
yup, everything works now, thank you @Jesssullivan for sharing your work! I also had the changes for `WGPUSupportedLimits/WGPURequiredLimits` and `WGPUBufferMapAsyncStatus` made myself, but this clears the confusion I had for the others. So the "cannot convert [x] to BigInt" was due to an emcc version mismatch probably?

image

There's a couple of things worth to mention for the code. This version requires `base version >= 4.20.x`, so `ghc version >= 9.10.x`. I have also ran into a couple of errors when trying to build this futhark version, so I ended up having to add/modify a couple of dependencies to the .cabal file.

image

athas:
I don't understand why we need more recent versions of these dependencies, but we can look at that when we are ready to merge into `master`.

TudorPal:
> I don't understand why we need more recent versions of these dependencies, but we can look at that when we are ready to merge into master.


yeah, it seems the errors might happen because of some cache in .cabal. I ended up changing the base version to `base >=4.19` (and removing the changes I made to the .cabal file) and running 
`rm -rf ~/.cabal/packages/hackage.haskell.org/[package_name]`, then make clean, configure, build for each error I got and it seems to work (waiting for the last build now)

`rm -rf ~/.cabal/packages/hackage.haskell.org` should do the same job, but I figured redownloading everything would take longer than manually testing and see where my cache issues are

TudorPal:
removing cache worked, but there seems to be a difference between ghc 9.8.4 and 9.10.x in `foldl'` not being exported by the Prelude module in the former. Should I go ahead and import it from Data.List? It would be done in 3 files:

`src/Futhark/IR/Prop/Scope.hs`
`src/Futhark/Internalise/LiftLambdas.hs`
`src/Futhark/Internalise/Defunctionalise.hs`

image

as per this [discussion](https://forum.exercism.org/t/foldl-v-foldl-foldl-seemingly-missing/13903)

athas:
Everything works fine with GHC 9.10 for me. Building in CI works too, so there is no reason to do anything.

TudorPal:
For 9.10 yes, i was talking about backwards compatibility with 9.8.4, but if that's not needed all's well

athas:
Whether or not it is needed is orthogonal to WebGPU. I think I got rid of it precisely due to this foldl business.

PR #2369, opened 2026-02-14
Unification of record and array updates. 

armadi1809:
This PR aims to add the possibility of modifying record fields for records in an array. 
The syntax proposed looks like this: 

```futhark
def main (x: i32) : i32 =
  let xs = [{f=1i32}, {f=2i32}]
  let ys = xs with [0].f = x
  in ys[0].f
```

### DIFF
--- a/docs/language-reference.rst
+++ b/docs/language-reference.rst
@@ -497,16 +497,14 @@ literals and variables, but also more complicated forms.
       : | `exp` [ ".." `exp` ] "..>" `exp`
       : | "if" `exp` "then" `exp` "else" `exp`
       : | "let" `size`* `pat` "=" `exp` "in" `exp`
-      : | "let" `name` `slice` "=" `exp` "in" `exp`
-      : | "let" `name` "." `fieldid` ("." `fieldid`)* "=" `exp` "in" `exp`
+      : | "let" `name` (`slice` | "." `fieldid`)+ "=" `exp` "in" `exp`
       : | "let" `name` `type_param`* `pat`+ [":" `type`] "=" `exp` "in" `exp`
       : | "(" "\" `pat`+ [":" `type`] "->" `exp` ")"
       : | "loop" `pat` ["=" `exp`] `loopform` "do" `exp`
       : | "#[" `attr` "]" `exp`
       : | "unsafe" `exp`
       : | "assert" `atom` `exp`
-      : | `exp` "with" `slice` "=" `exp`
-      : | `exp` "with" `fieldid` ("." `fieldid`)* "=" `exp`
+      : | `exp` "with" (`slice` | `fieldid`) (`slice` | "." `fieldid`)*  "=" `exp`
       : | "match" `exp` ("case" `pat` "->" `exp`)+
    slice: "[" `index` ("," `index`)* [","] "]"
    field:   `fieldid` "=" `exp`


armadi1809:
@athas I changed the code to unify Update and RecordUpdate in a single new constructor. 

This is definitely still in draft mode. I tested the basic functionalities for record and array updates. I would like your opinion on this, especially around `src/Futhark/Internalise/Exps.hs`, `src/Language/Futhark/TypeChecker/Consumption.hs` and `src/Language/Futhark/TypeChecker/Terms.hs`

The added `update-index-field1.fut` test shows field index with the `x with f[i]` syntax.

src/Futhark/Internalise/Exps.hs
```diff
@@ -745,49 +745,97 @@ internaliseExp desc (E.Not e loc) = locating loc $ do
...
                     m
               k = internalisedTypeSize t
-              (bef, to_update, aft) = splitAt3 i k src'
-          src'' <- replace t fs ve' to_update
-          pure $ bef ++ src'' ++ aft
-    replace _ _ ve' _ = pure ve'
+              (bef, _to_update, aft) = splitAt3 i k src_vals
+           in pure $ bef ++ ve_vals ++ aft
+    replace _ _ ve_vals _ = pure ve_vals
+internaliseExp desc (E.UpdatePath src steps ve _ loc) = locating loc $ do
```
athas: I am somewhat uneasy about expressing the internalisation as a rewrite on the source language. It's something I've done before, but I usually end up regretting it. In this case, I expect that the rewrite may produce something that is not type-correct IR - not for any interesting reason, but because of redundant computation. I'll have to think if I can come up with a case where it fails.

src/Language/Futhark/TypeChecker/Consumption.hs
```diff
@@ -876,13 +883,24 @@ checkExp (AppExp (LetWithField dst src fields ve body loc) appres) = do
...
-  slice' <- checkSubExps slice
+checkExp (UpdatePath src steps ve t loc) = do
+  steps' <- mapM checkStep steps
   (ve', ve_als) <- checkExp ve
   (src', src_als) <- checkExp src
-  overlapCheck (locOf ve) (src', src_als) (ve', ve_als)
-  consumeAliases (locOf loc) $ aliases src_als
-  pure (Update src' slice' ve' loc, second (const mempty) src_als)
+  let hasIndex = any isIndex steps
+  when hasIndex $ do
```
athas: This seems fine to me. It is certainly not less coarse than the old check. But I would suggest checking `hasIndex` only once, e.g. write something like:

```Haskell
res_als <- if hasIndex then do
             overlapCheck ...
```

athas:
@WilliamDue Do you have any opinions on this?

LiljeDue:
I think this change will be useful in improving readability, from my limiting understanding I think the changes seems like the way to go about it.

src/Futhark/Internalise/Exps.hs
athas: I think you could base it on the old implementation of `Update`. The trick is mostly to ensure that you don't end up evaluating source expressions or indexes more than once. E.g. for `src with ...steps... = v`, you'd first internalise once and for all `src` and `v`, both of which turn into `[VName]`s, and then you just recurse down the `steps`, and for each one implement the update.

athas:
I have taken the liberty of pushing the following changes:

1. Putting tests into subdirectories to reduce clutter.
2. Rewritten some error cases in internalisation to be more concise.

athas:
I have also added two tricky tests to make your life more interesting. One of them fails.

armadi1809:
@athas All tests should now be passing. I messed up the order in `getOrdering` at `FullNormalise.hs` for the update case and it was mismatching the consumption order defined in `checkExp` at `Consumption.hs`. It is now fixed and the tests are passing for me :)

athas:
Is this still intended to be a draft PR?

armadi1809:
> Is this still intended to be a draft PR?

No, I don't think so. I just changed the title accordingly.

athas:
This is still missing the implicit form (`let x[i].f = y in z`), isn't it?

armadi1809:
I just added support for the last missing form in the parser. Also added a test to cover it.

PR #2378, opened 2026-02-23
Output f32 and f64 values unambiguously

Jesin:
The C backend's output fails to distinguish between adjacent floating-point values even when those values are between 4 and 8, and for values with small exponents it fails to even distinguish them from 0. Meanwhile, values very far from 0 get printed with way too many digits. The `g` format specifier has been in the standard for a long time, so let's use it.

`FLT_DIG` and `DBL_DIG` are the wrong values to use for unambiguous output. `FLT_DECIMAL_DIG` and `DBL_DECIMAL_DIG` were introduced in C11 for this purpose. Assuming we want to keep C99 support, I have included a workaround.


athas:
This is only for the text format output, which is very rarely used. I think it is only used when compiling a program to a non-server executable, and then not using binary output.

It should still be correct, of course, so thanks for the PR.

Issue #1008, opened 2020-06-08
'futhark autotune' should also have a progress bar


athas:
I often use `-v` just to see that something is going on, but a progress bar (enabled by default) would serve the same purpose and be more concise.

We can just factor out the code from `futhark bench` for this.

PR #2380, opened 2026-02-24
'futhark autotune' should also have a progress bar

Nader-Rahhal:
In regards to #1008

### DIFF
--- a/src/Futhark/CLI/Autotune.hs
+++ b/src/Futhark/CLI/Autotune.hs
@@ -356,7 +356,8 @@ tuneThreshold opts server datasets (already_tuned, best_runtimes0) (v, _v_path) 
                     | fromIntegral rt * epsilon < fromIntegral best_t -> do
                         T.putStrLn $
                           T.unwords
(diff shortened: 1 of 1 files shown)


athas (changes requested):

athas:
Thanks for the PR. This looks promising. Remember to also add a mention to the `CHANGELOG.md` file, including whatever name you wish to be known under.

athas:
Apart from the style issues, I'll try this out tomorrow. The code seems good, but I need to see what it looks like (and that the non-fancy terminal case works well).

Nader-Rahhal:
> Apart from the style issues, I'll try this out tomorrow. The code seems good, but I need to see what it looks like (and that the non-fancy terminal case works well).

Sounds good. I have yet to see what it looks like on non-fancy either and will try to test it today.

Nader-Rahhal:
> > Apart from the style issues, I'll try this out tomorrow. The code seems good, but I need to see what it looks like (and that the non-fancy terminal case works well).
> 
> Sounds good. I have yet to see what it looks like on non-fancy either and will try to test it today.

Update to this, it fine for when I redirect output to a file, which (I think?) should emulate non-fancy

athas:
Thanks for your PR!

PR #2382, opened 2026-02-24
Correctly-rounded fma primitive

### LINKED ISSUE #2377: fma is not actually fused in futhark repl
`echo 'entry main = f32.fma' >fmatest.fut && futhark c fmatest.fut && echo '24871.96875 41563.3984375 -1.03376352e9' | ./fmatest` outputs `27.081299f32`, which is the correct value of `f32.fma` on those inputs.
Evaluating `f32.fma 24871.96875 41563.3984375 (-1.03376352e9)` inside `futhark repl` produces `0.0` due to intermediate rounding, which would be fine for `mad` but does not match the specification of `fma`. 
It would be convenient for me to be able to test algorithms that actually want the improved precision of `fma` inside the repl. It appears GHC.Base does provide `fmaddFloat#` and `fmaddDouble#` for purposes like these. I hope that means it would be easy enough to implement, though I am not too familiar with futhark's internals.

Jesin:
Fixes #2377. If you want to avoid depending on GHC.Exts in this file, I think the implementation I wrote for `f16.fma` should also work for f32 and f64, but I believe futhark already depends on GHC so I think using the fast implementation should be fine.

### DIFF
--- a/src/Language/Futhark/Primitive.hs
+++ b/src/Language/Futhark/Primitive.hs
@@ -139,7 +139,7 @@ import Foreign.C.Types (CUShort (..))
 import Futhark.Util (convFloat)
 import Futhark.Util.CMath
 import Futhark.Util.Pretty
-import GHC.Exts (Double(D#), Float(F#), fmaddDouble#, fmaddFloat#)
+import GHC.Exts (Double (D#), Float (F#), fmaddDouble#, fmaddFloat#)
 import Numeric (log1p)
 import Numeric.Half
 import Prelude hiding (id, (.))


Jesin:
I just noticed the fma16 implementation can return nan when it should return inf, specifically in the case where c is infinite and has the opposite sign of (a * b) while a and b are both finite but (a * b) overflows. Gonna fix that real quick.

Jesin:
Ok, I believe that's fixed now.

athas:
I am OK with this approach. Out of curiosity, what are you working on that requires this degree of correct rounding?

Jesin:
I'm working on some geometric problems involving Bezier curves, some of which require solving several quartic equations per frame, then several cubic equations per pixel per frame. There are also some other small-matrix determinants (and/or cross products) in other parts of the code. It turns out if you have a decent fma implementation available and you do [this trick](https://pharr.org/matt/blog/2019/11/03/difference-of-floats) with it, these all become much more numerically stable and it becomes actually viable to do a lot of this stuff in f32 without getting so many artifacts around edge cases, which is much faster than switching to f64 on many GPUs.
(I personally have no use for f16, but I implemented that anyway because I wanted to be able to call this bug fixed.)

Jesin:
I just realized my f16.fma implementation sometimes returns `0.0` when it should return `-0.0`. I believe this rarely matters, but it is technically a bug.
I believe a lot of the operations provided by the upstream `half` package are also not correctly-rounded, since they're implemented in terms of operations on f32 without accounting for which way those f32 operations were rounded. This includes `fromRational`, so I think it is not actually possible for us to implement this function correctly on f16 using the interfaces provided by `half` until `half` is fixed. Maybe I'll find an example and report that upstream in case anyone cares.

PR #2385, opened 2026-03-01
Unify sections

### LINKED ISSUE #2375: Generalising index sections
With #2362 and #2369, we have unified array and record updates. Sections are still divided, however: while we can write `(.[0])` and `(.f)`, we cannot write `(.[0].f)` (which would be syntactic sugar for `(\x -> x[0].f)`. What do you say @armadi1809; are you up for doing this as well?

armadi1809:
Resolves #2375


athas:
There should also be negative tests that detect type errors, and some with slices `(.f[i:j])` for good measure.

armadi1809:
@athas I just added more tests. Let me know if we need more.

PR #2390, opened 2026-03-05
Add commands to introspect and manipulate composite values in the C server

vox9:


athas:
It seems that fully opaque types (which turn out to be *quite rare*, much to my surprise) are considered "scalar" in this work:

```
$ futhark c --server tests/issue2231_a.fut
$ tests/issue2231_a
%%% OK
kind t
scalar
%%% OK
```

This is surprising to me, unless the name "SCALAR" is very precisely defined. I think I would prefer separate "PRIMITIVE" and "OPAQUE" kinds.

vox9:
The documentation desperately needs to be rewrapped, this is killing me. Can I do it in this PR?

vox9:
Furthermore, what's the status on the array `new` and `update` functions for the C API?

athas:
I'll try to implement the array functions over the weekend. Feel free to rewrap the documentation. It's currently wrapped at 72 columns for historical reasons, but all recent work uses 80 columns.

athas:
Don't rewrap the *entire manual*, just the single file you touch. Also, you rewrapped some code blocks, and I do not have the time to manually review thousands of lines right now.

vox9:
Okay, okay, I went a little mad with power there. I found some smaller spelling / documentation errors earlier, should I fix those?

vox9:
Oh, you already sneakily fixed it when you looked at my other work! Nothing more from me, then :)

PR #2393, opened 2026-03-12
Test Cases for the Language Server

Hyalunar:


Hyalunar:
I am aware of the fact that the test I just introduced is currently failing

Hyalunar:
Tests are running and passing but I have not investigated the nix build failure.
I have refactored `logStringStderr` in the Language Server because it kept interfering with hunit output, I'm not sure about the introspectability yet.

Hyalunar:
I can fix the style issue but it's really a trailing whitespace in a string literal, so that's weird

athas:
Another complication is that multi line strings are a GHC 9.12 addition and Futhark CI uses (I think) GHC 9.10. It's not particularly important to me that we maintain support for older GHCs, however.

athas:
Actually, for practical reasons I'd rather stick to the default GHC in Nixpkgs (which is 9.10). Consider replacing your multi-line strings with quasiquoted multi-line strings, like here: https://github.com/diku-dk/futhark/blob/master/src/Futhark/CodeGen/Backends/PyOpenCL/Boilerplate.hs#L57

It's not quite as nice, but it's also not very bad.

Hyalunar:
> Another complication is that multi line strings are a GHC 9.12 addition and Futhark CI uses (I think) GHC 9.10. It's not particularly important to me that we maintain support for older GHCs, however.

I did that in https://github.com/diku-dk/futhark/pull/2393/commits/9e05e76b31094b249e00b5f67aeaf7134e749c83, uses `neat-interpolation` like `futhark` itself does. I wasn't aware of the GHC version needed for `MultilineStrings` since I am always on the newest version anyway :] (futhark does build with `ghc-9.14.1`, by the way)

Hyalunar:
Do you have specific LSP-Workflows you would like to be tested? I'd add more but they increase the test suite runtime noticeably

athas:
Ideally every LSP feature should be tested, but I don't really expect that for this PR. It's good enough to do a handful, and otherwise focusing on setting up the infrastructure allowing tests to be added for future features.

PR #2395, opened 2026-03-14
Add array rank, join, and set functions to the c server

vox9:
Also fixes an error in the documentation of `elemtype`.


vox9:
(Fixed both in the above commit)

PR #2399, opened 2026-03-16
Implement calling of Futhark Server programs from the interpreter

vox9:
I added ugly starter code. Please don't roast me too hard, I prefer my meat medium rare.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -526,13 +526,13 @@ jobs:
 
     - uses: mymindstorm/setup-emsdk@v14
       with:
-        version: 2.0.18
+        version: 4.0.23
         actions-cache-folder: 'emsdk-cache'
         cache-key: emsdk-wasm
 
-    - uses: actions/setup-node@v3.5.1
+    - uses: actions/setup-node@v4
       with:
-        node-version: '16.x'
+        node-version: '22.x'
 
     - uses: actions/download-artifact@v4
       with:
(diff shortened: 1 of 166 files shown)


vox9:
At the time being, most utility functions of the interpreter are unsupported. This will change, when I have the time to do so. Furthermore, primitive values stored anywhere inside arrays are lazy pulled from the server. Right now, you have to force that by performing a binary operation on them in the end. This should be an easy fix, but not tonight :)

athas:
This deserves a closer look than I have time for today. However, I am somewhat suspicious of the three modules added to `Futhark.Util`, as I do not envision they will actually ever be used outside the interpreter. I don't think anything is gained by making them look more general than they really are, and I would strongly prefer moving them under the `Language.Futhark.Interpreter` hierarchy, to emphasize that this is their home (I also have a long-term dream of separating the interpreter further from the compiler, but let's not worry too much about that).

athas:
Also, run the formatter and kill the warnings. I won't do that here, because that would probably result in ruinous conflicts with your own upstream code.

athas:
Actually I am going to reformat this and fix all the warnings, because there are other things I also want to modify. Make sure to pull after that.

athas:
Reformat done. I also fixed some warnings while I was at it.

athas:
I have beefed up the `futhark eval` user experience significantly. Note that it now only uses compiled code when `--backend` is passed.

vox9:
Looking good, but remember that only the C server currently supports all the necessary commands.

athas:
That is the only one that matters.

athas:
Closed in favour of #2507.

PR #2405, opened 2026-03-24
Property-Based testing tool for futhark (resolve #2163)

### LINKED ISSUE #2163: Property-based testing
It is time to develop a property-based testing system for Futhark. The main challenge is that Futhark itself is not rich enough to directly implement a QuickCheck-like system, so the actual driver loop has to be external. It would still be preferable for generators (and perhaps even shrinkers) to be Futhark entry points.

Last time we tried it, we didn't have the server protocol, so we were quite limited in which kinds of values we could manipulate externally. This is no longer the case.

Perhaps we can directly piggyback on an existing Haskell library (e.g. possibly QuickCheck itself), but it must not be dependent on a Haskell compiler at run time.

Black-Speedy:
This is not the finished work but it's a start. We have made a basic integration of a separate property based testing tool into futhark test.  The basic use inside of a futhark program is:

```futhark
entry bar (size: i64) (seed: i32): testType

#[prop(gen(bar), shrink(auto)]
entry foo (x: testType) : bool

-- possible shrinker with type:
entry baz (x: testType) (strat: i32) : (testType, i8)

-- if you have an opaque type or you want custom pretty printing you can provide it
entry fooz (x: testType) : []u8
```
The attribute can take a size, seed, and pretty printer as well. So a fully custom attribute could be: 
`#[prop(gen(bar), shrink(baz), pprint(fooz), size(10), seed(12347)]`.
The generator, property, and shrinker must be entry points or it will not work. Every attribute is assigned to a property and is specific to each property, they do not share anything. For the shrinker you can write auto to use our auto shrinking strategy or omit it completely which will not shrink. If you have created your own shrinker use its function name instead.

Note that currently it is not fully working perfectly together with the regular test. If the regular test fails on any test it will overwrite some of our printed outputs.


athas:
I would like to see the tests related to testing property-based testing to a directory `tests_property` in the root of the repository. Long term, we will of course use property-based tests as part of the normal integration test suite in `tests`, but it is good to have tests for the testing tool itself as well.

src/Futhark/CLI/Test.hs
```diff
@@ -871,3 +1041,55 @@ main = mainWithOptions defaultConfig commandLineOptions "options... programs..."
...
+              <> name
+              <> "' block in the test specification."
+          ]
+      | name <- declaredWithoutRequest
+      ]
+
+-- Save to file helper functions
+propToFile :: FilePath -> [TestResult] -> IO ()
+propToFile testFile results = do
+  let fileName = testFile <> ".propResult"
```
athas: Style nit: Futhark convention is for all files to use `snake_case`, except where other languages (such as Haskell) demands otherwise.

athas:
I noticed these defects:

* Plain `#[prop]` does not work; I must write `#[prop()]`.
* A generator is required, even for builtin types such as arrays of primitives.

athas:
I found a rather serious error: if a seed is explicitly provided by the user (with `--seed`), then it is used unchanged all the time, i.e., every seed argument to a generator will be the same. This obviously makes it completely useless. The provided seed should be an *original* seed, but it should be modified (deterministically) after every use, e.g. by passing it through a hash function, incrementing it by one, or whatnot.

athas:
Actually, I have an idea for how to fix it in a clean and nice way; let me try it out this afternoon.

Black-Speedy:
> I found a rather serious error: if a seed is explicitly provided by the user (with `--seed`), then it is used unchanged all the time, i.e., every seed argument to a generator will be the same. This obviously makes it completely useless. The provided seed should be an _original_ seed, but it should be modified (deterministically) after every use, e.g. by passing it through a hash function, incrementing it by one, or whatnot.

I had noticed this as well, it was a problem when providing a seed but not otherwise (this was for before when seed was random every iteration which got removed)

> Actually, I have an idea for how to fix it in a clean and nice way; let me try it out this afternoon.

I missed your message before I pushed my latest update. Please take a look and if it doesn’t align with what you had in mind, I’m happy to go with your version instead.

athas:
I made my change. I think it is good. Take a look and make sure you understand its implications.

athas:
I discovered that automatic generators would never produce negative floating-point numbers. This is now fixed. The range is now `(-1,1)`, which is perhaps still problematic, but I think that is also what QuickCheck does. Further notes:

* I have realised that it would be more natural if generators received a `u64` as their seed.
* It seems wrong that automatic generator always produce square arrays (that is, all sizes are the same).

athas:
In your estimation, what remains before this can be merged?

athas:
Would it also not make sense for the seed passed to shrinkers to be `u64` now? I think they are not consecutive anyway.

Black-Speedy:
> In your estimation, what remains before this can be merged?

I think we are pretty close. I think we need a bit of code cleanup, but most of the features should be there. We would also like to cleanup in our test directory e.g. removing redundant and unnecessary tests. 

> Would it also not make sense for the seed passed to shrinkers to be `u64` now? I think they are not consecutive anyway.

We could do this, I was also considering it but hadn't done anything about it.

athas:
Please do so.

I'm also considering ripping out the prettyprinting machinery entirely. While it works, it is so awkward that I don't expect anyone to ever use it, and the automated prettyprinting is pretty good.

athas:
Now the tests fail. Is this intentional? I don't understand the purpose of the most recent commit.

Simoerk:
> Now the tests fail. Is this intentional? I don't understand the purpose of the most recent commit.

Sorry, not intentional. Removing for now and fixing it later.

athas:
Assuming the tests now work, is there any reason not to merge this?

Simoerk:
> Assuming the tests now work, is there any reason not to merge this?

We are doing cleanup in the test folder. 

After that, we plan to adjust the critical error messages so they make it clearer to the user that these are unexpected/internal failures rather than ordinary property-test failures.

Paranoia mode for shrinking is not implemented yet. Do you think this should block the merge, or would it be acceptable to leave it as future work or do it after?

athas:
It is fine to implement additional features later.

athas:
There is no way to use `peekWord64` correctly. Something has gone very wrong with how you think about the RNG state.

athas:
I have fixed it. I strongly recommend not playing games with RNG states. Always use them in a linear fashion; otherwise the code is impossible to audit.

Black-Speedy:
It should be ready to merge now. The unnecessary tests have been removed, and the rest have been renamed to follow snake_case. 🎉🎉🎉

PR #2406, opened 2026-03-24
Code Actions to insert inferred type ascriptions

Hyalunar:
Every inlay hint is now accompanied by a code action that can insert the type ascription it shows.
insert-type-ascription

This is open for discussion, because there are still some things left to figure out:
- what to do with the compiler-generated unicode-typenames (they are not recognized by the parser)
- why is it impossible to insert return types?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Scan chunk size is now exposed as tuning parameter in `cuda` and `hip`
   backends.
 
-+ `futhark lsp` offers code actions, they insert type ascriptions for
+* `futhark lsp` offers code actions, they insert type ascriptions for
   inferred types. Every named binding has an action.
+
+* `futhark bench` and `futhark test` can now handle entry points that return
+  opaque values, as long as there is no expected result.
 
 ### Removed
 
@@ -29,6 +32,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 * `i64.set_bit`/`u64.set_bit` would produce wrong results in C-based backends. (#2396)
+
+* Some uses of higher order modules could cause infinite loops. (#2407)
 
 ## [0.25.36]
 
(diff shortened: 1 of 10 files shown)


athas:
> what to do with the compiler-generated unicode-typenames (they are not recognized by the parser)

You could rewrite them to be non-Unicode. The point of them was to make compiler-generated names completely distinct from programmer-written names, but I'm not sure that ever bore much fruit, so we can maybe also just ditch it.

> why is it impossible to insert return types?

I don't know. Why is it?

Hyalunar:
> You could rewrite them to be non-Unicode. The point of them was to make compiler-generated names completely distinct from programmer-written names, but I'm not sure that ever bore much fruit, so we can maybe also just ditch it.

Yes, I will probably do that. From what I've gathered from the source code, only the subscript digits are inserted so it should suffice to replace them with normal digits.



> I don't know. Why is it?

That was a me problem, it is possible, I had not found it.
It's because I cannot specify where the code action will be shown, the language server protocol client figures that out by itself, somehow. This makes me fear that the experience might differ based on which editor is used.

athas:
I wonder if this works in Emacs with eglot.

Now that this infrastructure is here, maybe it would be worth considering if we can do more interesting things than just inserting inferred types. Certainly, future program analyses we write ought to take LSP integration into serious account.

Hyalunar:
> I wonder if this works in Emacs with eglot.

I certainly hope so, the [github repository does suggest it](https://github.com/joaotavora/eglot?tab=readme-ov-file#code-actions) but I have no clue about its' authenticity.



> Now that this infrastructure is here, maybe it would be worth considering if we can do more interesting things than just inserting inferred types.

The classics, which are also [baked into the Language Server Protocol](https://hackage.haskell.org/package/lsp-types-2.3.0.1/docs/Language-LSP-Protocol-Types.html#t:CodeActionKind) are stuff like:
- inline variable body
- extract function
- extract variable

what are you imagining/thinking about?

Hyalunar:
Personally, I'd love to try hole-filling, but that seems to be a different kind of beast.

Hyalunar:
I'm finished, I can't think of any more edge cases to adress :]

I've recorded a small demo:
ActionDemo

Features include:
- parameter types
- return types
- lambda argument types
- automatically introduces
  - referenced type parameters
  - referenced size parameters

Issue #2403, opened 2026-03-23
Changing how entry points return tuples


athas:
At the moment, if you write an entry point that returns a tuple, such as

```
entry main (x: i32) = (x+1, x+2)
```

then at the C API level the tuple will be unpacked into multiple output values. This happens only for tuples - any other return type results in a single output value.

This design has been in place for a long time, and actually originates from before we had a C API at all, when standalone executables were the only way to run Futhark programs. The motivation was being able to test functions that returned multiple values as tuples. When we added the C API, tuples were completely opaque, and so this implicit unpacking was maintained. However, for a long time it has now been possible to unpack opaque tuples via the C API, and so the implicit unpacking is no longer necessary. Further, I have seen users be occasionally confused by it, and it causes a rather arbitrary special case when interacting with Futhark from other languages. Hence I am strongly considering removing implicit unpacking of tuples, and treating them like any other type, meaning entry points will *always* produce just a single result.

For `futhark test` (and `futhark bench`) we can maintain the current illusion of multiple values, simply by unpacking the tuple using the relevant API commands.

All of the bridges to other languages should continue working, as they are already able to handle entry points that return just a single opaque value, and hopefully they also support the tuple unpacking functions.

Nevertheless, this *will* break API compatibility, specifically for users that invoke the C API directly. This is not something we should do lightly, especially as the C API has been quite stable for a long time.

FluxusMagna:
While I have gotten used to the current system(and this would inevitably break a bunch of my code) I do see the appeal of consistency. Could you make a separate branch to try this out before it hits the main line? It would be a nice motivator to finish the new version of futhask, which includes handling of records. Since the old days I've been using entry points to unpack values and being forced to replace that mess is probably a good thing in the long-run.

athas:
Sure, it's not exactly urgent.

FluxusMagna:
There is potentially an issue that could arise from naively implementing this that would be very annoying (at least for me) to deal with: Currently we can return tuples of *named* types by specifying their type in the entry point return value, which is of great convenience when working with types whose explicit signatures are big. As far as I know this naming system does not go deeper than what is explicit in the entry point, so implementing this could effectively remove the ability to return multiple values with named types from one entry point. This would pretty much force users to use the compiler generated names, in which case I think the explicitly named types should disappear from the C API entirely. 

[...]

PR #2410, opened 2026-03-28
Make entry points always return a single value.

athas:
This implements the proposal from #2403.

If we merge this, then we should also modify the server protocol and manifest slightly, as entry points can no longer return multiple outputs. It is not urgent, as the existing representation does still work.


athas:
Although this is indeed a breaking change, I am convinced that it is the right thing to do. I will merge this PR once I have confidence that it works. I am not greatly looking forward to updating all the downstream users of Futhark's C API...

PR #2437, opened 2026-05-02
Fix `convienent` -> `convenient` typo in tools README

pikammmmm:
Line 12 of `tools/README.md` has `convienent` (transposed letters).


athas:
Thanks for the fix.

PR #2438, opened 2026-05-02
Move Nix pinning system from 'niv' to flakes, spring cleaning in Nix code.

swagtop:
## Why move to flakes

Despite still being labeled experimental, flakes have become the de facto pinning system that almost everyone in the Nix ecosystem uses. They come with benefits other than pinning, like the new `nix` command CLI, truly reproducible and deterministic builds with restricted evaluation.

The flake in this PR is introduced as a superset of the original behaviour. Merging this would allow flake and non-flake users to interact with the Nix in this codebase in the way they prefer. 

With these changes the command `niv update nixpkgs -b master` is superseded by `nix flake update`.

## How does this affect non-flake users

No changes in workflow are needed for non-flake users. This includes the GitHub CI, which I have tested and found to work just fine. These changes would roll out without non-flake users noticing anything had changed.

### 1. Move pinning system from `niv` to Nix flakes

Implemented by having the `nix/sources.nix` file read the `flake.lock` file using `flake-compat`, and returning the flake inputs. In essence the old `sources.nix` and `sources.json` are replaced by the new `sources.nix` and `flake.lock`.

The result of this is that `import ./nix/sources.nix` keeps the same behaviour as with the `niv` setup. 

The `flake.nix` that flake users interact with is very simple, and basically sets up the devshells, formatter, and a couple of packages for all flake exposed systems.

Last, the nixpkgs version in the `flake.lock` file is the same as the current one `niv` setup is pinned to.

### 2. Format code with official nixpkgs formatter

Have all Nix code follow the same, modern formatting conventions. In the future this formatter can be run with `nix fmt`.

### 3. Update some expressions to modern conventions

Shells defined with `pkgs.stdenv.mkDerivation { buildInputs = [ ... ]; }` become `pkgs.mkShell { packages = [ ... ]; }`, remove all uses of `rec`, introduction of flakes, modern formatting, etc.

### 4. Split futhark release-tarball derivation creation up into clear separate steps

The current definition of the tarball made by running `nix-build` is a mouthful. I have split it up into three steps in the new `default.nix`. First, the base `futhark` derivation is defined, then the `futhark-static` derivation overrides the first by making the build static, then the tarball derivation is defined, taking the `futhark-static` derivation in as an input.

All three derivations are available as packages in `flake.nix`, with the default one being the tarball. Running `nix-build` still results in the tarball being built.

Splitting things up like this should make it much easier to maintain, and comprehend how the whole thing fits together.

### 5. Remove unused file (nix/zlib.nix)

In the spirit of cleaning up the Nix code, I have removed this file as it isn't referred to anywhere in the codebase.


athas:
I am going to test this out now.

athas:
Using the old CLI (which is actually not so important to me):

* `nix-shell` works fine.
* `nix-build` works as otherwise CI would fail.

Using the new CLI:

* `nix develop` works fine.
* `nix shell` is completely useless because the derivation produces a `.tar.xz`.
* `nix build` works fine.

athas:
I see no reason to switch halfway, so I would also like to replace `nix-build` in CI with `nix build`. However, then we need to adapt our use of `--argstr`.

athas:
Let us do that in a followup PR. I don't see a reason not to fully switch to Flakes, as long as we do not lose functionality.

PR #2439, opened 2026-05-04
Use shared memory in double buffering when possible.

WilliamDue:
I am not quite sure this is the correct way of handling it, but I tried to allow the usage of shared memory for the double buffering pass. My concerns are that I actually do not account for if there is enough space for it to be in shared memory, it just mainly tries to allocating it if it in a `SegBlock` and it is of a constant size.

### DIFF
--- a/src/Futhark/Optimise/DoubleBuffer.hs
+++ b/src/Futhark/Optimise/DoubleBuffer.hs
@@ -274,10 +274,10 @@ optimiseLoop (Pat pes) merge body@(Body _ body_stms body_res) = do
         Just arr_lmad <- findLmadOfArray arr_v,
         Just (arr_mem_out_alloc, body_stms'') <-
           extractAllocOf bound_in_loop arr_mem_out body_stms' = do
-          -- Use the space from the body alloc rather than the loop parameter,
-          -- since ExplicitAllocations may have given the parameter a strided
-          -- device layout (e.g. when fed from exscan_last), even though the
-          -- actual array only lives within the thread block.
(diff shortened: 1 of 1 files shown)


athas:
Is this just papering over a deficiency in ExplicitAllocations, or is DoubleBuffering actually doing something wrong?

LiljeDue:
I will have to look more at ExplicitAllocations to be sure, I have no experience with GPUMem.

athas:
Do you have a small example of DoubleBuffering doing the wrong thing?

LiljeDue:
Should something like this not be able to fit in shared memory?
```
def main xss =
  #[incremental_flattening(only_intra)]
  map (\xs -> iterate 10 (map (+ 2f32)) xs) xss
```
It does global memory accesses at least in the loop.

athas:
Judging from `futhark dev --gpu -a`, the choice to use `device` memory is made by ExplicitAllocations. Fixing it in double buffering is wrong.

athas:
If you write it like this, then the loop parameter is in shared memory:

```
def main xss =
  #[incremental_flattening(only_intra)]
  map (\xs -> iterate 10 (map (+ 2f32)) (manifest xs)) xss
```

The rule in ExplicitAllocations is pretty similar to the one in double buffering: a loop parameter is kept in the same memory space that the initial value is in.

LiljeDue:
In my more complicated piece of code I seem to fail at using manifest to force it to allocate it in shared memory.

athas:
You need to shrink it and figure out what is happening. This part of the compiler is too complicated to randomly modify.

PR #2453, opened 2026-05-14
Add zip and unzip commands to server protocol.

athas:


athas:
@vox9 You may find this useful.

PR #2458, opened 2026-05-17
Reconciling the WASM backend api to match the WebGPU format

TudorPal:
As per this [issue](https://github.com/diku-dk/futhark-webgpu/issues/8):

I added the `FutharkModule` interface to the WASM backend now and updated the `js_api.rst' documentation, so the preferred setup can also be:

```js
const fut = new FutharkModule();
await fut.init();
```

while `newFutharkContext()` and `FutharkContext` are still kept for backwards compatibility. I also added the WebGPU-style utility methods to the WASM wrapper (`context_sync`, `clear_caches`, `report`, `pause_profiling`, `unpause_profiling`) and updated the tests to check both the old WASM API and the new shared `FutharkModule` API. 

I have also added the `tunnel` example from this repo to `tests_lib/wasm/tunnel` in the master branch and changed it to work with the files generated by WASM, so basically the only difference between this and the WebGPU one was that we load `tunnel.mjs` instead of `tunnel.js` and initialise it through the shared `FutharkModule` API with `await fut.init()` instead of passing the WebGPU runtime module. It works, but the FPS is much lower than the WebGPU version, which makes sense since this example computes every pixel on the CPU with WASM instead of on the GPU with WebGPU. So i changed the resolution of the canvas to prove the test works.


athas:
> while newFutharkContext() and FutharkContext are still kept for backwards compatibility. 

I don't think there are enough users to worry about that. I would suggest removing these.

docs/js-api.rst
```diff
@@ -138,10 +165,22 @@ are represented by ``FutharkOpaque`` objects.
 FutharkArray
 ------------
 
-The exact ``FutharkArray`` API differs slightly between backends.
+Arrays are represented by ``FutharkArray`` objects.  The common array API is
+shared by the WASM backend and the WebGPU backend.
+
+.. js:function:: FutharkArray.get_shape()
```
athas: I would prefer just `shape`. Or is that reserved by JavaScript for other reasons?

docs/js-api.rst
```diff
@@ -40,15 +40,28 @@ For the WebGPU backend, a build typically produces:
...
+
+For the WebGPU backend, ``init`` must be passed the generated backend
+runtime module:
+
+.. code-block:: javascript
+
+   const module = await Module();
+
+   const fut = new FutharkModule();
+   await fut.init(module);
```
athas: Could we not have the WASM backend require the same thing? I get that it is not necessary, but it is harmless, isn't it? Having an identical interface is valuable.

docs/js-api.rst
```diff
@@ -125,6 +133,25 @@ wrapper is initialised with the generated backend runtime module.
...
+
+   Asynchronously create a new ``FutharkContext`` object.
+
+.. js:class:: FutharkContext()
+
+   Backwards-compatible wrapper class for WASM-style backends.
+   ``FutharkContext`` supports the same main interface as
+   ``FutharkModule``, including ``entry``, ``types``, array constructors,
+   and ``free``.
+
```
athas: Just get rid of these.

athas (changes requested):
Very nice work! I only have a few remarks.

docs/js-api.rst
TudorPal: I don't think it is reserved by JS, but in WASM, `shape()` is currently a method that returns a normal JS array of `BigInt`s:

```js
shape() {
  this.validCheck();
  var s = this.fshape(this.ctx.ctx, this.ptr) >> 3;
  return Array.from(this.ctx.wasm.HEAP64.subarray(s, s + this.dim));
}
````

while WebGPU currently stores the shape as a field:

```js
constructor(name, arr, shape) {
  this.type_name = name;
  this.arr = arr;
  this.shape = shape;
}
```

and exposes it through:

```js
get_shape() { return this.shape; }
```

So if we want `shape()` to be the shared API, I think the clean change is to avoid using `shape` as a field in the WebGPU `FutharkArray` object, rename that internal field to something like `shape_data` or `shape_`, and then add:

```js
shape() {
  return this.shape_data;
}
```

For WASM, `shape()` can stay as it is, except we may want it to return a `BigInt64Array` instead of a normal array. I thought having something similar to a "getter" might be intuitive enough, as there are more functions that follow this naming convention, but we can do it however you might see fit.

athas: I think the clean solution is to have a `shape()` method that returns a `BigInt64Array` in all cases, and change the internal representation to whatever is necessary to make that work.

docs/js-api.rst
TudorPal: So right now both versions work and mean the same thing in the WASM version, so both of these snippets:

```js
// shared explicit style
import Module, { FutharkModule } from "./futlib.mjs";

const module = await Module();
const fut = new FutharkModule();
await fut.init(module);
```

and:

```js
// WASM convenience fallback
import { FutharkModule } from "./futlib.mjs";

const fut = new FutharkModule();
await fut.init();
```

But in order to have access to Module, the user needs to import it from the generated mjs file, while for WebGPU it comes from loading the generated runtime file separately (by loading the script like this in the html file)
```html
<script src="build/tunnel.js"></script>
```

Do you want me to enforce one way of writing or leave both available?

athas: I would like you to enforce one way. I will let you decide which is better.

TudorPal: Ok, I chose to enforce the explicit `init(module)` style for `FutharkModule`, since this is already required by WebGPU

docs/js-api.rst
athas: This hasn't been done yet, correct?

TudorPal: I have just pushed the [changes](https://github.com/diku-dk/futhark/commit/1a1b8873d6df5966852059a0a6e0101a751cad83) for the wasm backend. I will start working on the webgpu one and open a new pull request, as that needs to go to the webgpu branch, not the master one

athas: Great, thanks for your work.

PR #2463, opened 2026-05-19
Custom derivatives for reverse-mode.

athas:
This morning I realised that it would actually be quite easy to support custom derivatives for reverse mode (and possibly for forward mode, although it's less important). The idea is basically to provide a way for users to specify a different "VJP" operation for some fragment of code, which the AD transform will then just insert.

The user interface is this new intrinsic function:

```Futhark
-- | Provide custom reverse-mode adjoint code for a given function. This is
-- useful when the adjoint synthesised by AD is as good as one that is known
-- analytically.
--
-- The function `f` returns a result of type `b`. In the return sweep, the
-- function `f'` is invoked first with the result of `f` and second with the
-- cotangents of the result (be careful not to mix up the order), and must
-- return the sensitivity with respect to the input.
--
-- A common pattern is that `b` is a tuple where some part is the intended
-- primal result of `vjp_by`, and some part is only used in `f'`.
--
-- **Beware:** if `f` uses any free variables, these will not be taken into
-- **account when computing the adjoint. Make these part of the argument
-- **instead.
def vjp_by 'a 'b 'c (f: a -> b) (f': (res: b) -> (b_adj: b) -> a) (x: a) : b =
  intrinsics.vjp_by f f' x
```

I have written all of one test, but it seems to work quite well. I have not pondered whether this interacts with nested AD in any interesting way. The interpreter just ignores any use of `vjp_by`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * The restrictions for passing tuples as consumed function parameters have been
   loosened. (#2456)
 
+* The reverse-mode AD transformation now supports custom adjoints through a new
+  prelude function, `vjp_by`.
+
 ### Removed
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   loosened. (#2456)
 
 * The reverse-mode AD transformation now supports custom adjoints through a new
-  prelude function, `vjp_by`.
+  prelude function, `with_vjp`.
 
 ### Removed
 
--- a/prelude/ad.fut
+++ b/prelude/ad.fut
@@ -117,7 +117,7 @@ def vjp 'a 'b (f: a -> b) (x: a) (y': b) : a =
   (vjp2 f x y').1
 
 -- | Provide custom reverse-mode adjoint code for a given function. This is
--- useful when the adjoint synthesised by AD is as good as one that is known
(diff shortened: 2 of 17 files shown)


athas:
@UncatchableAlex you may find this relevant to your interests.

athas:
Here is an example of how it is used to provide better adjoint code for sorting:

```Futhark
def radix_sort_step [n] 't (f: t -> u32) (xs: [n]t) (b: i32) : [n]t =
  let bits = map (\x -> (i32.u32 (f x >> u32.i32 b)) & 1) xs
  let bits_neg = map (1 -) bits
  let offs = reduce (+) 0 bits_neg
  let idxs0 = map2 (*) bits_neg (scan (+) 0 bits_neg)
  let idxs1 = map2 (*) bits (map (+ offs) (scan (+) 0 bits))
  let idxs2 = map2 (+) idxs0 idxs1
  let idxs = map (\x -> x - 1) idxs2
  let xs' = scatter (copy xs) (map i64.i32 idxs) xs
  in xs'

def radix_sort [n] 't (f: t -> u32) (xs: [n]t) : [n]t =
  loop xs for i < 32 do radix_sort_step f xs i

def differentiable_radix_sort [n] 't (f: t -> u32) (xs: [n]t) =
  (vjp_by (\xs ->
             unzip (radix_sort (f <-< (.0)) (zip xs (iota n))))
          (\(_, perm) (xs_adj, _) ->
             scatter (copy xs_adj) perm xs_adj)
          xs).0

entry main_standard = vjp (radix_sort f32.to_bits)
entry main_custom = vjp (differentiable_radix_sort f32.to_bits)
```

athas:
As suggested by Tom Smeding, this construct was renamed `with_vjp`.

PR #2466, opened 2026-05-24
Fix: support for long files was broken on windows

jeromew:
While testing https://futhark-lang.org/blog/2026-05-22-benchmarking-a-real-futhark-application.html on windows it appeared that the benchmark could not reload the data.in file generated by the application.

The issue ended up coming from the fslurp_file function in util.h because of a problem related to files > 2GB.

The main issue is that `ftell` returns a `long` and `long` is 32 bit on windows so this API cannot be used.

This PR changes all usage of `ftell`/`fseek` by the `ftello`/`fseeko` API which was designed to return `off_t` instead of  `long`

It is also necessary to define `_FILE_OFFSET_BITS=64` to make sure that `off_t` uses 64 bits.
The PR defines it as an argument in runCC for all platforms which I think is a good practice to enable large file support.

With this PR, the benchmarking works for me on windows

### DIFF
--- a/rts/c/cache.h
+++ b/rts/c/cache.h
@@ -67,8 +67,8 @@ static int cache_restore(const char *fname, const struct cache_hash *hash,
   if (fseek(f, 0, SEEK_END) != 0) {
     goto error;
   }
-  int64_t f_size = (int64_t)ftell(f);
-  if (fseek(f, CACHE_HEADER_SIZE, SEEK_SET) != 0) {
+  int64_t f_size = (int64_t)ftello(f);
+  if (fseeko(f, CACHE_HEADER_SIZE, SEEK_SET) != 0) {
     goto error;
   }
 
--- a/rts/c/server.h
+++ b/rts/c/server.h
@@ -1624,7 +1624,7 @@ int restore_opaque(const struct opaque_aux *aux, FILE *f,
   // (which doesn't care if there's extra at the end), then we compute
   // how much space the the object actually takes in serialised form
   // and rewind the file to that position.  The only downside is more IO.
-  size_t start = ftell(f);
+  size_t start = (size_t)ftello(f);
   size_t size;
   char *bytes = fslurp_file(f, &size);
   void *obj = aux->restore(ctx, bytes);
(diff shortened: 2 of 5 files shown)


athas:
Thanks for finding the problem. I think it is however a bad idea to define `_FILE_OFFSET_BITS=64` in the C compiler invocation, as that will not work when using `--library` - it is better to define it in the generated code. Specifically, I think it should go [in here](https://github.com/diku-dk/futhark/blob/5217fa05f753d4169e627ec51561c9133cdc38a4/src/Futhark/CodeGen/Backends/GenericC.hs#L307-L317).

athas:
I clicked the wrong button - the fix naturally welcome, but please check if the issue remains fixed when `#define _FILE_OFFSET_BITS 64` is put in `gnuSource`.

jeromew:
ok I agree it makes sense to put in in gnuSource

I found a way to make it work ; I had a small difficulty because in order to have Futhark work I have another thing to do because `getline` is somehow missing in my toolchain so i use

`CFLAGS="-include ../custom.h -O -std=c99" futhark.exe`

with custom.h beeing the following file :

```c
#include <stdio.h>
#include <stdlib.h>

ssize_t getline(char** lineptr, size_t* n, FILE* stream);

ssize_t getline(char **lineptr, size_t *n, FILE *stream)
{
	size_t count = 0;
	int c;
	char *newLineptr;

	if (!*lineptr) {
		*n = (*n > 0) ? *n : BUFSIZ;
		*lineptr = (char*)malloc((*n) * sizeof(char));
	}

	if (!*lineptr) {
		return -1;
	}

	while ((c = fgetc(stream)) != EOF) {
		count++;

		if (*n <= count) {
			*n = *n * 2 > count ? *n * 2 : count * 2;
			newLineptr = (char*)realloc(*lineptr, *n);
			if (!newLineptr) {
				return -1;
			}

			*lineptr = newLineptr;
		}

		*(*lineptr + count - 1) = c;
		*(*lineptr + count) = '\0';

		if (c == '\n') {
			return count;
		}
	}

	return -1;
}
```

since custom.h is included before the gnuSource in GenericC.hs it was missing the `#define _FILE_OFFSET_BITS 64`.

For now I added the `#define _FILE_OFFSET_BITS 64` also at the top of my custom.h and it works.

would you be ok to have an if WIN32 in rts/c/server.h with this getline definition inside albeit without the 2 includes at the top which would not be needed in this case ?

athas:
`getline()` is a POSIX-function, so it is expected that it is missing on Windows.

jeromew:
I don't know what is the good way of bringing getline for windows support.
The custom.h solution works for me now as an implementation of `getline` is needed for `--server` usage.

athas:
`getline()` does not have to be a builtin; we could implement our own version. We'll do so if more Windows people show up.

jeromew:
Maybe I should document more precisely in the windows section / faq section the error i get when i don't have the getline implementation in custom.h. I'll probably push a PR at some point.

PR #2469, opened 2026-05-27
Sort consecutive imports to be in alphabetical order.

WilliamDue:
This pull request makes it so `futhark-fmt` will sort consecutive imports to be in alphabetical order.

### DIFF
--- a/src/Futhark/Fmt/Monad.hs
+++ b/src/Futhark/Fmt/Monad.hs
@@ -35,6 +35,7 @@ module Futhark.Fmt.Monad
     sepComments,
     sepLineComments,
     sepLine,
+    consecutive,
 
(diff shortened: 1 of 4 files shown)


LiljeDue:
This is probably a bit hacky to switch the locations, I am not quite sure how to do this in a nicer manner, because sorting them changes the location order.

athas:
This may change the semantics of the program.

LiljeDue:
Oh really? I never thought about how Futhark resolves overlap in imported declarations.

athas:
It uses shadowing. Last import wins. There is no way for the formatter to avoid the risk of changing the program.

LiljeDue:
Arh that is sad, oh well.

PR #2470, opened 2026-05-28
FP16 support for HIP backend (no emulation)

jeromew:
This PR makes sure that FP16 has direct hardware support in HIP backend.

Currently, the FP16 support on HIP is emulated, which creates a performance hit for FP16 applications.
This PR was tested on `fuchat` and it makes the FP16 version faster than the FP32 version, as was expected since FP16 uses less memory bandwidth for the same calculations.

There was one difficulty related to what i think is a bug in rocm FP16 support for which i created an issue on [github.com/ROCm/clr/issues/274.](https://github.com/ROCm/clr/issues/274).

The "<=" operator does not always return False when a NaN is compared. It broke the `tests/primitive/naninf16.fut`.

The proposed solution until the rocm issue is fixed is to prettyPrint `comple16(x,y)` instead of `x <= y` when we are dealing with Float16 scalars.

All the tests with backend=HIP are passing.

### DIFF
--- a/rts/c/scalar_f16.h
+++ b/rts/c/scalar_f16.h
@@ -9,7 +9,10 @@
 // under emulation, so the compiler will have to be careful when
 // generating reads or writes.
 
-#if !defined(cl_khr_fp16) && !(defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 600) && !(defined(ISPC))
+#if !defined(cl_khr_fp16) && \
+     !(defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 600) && \
+     !(defined(__HIP_DEVICE_COMPILE__)) && \
+     !(defined(ISPC))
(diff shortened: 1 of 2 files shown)


athas:
Somehow I always thought we did this, but I guess the f16 emulation is robust enough that we just overlooked the issue when we added the HIP backend.

jeromew:
@athas
"<=" and "<" for Float16 are now printed as cmple16 and cmplt16.
when i moved the implementation to be able to have a HIP impl. I mis-read the EMULATE_F16 serie of #ifdefs and forgot to add the default implementation in the emulated section. The last commit should hopefully fix the tests.

I am surprised the `test-opencl` seems to be using the emulated code. Maybe there is `!defined(cl_khr_fp16)` on the test machine.

athas:
Thanks for getting this fixed!

PR #2475, opened 2026-05-31
Reconcile WebGPU and WASM JavaScript APIs, as per diku-dk/futhark-webgpu#8

TudorPal:
The main changes are:

* Updated the WebGPU JavaScript wrapper so single-output entry points return the value directly, matching the WASM wrapper behaviour.
* Updated the WebGPU array wrapper to expose the shared array API:

  * `shape()`, returning a `BigInt64Array`
  * `values()`
  * `toTypedArray()`
  * `toArray()`
  * `futharkType()`
* Renamed the internal WebGPU array shape field to avoid conflicting with the public `shape()` method.
* Adjusted the wrapper to handle the current WebGPU manifest format, where entry points use `output` instead of `outputs`.
* Ensured the generated WebGPU runtime exports the Emscripten runtime methods needed by the wrapper, such as `ccall`. I'm not sure why in the `compileWebGPUAction` action from `Actions.hs` the option of `"sEXPORTED_RUNTIME_METHODS=cwrap,ccall,Asyncify,HEAP8,...` was not enough so i had to add the extra options to the `runEMCC` action.

I also added matching API tests/examples for both WASM and WebGPU:

* `test_inc`
* `test_array`
* `tunnel` example from the `futhark-webgpu` repo

The goal is that the WASM and WebGPU tests are structured similarly, so it is easier to compare the exposed APIs across the two backends.

There is also a `test_inc_headless` test for WebGPU. This runs through a headless browser and currently checks the parts that are stable in that environment. Some WebGPU utility methods, such as `context_sync`, can fail in the headless setup due to the issue related to unsupported `shader-f16`, so the headless test only checks that those methods are exposed. The full functionality can still be tested manually in the normal browser version.


athas:
Thanks for the PR. All the wasm backend tests now seem to fail. I'll see if I can figure out why.

PR #2487, opened 2026-06-21
Make units and arrays of units decomposable in C FFI

athas:
With this change, *all* types exposed via the FFI can be decomposed: they are either a record, a sum type, an array of opaques, a primitive, or an array of primitives. This means any Futhark value can be reflected completely in other languages (yes, this does imply the FFI actually ignores abstract types - I'm willing to discuss whether this is the right thing).


athas:
As a followup to this PR, we might also consider reworking the manifest to make it clear that an opaque type is expected to be *something* we can inspect. Currently it is very tacked-on.

PR #2492, opened 2026-06-25
Remove the neutral element from Scan SOACs and SegScan kernels

CoderDue:
`Scan` becomes a `newtype` over `Lambda rep` only — no more
`scanNeutral`.  Iteration 0 is handled with an **is_first trick**: a
boolean (init `true`, set `false` after one use) guards the scan
operator so that `op(ne, x) = x` is implemented by taking the map
result directly.  The same trick adapts the GPU single-pass / two-pass
SegScan kernels and the multicore SegScan.

The motivation is freedom from a value-level neutral for monoids
where one isn't natural (user-defined operators, non-`Prim` accs,
costly-to-materialise zeros), plus removing one source of plumbing
drift across fusion, AD, ISRWIM, and ImpGen.

## What's in the PR

- **IR**: `Scan` is `newtype Scan rep = Scan (Lambda rep)`; helpers
  recompute sizes from `lambdaReturnType`.  `SegScanOp` adapted.
- **Lowerings**: `FirstOrderTransform`, `soacToStream`, `ISRWIM`,
  `ExtractKernels`, `ExtractMulticore`, `DistributeNests` no longer
  thread a neutral.  `soacToStream`'s scanomap branch emits a
  per-chunk scanomap as a `Screma` SOAC plus a top-level shift map
  combining the chunk's last value with the cross-chunk accumulator;
  this preserves the baseline path through which
  `ExtractMulticore`/`ExtractKernels` recover an inner `SegScan` from
  scan-into-stream patterns such as `reduce (+) 0 (scan (+) 0 xs)`.
- **Backends**: `ImpGen/GPU/SegScan/{SinglePass,TwoPass}` and
  `ImpGen/Multicore/SegScan` reworked to the no-neutral kernel
  shapes.
- **AD**: `AD/Rev/Scan` builds adjoint scans neutral-free; `AD/Fwd`
  and the other `AD/Rev/*` modules updated mechanically.
- **AD fix**: `AD/Rev/Loop.revLoop` filters loop params whose body
  result is a `Constant` so the adjoint loop's body and merge stay
  in lock-step.  Without this, VJP scan fusion produces a malformed
  adjoint Loop (the precise failure mode of `tests/ad/issue2256.fut`).
- **Simplify**: a new `peelIsFirstParam` rule recognises the
  is_first shape (bool merge param, init `true`, body result
  `Constant False`, used only as the sole scrutinee of top-level
  `Match`es) and peels iteration 0, dropping the merge slot.
  Registered only in the per-backend rule books (`GPU`, `MC`, `Seq`,
  `Mem`), not in `standardRules`.

## Performance

Measured on a GTX 1660 Ti vs. baseline `f2a60a6aa`.  GPU performance
is broadly preserved or improved:

| Pattern                                              | Effect on HEAD |
|------------------------------------------------------|----------------|
| `scan (+) 0f32 xs` at 67 M                           | ~17% faster    |
| `map (scan ...) xs` (rows / cols), various shapes    | 8–18% faster   |
| Generic `scan` on i32/i64/f64/tuples/scanomap        | within noise   |
| Segmented scan, 67 M, short and long segments        | **~7% slower** |

The segmented dip is the only measured runtime regression.

## Downsides

1. **`soacToStream` introduces an in-loop `if`.**  The new scanomap
   stream body wraps the per-chunk scanomap with an
   `if is_first then chunk_scan else map (\x -> scan_lam(acc, x))
   chunk_scan` shift and an analogous `if`-guarded cross-chunk acc
   update.  Both ifs sit at the top level of the body, around the
   inner SOAC, not inside any inner loop.  Two consequences:

   - It triggered the AD `revLoop` mis-shaping bug (fixed here; see
     above).  Reverting the fix breaks VJP scan fusion.
   - Cleaning up the residual `is_first`-shaped loops relies on
     `peelIsFirstParam` in the per-backend simplify.  Any future
     pass that interposes before the simplify will see the un-peeled
     shape.

   In the `sequentialStreamWholeArray` path (which both `ExtractMulticore`
   and `ExtractKernels` use), `is_first` becomes the constant `true`,
   the shift collapses to identity, and the inner per-chunk scanomap
   reduces to a single full-width `Screma` — extracted as a
   `SegScan` by the multicore / GPU walker.

2. **Segmented SegScan ~7% slower on GPU.**  Root cause: the
   no-neutral kernel guards `blockScan` with a runtime-variable
   `valid_threads_cnt` instead of the constant `tblock_size`; NVRTC
   cannot constant-fold this, register pressure rises and per-SM
   occupancy drops.  Restoring parity requires reshaping the
   partial-last-block handling so the check is elided on the common
   path.  Not done here.

3. **`peelIsFirstParam` is conservative.**  Fires only on
   constant-bound loops (≥ 1) and only when the bool param is used
   solely as a top-level `Match` scrutinee.  Symbolic-bound loops,
   nested matches, alias chains, multi-scrutinee matches, and bool
   flowing into nested loops all cause the rule to skip.  AD
   correctness does not depend on the rule firing.

## AD correctness: `eBlank` padding in PPAD reverse-mode scan

`asLiftPPAD` pads `as_lift[n-1]` with `eBlank` (zero/default) rather
than the scan neutral element, and `ysRightPPAD` does the same for
`ys_right[0]`.  Investigation confirmed this does **not** cause
incorrect results:

- The `a3 = op(pa1, pa2)` chain in the lifted reversed scan carries
  the corrupted value, but `diffScan` discards it — only chunk index 2
  (`z`, the actual gradient values) is extracted via `(!! 2) . chunk d`.
  Chunk 1 (`a3`) is dead computation.
- `ys_right[0] = eBlank` is guarded by an explicit `i=0` branch in
  `finalMapPPAD` that returns `r_adj[0]` directly, bypassing the
  operator.

Verified with an assert-based probe (`assert (b1 != 0)` in the
operator, neutral `(0,1)`, `eBlank = (0,0)`) for `n = 2, 3, 4`, with
`simplifyLambda` disabled in `Pass/AD.hs` to prevent constant
propagation from masking the issue.  No assert fired and results were
bit-identical to the reference binary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * All opaque values available over the C API can now be decomposed into their
   constituents.
 
+* The manifest now contains documentation for entry points and opaque types.
+
 ### Removed
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   uses the hardware support for `f16`, similarly to the CUDA backend.
   Implemented by Jérôme Wagner. (#2470)
 
+* All opaque values available over the C API can now be decomposed into their
+  constituents.
+
 ### Removed
 
 ### Changed
--- a/cabal.project
+++ b/cabal.project
@@ -1,5 +1,5 @@
 packages: futhark.cabal
-index-state: 2026-05-14T15:59:02Z
+index-state: 2026-06-25T17:59:37Z
 multi-repl: True
 
 package futhark
--- a/docs/manifest.schema.json
+++ b/docs/manifest.schema.json
@@ -44,12 +44,18 @@
                             },
                             "additionalProperties": false
                         }
+                    },
+                    "doc": {
+                        "type": "string"
                     }
                 }
             }
         },
         "types": {
             "type": "object",
+            "doc": {
+                "type": "string"
+            },
             "additionalProperties": {
                 "oneOf": [
                     { "type": "object",
--- a/futhark.cabal
+++ b/futhark.cabal
@@ -479,7 +479,7 @@ library
     , free >=5.1.10
     , futhark-data >= 1.1.3.0
     , futhark-server >= 1.4.1.0
-    , futhark-manifest == 1.8.0.0
+    , futhark-manifest == 1.9.0.0
     , githash >=0.1.6.1
     , half >= 0.3
     , haskeline
--- a/nix/futhark-manifest.nix
+++ b/nix/futhark-manifest.nix
@@ -1,37 +1,16 @@
-{
-  mkDerivation,
-  aeson,
-  base,
-  bytestring,
-  containers,
-  lib,
-  QuickCheck,
-  quickcheck-instances,
-  tasty,
-  tasty-hunit,
-  tasty-quickcheck,
-  text,
+{ mkDerivation, aeson, base, bytestring, containers, lib
+, QuickCheck, quickcheck-instances, tasty, tasty-hunit
+, tasty-quickcheck, text
 }:
 mkDerivation {
   pname = "futhark-manifest";
-  version = "1.8.0.0";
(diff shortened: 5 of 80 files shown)


athas:
Should I just close this?

athas:
Closed because it remains nonobvious whether we want this. Feel free to reopen if/when this is desirable.

PR #2493, opened 2026-06-26
Fix non-exhaustive pattern detection when wildcards mix with constructors

### LINKED ISSUE #2483: Pattern matching exhaustive checking
In the below example, Futhark fails to detect that pattern matching is non-exhaustive. When applying the function `op` in the interpreter (`futhark repl`), an internal compiler error occurs:

```
bash-3.2$ futhark --version
Futhark 0.26.3.
Compiled with GHC 9.14.1.

bash-3.2$ futhark repl nonexhaustive.fut
[0]> op #M #P
Pattern match failure.
```

Here is the program `nonexhaustive.fut`:
```futhark
type s = i64
type t = #I (bool,s,s)
       | #F (bool,s,s,s)
       | #P
       | #M
       | #NE

def op (a:t) (b:t) : t =
  match (a,b)
  case (#NE, _) -> b
  case (_, #NE) -> a
  case (#I(s1,x,b1), #I(s2,y,b2)) ->
    assert (s2 == false) (#I(((s1, x * b2 + y, b1 * b2))))
  case (#I(s1,x,b1), #F(s2,y,b2,f2)) ->
    assert (s2 == false) (#F(((s1, x * b2 + y, b1 * b2, f2))))
  case (#F(s1,x,b1,f1), #I(s2,y,b2)) ->
    assert (s2 == false) (#F(((s1, x + y * f1, b1, f1 / b2))))
  case (#I(s,x,b), #P) -> #F(((s, x, b, 1 / 10)))
  case (#P, #I(s,x,b)) ->
    assert (s == false) (#F(((false, x / b, 1, 1 / b))))
  case (#M, #I(s,x,b)) ->
    assert (s == false) (#I(true,x,b))
  case (#M, #F(s,x,b,f)) ->
    assert (s == false) (#F(true,x,b,f))
  case (#F _, #P) -> assert false (([])[0])
  case (#P, #F _) -> assert false (([])[0])
  case (#F _, #F _) -> assert false (([])[0])
```

CoderDue:
Fixes #2483.

## Problem

When a match column contains both wildcard patterns and explicit constructors, `incompleteCase` computed `sigma` (the set of explicit constructors) but discarded the default-matrix witnesses without checking sub-patterns of sigma constructors. If sigma happened to cover all constructors, the function returned `[]` — no warning — even when combinations were genuinely unmatched.

## Fix

Add `recurseOnConstrHeads`: when wildcards are present alongside structural constructors (sum, tuple, record), specialise on each unique constructor head and recurse to find unmatched sub-patterns. Wildcard rows are automatically included via `specialise`, so the check is correct. For Sum types with incomplete sigma, missing constructors are still reported via the default matrix as before.

Literal patterns (`ConstrLit`) are excluded from this path since their domain is infinite and enumeration is not possible.

### DIFF
--- a/src/Language/Futhark/TypeChecker/Match.hs
+++ b/src/Language/Futhark/TypeChecker/Match.hs
@@ -81,6 +81,12 @@ isBool :: Match t -> Maybe Bool
 isBool (MatchConstr (ConstrLit (PatLitPrim (BoolValue b))) _ _) = Just b
 isBool _ = Nothing
 
+isStructuralConstr :: Match t -> Bool
+isStructuralConstr (MatchConstr (Constr _) _ _) = True
+isStructuralConstr (MatchConstr ConstrTuple _ _) = True
+isStructuralConstr (MatchConstr (ConstrRecord _) _ _) = True
+isStructuralConstr _ = False
+
 complete :: [Match StructType] -> Bool
 complete xs
   | Just x <- maybeHead xs,
@@ -154,16 +160,13 @@ findUnmatched pmat n
     -- When wildcards make the column complete, recurse on unique constructor
     -- heads (wildcard rows are included via specialise).
     recurseOnConstrHeads cs = do
-      c <- nubOrd [c | c@(MatchConstr {}) <- cs]
-      let ats = case c of MatchConstr _ args _ -> map matchType args; _ -> []
+      c@(MatchConstr c' args _) <- nubOrd $ filter isStructuralConstr cs
+      let ats = map matchType args
           a_k = length ats
           pmat' = specialise ats c pmat
       u <- findUnmatched pmat' (a_k + n - 1)
-      pure $ case c of
-        MatchConstr c' _ _ ->
-          let (r, rest) = splitAt a_k u
-           in MatchConstr c' r () : rest
-        _ -> MatchWild () : u
+      let (r, rest) = splitAt a_k u
+      pure $ MatchConstr c' r () : rest
 
     incompleteCase pt cs
       | null cs = do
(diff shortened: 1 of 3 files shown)


CoderDue:
All tests pass locally:
- `futhark test -i tests`: 2441/2441 passed
- `futhark test -s tests`: 2441/2441 passed
- `futhark test -c --backend=c tests`: 2441/2441 passed
- `futhark test -c --backend=multicore tests`: 2441/2441 passed
- `futhark test -c --backend=cuda tests`: 2434/2434 passed (7 excluded)
- `futhark test -c --backend=opencl tests`: 2434/2434 passed (7 excluded)
- `tools/testparser.sh tests`: passed
- `tools/testfmt.sh tests`: 2436/2436 passed (hash, idempotent, comment-order)
- `tests_repl/test.sh`: passed
- `tests_bench/test.sh`: passed
- `tests_adhoc/test.sh`: passed
- `tests_fmt/test.sh`: passed
- `tests_property/test.sh`: 62/62 passed
- `tests_server/test.sh`: passed
- `tests_pkg/test.sh`: passed
- `cabal check`: passed
- `cabal bench`: passed

src/Language/Futhark/TypeChecker/Match.hs
```diff
@@ -147,28 +151,65 @@ findUnmatched pmat n
...
+           in sigmaWitnesses ++ missingWitnesses
+      | any isWild cs,
+        not (null [() | MatchConstr (Constr _) _ _ <- cs]) =
+          recurseOnConstrHeads cs
+      | any isWild cs,
+        not (null [() | MatchConstr ConstrTuple _ _ <- cs]) =
+          recurseOnConstrHeads cs
+      | any isWild cs,
+        not (null [() | MatchConstr (ConstrRecord _) _ _ <- cs]) =
+          recurseOnConstrHeads cs
```
athas: Using list comprehensions like this is very weird style. Further, this is basically the same case: check if we have a wildcard pattern and a pattern on a sum type, tuple, or record. This should be reflected in the control flow.

athas (changes requested):

src/Language/Futhark/TypeChecker/Match.hs
CoderDue: Done — introduced an `isStructuralConstr` predicate and collapsed the three separate guards into a single `| any isWild cs, any isStructuralConstr cs` guard. Also replaced the list comprehension in `recurseOnConstrHeads` with `filter isStructuralConstr`.

athas (changes requested):
Add a mention to the CHANGELOG.

PR #2494, opened 2026-06-26
Fix crash when hole with function type is used in syntactic let-binding (#2488)

### LINKED ISSUE #2488: Hole with function type in syntactic function definition crashes the type checker
I noticed something off when I tried to add an argument to a let-defined function binding with at least one syntactic argument.

In my file `Hole.fut` I have the following declaration:

```futhark
def foo =
  let f _ = ???
  in f true false
```

When using `futhark check` to check this program, it produces:
```
checkArgs: *t₁
```

`futhark -V`:
```
Futhark 0.27.0 (prerelease - include info below when reporting bugs).
git: f05e40c (Sun Jun 21 19:12:17 2026 +0200)
Compiled with GHC 9.14.1.

```

CoderDue:
## Summary

Fixes #2488.

When a hole (`???`) is the body of a let-defined function with at least one syntactic argument (e.g. `let f _ = ???`), and the result is applied to multiple arguments (e.g. `f true false`), the consumption checker crashed with an unhandled `error` call.

**Root cause:** The consumption checker's `checkArgs` and `applyArg` functions pattern-matched exclusively on `Arrow` types. When the body of the local function is a hole, its return type is an unconstrained type variable (e.g. `*t₁`). Applying this result to a second argument caused `checkArgs` to hit its `error` fallback (`checkArgs: *t₁`), and then `applyArg` had the same issue.

**Fix:** Handle non-Arrow types (type variables from unconstrained holes) gracefully:
- `applyArg`: return the type with aliases cleared instead of crashing.
- `checkArgs`: fall back to treating remaining arguments as observed (conservative/safe) instead of crashing.

A regression test `tests/holes/letfun0.fut` is included.


CoderDue:
Full test suite run locally with `--backend=interpreter --backend=c`:

```
2441/2441 passed.
```

All tests pass with both the interpreter and C backends.

src/Language/Futhark/TypeChecker/Consumption.hs
```diff
@@ -580,7 +580,10 @@ returnType appres (Scalar (Sum cs)) d arg =
 applyArg :: TypeAliases -> TypeAliases -> TypeAliases
 applyArg (Scalar (Arrow closure_als _ d _ (RetType _ rettype))) arg_als =
   returnType closure_als rettype d arg_als
-applyArg t _ = error $ "applyArg: " <> show t
+-- When the function type is a type variable (e.g. the result of a hole),
+-- we cannot determine the return type, so keep the existing type with no
+-- new aliases.
+applyArg t _ = second (const mempty) t
```
athas, line 586: This is wrong; use the aliases of the argument.

athas (changes requested):

athas:
Actually I will just close this, as I have a cleaner fix.

PR #2498, opened 2026-07-01
benchcmp: add sorting flags, CI grouping, and man page

CoderDue:
## Summary

* Add `--sort-by={significant,geomean-significant,geomean-all}` and `--order={worst-first,best-first}` flags to `futhark benchcmp` for controlling the order program groups are printed. Default is unsorted (preserves original alphabetical behaviour); default order when sorting is worst-first.
* Move the `cmp-benchmark` CI inline script to a proper shell file (`.github/actions/cmp-benchmark/cmp-benchmark.sh`) and wrap benchmark output in three collapsible `::group::` sections — one per sort metric — so CI output has three clearly labelled comparison views.
* Both compiler runs (master and PR) benchmark the same suite, pinned to the submodule commit used by the nightly master build.
* Add a dedicated `futhark-benchcmp(1)` man page documenting the `FILE_A` (baseline) / `FILE_B` (new) argument convention and all options. Registered in `index.rst`, `conf.py`, and `SEE ALSO` of `futhark-bench(1)` and `futhark(1)`.
* Add a CHANGELOG entry for the new sorting options.


.github/actions/cmp-benchmark/action.yml
```diff
@@ -28,7 +28,7 @@ runs:
 
           # Find the benchmark suite commit pinned by the master compiler.
           MASTER_SHA=$(cat /tmp/futhark-master/commit-id)
-          BENCH_SHA=$(git ls-tree "$MASTER_SHA" futhark-benchmarks | awk '{print $3}')
+          BENCH_SHA=$(git ls-tree "$MASTER_SHA" futhark-benchmarks | cut -f3)
```
athas: `cut` by default requires tab-separation. Is that really what `git ls-tree` produces?

.github/actions/cmp-benchmark/action.yml
```diff
@@ -0,0 +1,68 @@
...
+          futhark benchcmp --sort-by=significant old-${{inputs.backend}}.json new-${{inputs.backend}}.json
+          echo "::endgroup::"
+
+          echo "::group::Sorted by geomean - significant only"
+          futhark benchcmp --sort-by=geomean-significant old-${{inputs.backend}}.json new-${{inputs.backend}}.json
+          echo "::endgroup::"
+
+          echo "::group::Sorted by geomean - all datasets"
+          futhark benchcmp --sort-by=geomean-all old-${{inputs.backend}}.json new-${{inputs.backend}}.json
+          echo "::endgroup::"
```
athas: This entire thing is ridiculous. Don't write a long shell script inline a YAML file; we are not unprincipled devops hackers. Put the shell script in a real file and just source it.

.github/actions/cmp-benchmark/action.yml
CoderDue: Yes, `git ls-tree` output is tab-separated (the fields are separated by `\t`, verified locally with `cat -A`). `cut -f3` is correct.

.github/actions/cmp-benchmark/action.yml
CoderDue: Fair point. Moved the script to `.github/actions/cmp-benchmark/cmp-benchmark.sh` — the action now just calls it with the backend and system as arguments.

.github/actions/cmp-benchmark/action.yml
athas: This is incorrect - the three first columns (and you want the third one) are space-separated. Only the file name (the fourth column) is separated from the first three by a tab character.

.github/actions/cmp-benchmark/cmp-benchmark.sh
```diff
@@ -0,0 +1,52 @@
...
+BENCH_SHA=$(git ls-tree "$MASTER_SHA" futhark-benchmarks | awk '{print $3}')
+git -C futhark-benchmarks checkout "$BENCH_SHA"
+cp -r futhark-benchmarks futhark-benchmarks-master
+git -C futhark-benchmarks checkout -
+
+# Download data for each suite independently, as they may differ.
+module load perl
+(cd futhark-benchmarks-master && ./get-data.sh external-data.txt)
+(cd futhark-benchmarks && ./get-data.sh external-data.txt)
+
```
athas: The same benchmark suite should be used; otherwise a comparison is meaningless. In those cases where the benchmark suite differs, it makes no sense to make a comparison this way.

athas (changes requested):

athas:
Should I close this or do you intend to do more with it?

CoderDue:
I tried running it to completion but I think there is something with the hip backend I never looked into, do you not think something like this is worth merging when the kinks are worked out?

athas:
Yes, but are you going to do so, or is this just an abortive experiment?

I don't see why the HIP backend should have any bearing on this work, so any CI issues are likely just cluster noise.

LiljeDue:
I will merge it then I assume nothing is wrong and I think the code qauality seems satisfactory.

PR #2499, opened 2026-07-07
Add OpenBSD support for C targets

omegametabroccolo:
The full test suite passes except `soacs/scan9`, but that seems unrelated to OpenBSD itself:

```
Compiling with --backend=multicore:
Running compiled program:
Running ./tests/soacs/scan9:
Entry point: scan_arr; dataset: 1000000i64 100i64 [100000000]i32:
Cannot load value from generated byte stream:
not enough bytes
```

### DIFF
--- a/rts/c/scheduler.h
+++ b/rts/c/scheduler.h
@@ -101,6 +101,10 @@ static inline int scheduler_execute_task(struct scheduler *scheduler,
 #include <sys/sysinfo.h>
 #include <sys/resource.h>
 #include <signal.h>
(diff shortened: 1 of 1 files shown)


athas:
I do wonder what that error is about. Maybe we assume a 64-bit size a place where it isn't guaranteed? That gave us trouble on Windows at one point, but the test in question does not produce particularly large arrays. Thanks for the PR! I'll just merge it, as we don't test OpenBSD in CI anyway.

PR #2502, opened 2026-07-13
Mistake in Fusion.

WilliamDue:
The splitting of lambdas was incorrect in `fusible` I am sure, and it seems like may be possible to get it to throw an error but I have not created program that would cause an error. The inputs should have the same size as the number of lambda parameters and the outputs should have the same size as the lambda return type. These values are extra and used in certain cases but not for fusible error which is likely why there have been no problems.

I am pretty sure I found another mistake, check for overlapping accumulators in result is redundant because there is performed a partition on the results so they would never overlap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * A case where complex sizes referring to explicit parameters was mishandled by
   monomorphisation (#2230).
 
+* An issue where `#[scratch]` would apply to subexpressions in undesirable ways,
+  changing the type of the expression at the IR level.
+
 ## [0.26.4]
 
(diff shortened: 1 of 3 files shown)


src/Futhark/Optimise/Fusion/Screma.hs
```diff
@@ -141,9 +138,9 @@ fusible ::
   ScremaForm SOACS ->
   [VName] ->
   m ()
-fusible inp_p form_p out_p inp_c form_c out_c = do
+fusible _inp_p form_p out_p inp_c form_c _out_c = do
```
athas: Why have parameters you don't use?

LiljeDue: That is a very fair point I will go remove them.

PR #1740, opened 2022-10-16
Implementation of irregular flattening

athas:
This highly WIP PR contains an implementation of full flattening as a transformation that goes from the SOACS representation to GPU.  It is the result of a few hours of hacking and can handle the following program:

```
def main = map (\n -> #[unsafe] i64.sum (iota n))
```

Flattened versions of all core Futhark constructs must be defined, and so far I have only done Iota and Reduce.  The main design challenge was the representation of irregular arrays in the compiler, as well as the overall structure of the algorithm.  It is currently based on (irregular) distribution, much like the moderate flattening algorithm.  I think that is the best way to do it.

The ultimate goals are:

1. Replace the current implementation of flattening with one that is more principled (the old one is *really* dirty in places because I had no idea what I was doing).
2. Maintain incremental flattening, just in a more principled framework.  The "fully flattened" code version will only be used as a last resort.
3. Experiment with whether (possibly virtualised) full flattening may perhaps have acceptable performance in the *intra-group* case.
4. Support compiling *any* Futhark program to GPU code; not just the subset that uses only regular nested parallelism.
5. Bring back recursive functions.

An explicit *non-goal* is adding support for irregular arrays in the source language.

PR #2505, opened 2026-07-17
Add support for recursive functions.

athas:
This is WIP and intended to see which parts of the compiler needs to be updated. It cannot be merged until flattening is merged in #1740.

Known weaknesses:

* The type checker doesn't do anything with the return type variable; this it should.
* We cannot defunctionalise recursive functions.


athas:
I can't be arsed getting this working well for the current type checker, so I'll use this branch to get the internalisation machinery working, but actual support for recursive functions will also have to wait for #2302.

athas:
Some nasty test cases


```Futhark
def repeat (n: i64) (f: i64 -> i64) (x: i64) : i64 =
  if n == 0
  then x
  else let g = repeat (n - 1) f  -- partial application, bound to 'g'
       in g (f x)                -- applied indirectly, not as 'repeat ...'

entry main (n: i64) (x: i64) = repeat n (\y -> y + 1) x
```

```Futhark
def go (g: i64 -> i64) (n: i64) : i64 =
  if n == 0
  then g 0
  else go (\x -> g x * 2) (n - 1)  -- different function each recursion

entry main (n: i64) = go (\x -> x + 1) n
```

It is easy to avoid the latter with a syntactic check, but the former is a bit nasty, unless we simply require full application in the recursive case.

We always have the option of simply not allowing recursion for higher-order functions, which seems mighty tempting.

athas:
We should make sure this works:

```Futhark
-- Polymorphic size recursion

def sum [n] (xs: [n]i32) : i32 =
  if n == 0 then 0 else xs[0] + sum (tail xs)

entry main (xs: []i32) = sum xs
```

athas:
All the weaknesses have been addressed in `master` and we're not just waiting for flattening. All this PR does is enable recursion unconditionally and add a few tests.

athas:
We might as well merge this and see what happens. Recursion back in Futhark!

PR #2507, opened 2026-07-23
Calling server functions from the interpreter

vox9:


src/Futhark/Eval.hs
```diff
@@ -103,67 +111,129 @@ runExpr (InterpreterState (src, env, ctx)) str = do
...
-
-data InterpreterConfig = InterpreterConfig
-  { interpreterPrintWarnings :: Bool,
-    interpreterFile :: Maybe String
+    Left err -> abort $ I.prettyInterpreterError err
+    Right val -> do
+      val' <-
+        liftIO
+          ( either error id <$> case s of
+              Nothing -> pure $ Right val
```
athas: I modified some code here (it used to be an error), but what is the actual story for using `futhark eval` without a server? I assume it should just work in all cases, and never produce a foreign value, right? Where would it come from?

src/Language/Futhark/Interpreter/Values.hs
```diff
@@ -106,6 +107,8 @@ data Value m
     ValueAcc ValueShape (Value m -> Value m -> m (Value m)) !(Array Int (Value m))
   | -- A primitive value with added information used in automatic differentiation
     ValueAD AD.Depth AD.ADVariable
+  | -- A lazy reference to a value on a Futhark Server
+    ValueLazyFFI ValueShape ValueRef [Int64]
```
athas, line 112: Document the purpose of the `[Int64]`.

src/Language/Futhark/Interpreter/FFI/ServerM.hs
```diff
@@ -0,0 +1,294 @@
...
+    )
+
+startServer :: S.ServerCfg -> IO Server
+startServer cfg = Server <$> S.startServer cfg <*> AL.new
+
+runServerM :: Server -> ServerM a -> IO (Either String a)
+runServerM s (ServerM m) = runExceptT $ runReaderT m s
+
+gc :: ServerM ()
+gc = ServerM $ do
```
athas: This is not how you use a monadic abstraction. Use the `ServerM` value constructor only to define a few simple building blocks (like `askServer :: ServerM Server`) instead of like this.

PR #2512, opened 2026-07-24
Use device-wide atomic scope on OpenCL 2.0 or later

### LINKED ISSUE #734: Fix global memory fences in OpenCL kernels on NVIDIA
I [found out an awful truth](https://github.com/diku-dk/futhark/commit/49214b807a9bffc339c2c4bc1ac4610125e41509#commitcomment-32890492) that on NVIDIA GPUs, our supposed global memory fences only take effect within the current workgroup (or more likely, SM).  This is *not* an NVIDIA bug - it is us who are technically violating the OpenCL spec.  The barriers work fine with the CUDA backend, because CUDA has fewer restrictions on cross-workgroup synchronisation.  The solution is to generate generate inline assembly with an explicit `membar.gl` instruction when we are running on an NVIDIA GPU.  I think the solution is for us to define our own wrapper function:

```
void mem_fence_global() {
#ifdef NVIDIA
  asm("membar.gl;");
#else
  mem_fence(CLK_GLOBAL_MEM_FENCE);
#endif
}
```

The only trick is to have some sort of preprocessor symbol (here, NVIDIA) that is set when we detect an NVIDIA platform.  I couldn't find any documentation about NVIDIA setting this themselves, so we'll have to invent our own policy.

rowanG077:
Atomics now use device-wide memory scope when OpenCL 2.0 or later is available, fixing cross-workgroup synchronization in both the C and Python OpenCL runtimes.

(re-)Fixes https://github.com/diku-dk/futhark/issues/734

I'm definitely not a opencl expert, this is basically the second time I ever had to move down to this level. so please review it carefully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,8 +30,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Simplified fusibility check by removing redundant accumulator overlap check
   and fixed fusibility check by giving the correct number of elements.
 
+
 * A compiler crash due to missing double buffering inside sequential code
   migrated to GPU. (#2513)
+
+* Atomics now use device-wide memory scope when compiling for OpenCL C 2.0, or
+  OpenCL C 3.0 with the required atomic features. OpenCL C 2.0 is selected
+  automatically for Rusticl on Asahi; elsewhere it must be selected explicitly.
+  This fixes incorrect results from cross-workgroup operations
+  ([#734](https://github.com/diku-dk/futhark/issues/734)).
 
 ## [0.26.4]
 
--- a/rts/c/atomics32.h
+++ b/rts/c/atomics32.h
@@ -28,6 +28,10 @@ SCALAR_FUN_ATTR int32_t atomic_xor_i32_shared(volatile __local int32_t *p, int32
 SCALAR_FUN_ATTR int32_t atomic_xchg_i32_global(volatile __global int32_t *p, int32_t x) {
 #if defined(FUTHARK_CUDA) || defined(FUTHARK_HIP)
   return atomicExch((int32_t*)p, x);
+#elif defined(FUTHARK_OPENCL_DEVICE_ATOMICS)
+  return atomic_exchange_explicit((volatile atomic_int*)p, x,
+                                  memory_order_acq_rel,
+                                  memory_scope_device);
 #else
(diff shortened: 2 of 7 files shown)


rowanG077:
Arghh, while it works if manually setting opencl to 2.0 rusticl reports 1.2 in `CL_DEVICE_OPENCL_C_VERSION`. What a total shitshow this is. I will make a small probe kernel that tests if the relevant ops are available.

athas:
I really don't like the idea of a probe kernel. Is that truly the best way to solve this? Is rusticl not reporting the right OpenCL C version?

rowanG077:
rusticl reports 1.2 with `CL_DEVICE_OPENCL_C_VERSION` There is also the opencl 3.0 only `CL_DEVICE_OPENCL_C_ALL_VERSIONS` which returns:

```
  OpenCL C 3.0
  OpenCL C 1.2
  OpenCL C 1.1
  OpenCL C 1.0
```

It seems to not return 2.0 since not all required features are supported. 3.0 mode those features optional. An alternative would perhaps be to:

1. Check `CL_DEVICE_OPENCL_C_VERSION` if 2.0 then good we can continue.
2. If not check `CL_DEVICE_OPENCL_C_ALL_VERSIONS`. If 2.0 is listed then great and we can continue. If 3.0 is listed we can try to query `CL_DEVICE_OPENCL_C_FEATURES` and check for `__opencl_c_atomic_order_acq_rel` and `__opencl_c_atomic_scope_device`.

I can check if that works.

rowanG077:
Still doesn't work. `__opencl_c_atomic_order_acq_rel` and `__opencl_c_atomic_scope_device` are not set for rusticl...

I have removed all detection from this PR. users/I will just manually have to set the opencl 2.0 flag.

athas:
How about detecting rusticl based on the name? It's no different than our detection of NVIDIA.

rowanG077:
rusticl has many different backends. Not only asahi where this is supported. It;s essentially the generic opencl frontend in mesa.

athas:
That does not prevent detection based on the name of the device.

rowanG077:
I considered that to hacky, but if you're fine with. I have added detection purely based on whether it's rusticl + Apple M GPU.

athas:
Thanks for the fix! It is nice to see it could be done without touching compiler code.

PR #2519, opened 2026-08-05
Fix uniqueness violation in diffUpdateAcc for Safe scatter over 2D arrays

CoderDue:
## Summary

When applying \`vjp\` to a \`scatter\` over a \`[n][k]f64\` array with a concatenated index array (e.g. \`(0..<n) ++ (0..<n)\`), the compiler crashed with an internal type error after the \`ad\` pass: \`Variable X referenced after being consumed\`.

**Root cause:** In \`diffUpdateAcc\`, the \`Safe\` branch built an \`if\`-expression whose **true branch body** referenced \`adj\` twice in a sequential scope: first observing it via \`Index adj slice\` (to read out \`adj_i\`), then consuming it via \`Update Unsafe adj slice zero\` (to produce \`acc_adj\`). Even though cross-branch consumption is legal in Futhark (it is fine for one branch to consume a variable that the other only observes), within a single branch body the two uses are sequenced, and the type checker fires when a variable is consumed after being observed in the same scope. The bug only triggers for 2D (or higher) element types because those yield \`Safe\` \`UpdateAcc\` statements; 1D scatters use \`Unsafe\` and are unaffected.

**Fix:** Made \`zeroed\` take the array name as a parameter. In the \`Safe\` branch, a copy of \`adj\` (\`adj_copy\`) is made before the \`if\` so the true branch consumes the copy via \`Update Unsafe adj_copy slice zero\` while still observing the original \`adj\` via \`Index adj slice\`. The false branch returns the original \`adj\` unchanged. This eliminates the within-branch conflict entirely.

**Concern:** The copy added in the \`Safe\` branch should be eliminated by simplification in normal cases, but could leave a constant memory overhead if simplification doesn't fire. Also, the \`copyConsumedArrsInStm\` mechanism handles this class of problem for the forward pass generically — generated return-sweep code is not covered by it, so similar issues in other \`diff*\` functions would need their own fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Fixed
 
+* A compiler crash (internal type error after the `ad` pass) when applying
+  reverse-mode AD to a `scatter` over a 2D array with a concatenated index
+  array.  The reverse pass generated an `if`-expression that consumed the
+  accumulator adjoint in the true branch via `Update` while the false branch
+  also referenced it, violating uniqueness.
+
 * A case where complex sizes referring to explicit parameters was mishandled by
   monomorphisation (#2230).
 
--- a/src/Futhark/AD/Rev/Acc.hs
+++ b/src/Futhark/AD/Rev/Acc.hs
@@ -392,22 +392,25 @@ diffUpdateAcc pat aux safety acc is vs m = do
         -- zero sensitivity. XXX: this is only true for scatter-like
         -- accumulators, but these are probably the only ones that need this
         -- handling, as they are all that appear in source programs.
-        zeroed = do
+        zeroed a = do
           z <- letSubExp "acc_adj_zero" $ zeroExp elem_t
-          pure $ BasicOp $ Update Unsafe adj slice z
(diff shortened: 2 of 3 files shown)


athas:
The explanation of the bug is not sensible. It is valid for one branch to consume a variable and the other to observe it.

athas (changes requested):
The explanation for the bug is not correct.

CoderDue:
You're right, thank you for the correction. Cross-branch consumption is perfectly legal — it's fine for one branch to consume a variable that the other only observes.

The actual violation is within the **true branch body** itself: it references `adj` twice in a sequential scope — first observing it via `Index adj slice` (to produce `adj_i`), then consuming it via `Update Unsafe adj slice zero` (to produce `acc_adj`). The type checker's `seqOccurences` check fires because `adj` is consumed after being observed in the same body. This only manifests in the `Safe` path because 2D (or higher) element types cause `Update` to consume the array; 1D scatters take the `Unsafe` path and are unaffected.

The fix — copying `adj` before the `if` so the true branch consumes the copy while both branches freely observe the original — is correct for this reason. I've updated the PR description accordingly.

athas:
The bug is real, but the explanation keeps being nonsensical. I will fix this on my own.

PR #2527, opened 2026-08-11
Single kernel partition and remove redundant computation in filter.

LiljeDue:
This pull request changes the following:               
* Rewrote partition to be a single kernel.
* Filter now uses the predicate once per element instead of twice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,9 +35,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 * Filter now uses the predicate once per element instead of twice.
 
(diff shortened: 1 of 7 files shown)


athas:
Can you split the bugfix into a separate PR?

LiljeDue:
Yes I can, but I have no other program that reproduces the error than the partition found here, for some reason when I move the partition out of prelude it stops being an error. Also it will have to be later today I do not have time right now.

LiljeDue:
I have made this pull request specific to the changes in the SOACS, but I would like to highlight it will not work before the bug fix is implemented in my other pull request.

athas:
Rebase on master.

PR #2529, opened 2026-08-12
Add a semipartition

LiljeDue:
This pull request adds a semipartition and tests, it matches the partition implementation found in CUB.
True elements are ordered relative to their indices but false elements are in reverse order relative to their indices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Added
 
+* Added semipartition, faster than partition but it is not stable for
+  false elements.
+
 * Rewrote partition to be a single kernel.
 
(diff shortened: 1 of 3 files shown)


athas:
I am pretty sure that we should merge this, but at some point we need to think about what belongs in the prelude. This doesn't *have* to be there, does it? Arguably, neither does partition.

LiljeDue:
I think I am fine with it not being in prelude since it is quite similar to partition. Both partitions are single kernels but partition does 3n io operations while semipartition does 2n. This seems like it can not be that bad of a difference? Maybe different partitions could be useful to have in the Futhark examples?

athas:
Sure, having them as examples is a good idea, but I'm increasingly lukewarm on having a function that relaxes properties in return for being faster - especially when `semipartition` does not exploit any intrinsics. The main reason we have `partition` is that we always had `partition` (it was a builtin SOAC in the very early days). Today I would not add it.

PR #2530, opened 2026-08-25
Fix #1675.

### LINKED ISSUE #1675: Should functions not be allowed to return aliases by default?
I am plugging some holes in alias tracking, and I discovered something uncomfortable.  We currently have a rule that functions may not return aliases of global variables.  This rule was instituted such that we can assume that the value returned by e.g. `transpose A` can *at most* alias `A`, which allows us to make in-place updates to transposed arrays, as long as we can consume the original array.

Now consider the following program:

```Futhark
module type monoid = {
  type t
  val zero : t
  val plus : t -> t -> t
}

module pm (M: monoid) = {
  def f (x: M.t) : M.t =
    M.plus M.zero x
}
```

While this is currently allowed by the type checker due to a bug, it should actually be a type error.  The reason is that the result of the `f` function aliases `M.zero` (a global variable), which we do not want to allow.  This is because the `plus` function does not declare its return type to be unique/alias-free, meaning it is allowed to alias either of its arguments.  We may know that it is unlikely that a monoidal plus operation will do that, but the type checker does not.  Worse, *all* of our builtin mathematical module types define their operations like this, and allow the result to alias the operands.  Fixing this bug will require fixing a significant number of programs.  Now, that's doable.  The fix is to add an asterisk to the return types, indicating that the result is alias-free:

```Futhark
module type monoid = {
  type t
  val zero : t
  val plus : t -> t -> *t
}
```

...but to be honest, I think this adds a lot of visual clutter *everywhere*.  I am considering whether we should change the default behaviour.  Maybe `t -> t` should denote a function that *does not* alias any of its parameters (i.e make our current treatment of `*` in return types the default).  We could then use another symbol, say `@`, to indicate that the result may alias the parameters (but still not global variables).  For example, `transpose` would then be defined as

```Futhark
def transpose 'a [n] [m] (A: [n][m]a) : @[m][n]a = ...
```

This is based on the assumption (which I have not yet checked) that the *common case* is that functions return fresh results, and that aliasing is somewhat rare.

This is a pretty major bug that we should fix quickly, but the fix could have significant impact on the language.  @melsman?  @coancea?  Anyone else?  I'd like input.

athas:
While this fixes a real soundness bug in the type checker, I am somewhat concerned by the amount of code this breaks. A lot of polymorphic module-level code played fast and loose with uniqueness and aliases of abstract types. While I think this is overall the simplest solution, I would like to see if we can make some improvements to alias analysis to limit the scope of the breakage.


FluxusMagna:
How does this interact with SOACs and other higher order functions, like `map`? 
Will they be polymorphic in the functional argument and transfer uniqueness properties from the function to the arrays they operate on? Can this be annotated with the current type syntax?

athas:
It does not change anything about SOACs or higher-order functions specifically. This only affects the aliases of top level bindings of values of abstract types (meaning abstract through the module system, as type parameters do not exist at top level).

athas:
This is turning into a significant overhaul of aliasing/consumption in the source language. It does the following:

* Fixes #1675, which is pretty simple.
* Modifies types in the prelude to make programming with the #1675 fix more ergonomic, as that fix causes more things to be aliased.
* Fixes #2531, which on its own is not so difficult, but the naive fix makes the ergonomics quite poor. See that issue for details.
* A yet-undetermined amount of changes to alias inference to fix the ergonomic problems. A lot of this is based on exploiting parametricity. For ease of implementation, some of the current adjustments are based on *syntactic* rules as well.

athas:
Cosmin suggested an interesting simplification: ban self-aliasing in any function return value. I suspect it is too excessive, but let's see.

PR #2532, opened 2026-09-02
Set up framework for IR interpreter.

athas:


athas:
@armadi1809 This will be the basis for your work.

PR #2533, opened 2026-09-06
Add program and top-level profiling indexes

### LINKED ISSUE #2347: futhark profile: generate more index files
Currently `futhark profile` produces an HTML index file for each dataset. In addition, I would like:

* An index file for each program, with links to the per-dataset index files.
* A top level index file with links to each program index file.

The idea is to make everything navigable via a browser. It is easy to imagine an automated benchmark setup that publishes these profiling reports on the web.

emecii:
Closes #2347.

Add a top-level `index.html` linking program indexes, and a program index linking each successfully generated dataset report. Users can open one page to browse a profiling run. Labels are HTML-escaped and relative URL paths are UTF-8 percent-encoded; datasets without an HTML report are omitted from the links.

Retain the program basename when stripping the shared source prefix, so a single program's index cannot overwrite the top-level index. This changes the output-directory layout for single-program/shared-basename-prefix inputs. Existing dataset report contents are unchanged; the four-dataset comparison preserved all 24 existing artifacts byte-for-byte.

Validation on macOS arm64 with GHC 9.10.3:

- Five synthetic integration tests fail on the original executable and pass with the change, including nested programs, multiple entry points, missing profiles, special characters, and empty input. Independently rerun successfully.
- Existing benchmark/profile property passes 100 generated cases. The existing C and multicore profiling fixture passes; OpenCL was omitted.
- Executable/unit build, edited-module `-Wall -Werror`, and the repository style checker on all four changed files pass.

Full compiler tests, full `make check`, GPU/OpenCL, and other OSes were not run. The existing detail-page URL generation is unchanged.

AI assistance: OpenAI Codex (GPT-6) was used for implementation, review, and test execution.

### DIFF
--- a/src/Futhark/CLI/Profile.hs
+++ b/src/Futhark/CLI/Profile.hs
@@ -416,10 +416,10 @@ analyseBenchResults json_path bench_results = do
   programs <- mapM (onBenchResult top_dir) bench_results
   writeNavigationIndex (top_dir </> "index.html") "Program Index" programs
   where
-    prefix =
-      takeDirectory $
-        longestCommonPrefix $
-          map (takeWhile (/= ':') . benchResultProg) bench_results
+    programPaths = map (takeWhile (/= ':') . benchResultProg) bench_results
+    prefix = case S.toList $ S.fromList programPaths of
+      [path] -> path
+      paths -> takeDirectory $ longestCommonPrefix paths
 
     -- Eliminate characters that are filesystem-meaningful.
     escape '/' = '_'
@@ -431,7 +431,15 @@ analyseBenchResults json_path bench_results = do
     onBenchResult top_dir (BenchResult prog_path data_results) = do
       let (prog_path', entry) = span (/= ':') prog_path
           prog_name = makeRelative prefix prog_path'
-          prog_dir = top_dir </> dropExtension prog_name </> drop 1 entry
+          relative_dir = dropExtension prog_name </> drop 1 entry
+          -- Preserve the established <entry>/<dataset>-index.html layout for
+          -- one source file. A file without an entry point still needs its
+          -- own directory so it does not collide with the top-level index.
+          prog_dir =
+            top_dir
+              </> if null relative_dir || relative_dir == "."
+                then "program"
+                else relative_dir
       createDirectoryIfMissing True prog_dir
       datasets <-
         catMaybes <$> mapM (onDataResult prog_dir (T.pack prog_name)) data_results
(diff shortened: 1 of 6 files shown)


src/Futhark/CLI/Profile.hs
```diff
@@ -424,29 +431,60 @@ analyseBenchResults json_path bench_results = do
...
           let tf =
                 TargetFiles
                   { summaryFile = name' <> ".summary",
                     timelineFile = name' <> ".timeline",
                     htmlIndexFile = name' <> "-index.html",
                     htmlDir = name' <> ".html/"
                   }
-           in writeAnalysis tf r
+          writeAnalysis tf r
+          exists <- doesFileExist $ htmlIndexFile tf
```
athas: I don't understand the logic here. `writeAnalysis` unconditionally writes the HTML index file as far as I understand it. And even if it didn't, querying file system state (for generated files!) to make logic choices is unlikely to be right for a program like `futhark profile`.

athas (changes requested):

tests_adhoc/profile-index/test.py
```diff
@@ -0,0 +1,179 @@
...
+    def setUp(self):
+        self.temp = tempfile.TemporaryDirectory()
+        self.addCleanup(self.temp.cleanup)
+        self.root = Path(self.temp.name).resolve()
+
+    def profile(self, programs):
+        input_path = self.root / "results.json"
+        input_path.write_text(json.dumps(programs))
+        run = subprocess.run(
+            [os.environ.get("FUTHARK", "futhark"), "profile", str(input_path)],
```
athas: Same as in #2534, I'd rather we just use PATH for now.

athas (changes requested):

tests_adhoc/profile-index/test.py
emecii: Updated to use futhark from PATH in d79c79500.

emecii:
The rebased branch moved single-program reports from `main/` to `prog/main/`, which caused the tools failures. Commit 585d94f5a preserves the existing single-program layout while retaining distinct directories for multi-program reports. The focused profile-index and profile-html suites now pass locally (5 + 6 tests).

emecii:
The remaining `test-tools` failure came from the Unicode escaping case running under an ASCII locale. Commit 89bd97355 gives the subprocess a UTF-8 locale; all 5 profile-index and 6 profile-html tests pass locally, including with the parent test process under `LC_ALL=C`.

athas:
That looks like a fix, but I want to see if we can force a UTF-8 locale in the runner, since this is unlikely to be the last time this bites us.

athas:
I really dislike locale errors... astounding that software doesn't just default to UTF-8 in all cases.

`futhark` itself (which is what is crashing) actually has code to force a UTF-8 locale [right here](https://github.com/diku-dk/futhark/blob/4ed18c8a51da89c3d7e254546e903ae67a438753/src/Futhark/CLI/Main.hs#L150-L152), but perhaps it does not apply to whatever is doing file name encoding.

Complicating the issue, I cannot reproduce it locally. How annoying.

athas (approved):

PR #2534, opened 2026-09-07
futhark profile: add logs and timeline to HTML reports

### LINKED ISSUE #2348: futhark profile: make logs and timeline also available in HTML format
Like #2347, I think all information should be available in the HTML files.

emecii:
Closes #2348.

`futhark profile` already writes each dataset's execution log and event timeline
as text files, but its HTML landing page only linked to source annotations. This
adds Log and Timeline sections to every dataset HTML report, so the browser view
contains the same text as the companion files.

The report is now still useful when source files cannot be loaded: it renders the
log and timeline and explains that source annotation is unavailable. A dataset
with only a log also receives an HTML report. Failed datasets and datasets with
neither a log nor a profile still do not create one.

The new end-to-end tests exercise the actual CLI with benchmark and raw-profile
JSON, including escaping, Unicode, empty and missing fields, invalid sources,
and repeated invocations.

Validation:

* `cabal build exe:futhark`
* `FUTHARK="$(cabal list-bin exe:futhark)" python3 tests_adhoc/profile-html/test.py` (6 passed)
* `cabal test unit --test-show-details=direct` (637 passed)
* `tools/style-check-file.sh src/Futhark/CLI/Profile.hs`
* `tools/style-check-file.sh src/Futhark/Profile/Html.hs`

### DIFF
--- a/tests_adhoc/profile-html/test.py
+++ b/tests_adhoc/profile-html/test.py
@@ -1,0 +1,223 @@
+"""Test the real profiling CLI's HTML logs and timelines without a GPU."""
+
+from __future__ import annotations
+
+from html.parser import HTMLParser
+import json
+import os
+from pathlib import Path
+import subprocess
+import tempfile
+import unittest
+from urllib.parse import unquote, urlsplit
+
+
+class Page(HTMLParser):
+    def __init__(self, path: Path):
+        super().__init__()
+        self.tags: list[str] = []
+        self.ids: list[str] = []
+        self.links: list[str] = []
+        self.blocks: dict[str, str] = {}
+        self.block: str | None = None
+        self.text = ""
+        self.feed(path.read_text(encoding="utf-8"))
+
+    def handle_starttag(self, tag: str, attrs: list[tuple[str, str | None]]):
+        self.tags.append(tag)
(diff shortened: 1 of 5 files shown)


tests_adhoc/profile-html/test.py
```diff
@@ -0,0 +1,223 @@
...
+        self.temp = tempfile.TemporaryDirectory()
+        self.addCleanup(self.temp.cleanup)
+        self.root = Path(self.temp.name).resolve()
+
+    def run_profile(self, payload, filename="results.json"):
+        path = self.root / filename
+        original = json.dumps(payload, ensure_ascii=False)
+        path.write_text(original, encoding="utf-8")
+        run = subprocess.run(
+            [os.environ.get("FUTHARK", "futhark"), "profile", str(path)],
```
athas: I don't think we have precedence for using a `FUTHARK` environment variable to point at the binary - all the other tests just do a `PATH` lookup. Arguably allowing a variable would be nicer, but I'd rather have consistency for now.

athas (changes requested):
This is a good improvement, and I have only a minor request.

PR #2535, opened 2026-09-08
Replace "uniqueness" terminology with consumption/fresh.

athas:
We were already halfway there, as "uniqueness types" has for a long time been a poor way of describing what Futhark's type system is actually like. This is worth emphasizing: it is not uniqueness types *at all*, and probably has never been, possibly except for the very earliest days. Hence it is simply wrong to use that terminology, and it is high time we change it.

A function parameter may now be "consuming" (or "consumed") or "observing".

A function result may now be "fresh" or "nonfresh".

Whether a type is consuming/observing or fresh/nonfresh is called its "mode", although this is mostly internal lingo. It does not occur anywhere in the language specification.

This is mostly a conceptual change, so it does not actually affect code at all. The one exception is that it requires a slight change to the manifest file format.


athas:
@melsman Do you have any thoughts about this?

athas:
I have rewritten the documentation, but it turned out to not be a large change, as I have shied away from the "uniqueness" nomenclature for years anyway.

PR #2537, opened 2026-09-10
Fix reverse AD for array-valued scans

### LINKED ISSUE #2208: AD crash
This produces invalid code after AD:

```
entry scan_arr_add [n]
                   (inp: [2][n]f32)
                   (adj: [2][n]f32) : [2][n]f32 =
  let adj =
    vjp (scan (\x y -> [x[0] + y[0], x[1] + y[1]])
              (replicate 2 0))
        (transpose inp)
        (transpose adj)
  in transpose adj
```

The problem is that some part of the differentiation rule assumes that the operands are never arrays.

emecii:
Fixes #2208.

Reverse AD can select the scalar Jacobian specialization for an array-valued scan operator, producing an internal type error. Check all scan result types before constructing that Jacobian and use the existing PPAD algorithm when any result has positive rank. Scalar selection is unchanged.

Adds 17 datasets covering the reported array addition, empty/singleton inputs, noncommutative affine composition with signed adjoints and a zero multiplier, mixed scalar/array types, rank-two results, and a nested Hessian-vector product. Expected derivatives are analytical and independently checked with finite differences. Four programs fail on the base revision; the mixed-type tuple is an existing-behavior check.

Validation on macOS arm64, GHC 9.10.3:

- All 184 AD programs and 56 additional scan programs pass on both C and multicore backends.
- All 637 unit tests pass.
- `cabal build all --ghc-options=-Werror` and `cabal check` pass.
- All 393 source/style checks pass with the repository-pinned Ormolu 0.8.0.2; new fixture style and whitespace checks pass.
- Three representative scalar scans produce byte-identical optimized SOACS before and after the fix.

GPU backends and performance were not measured. This restores the existing general algorithm for array results; it does not establish support for every possible array-valued operator.

Prepared and tested with assistance from OpenAI Codex.

### DIFF
--- a/src/Futhark/AD/Rev/Scan.hs
+++ b/src/Futhark/AD/Rev/Scan.hs
@@ -247,6 +247,10 @@ cases d t mat = case subMats d mat $ zeroExp t of
 -- Jacobian of the scan op. Figure out if the Jacobian has some
 -- special shape, discarding the temporary lambda.
 identifyCase :: VjpOps -> Lambda SOACS -> ADM ScanAlgo
+identifyCase _ lam
+  -- The IFL23 specialisation represents Jacobian entries as scalars.
+  -- Use PPAD for array results, before constructing the Jacobian.
+  | any ((> 0) . arrayRank) (lambdaReturnType lam) = pure GenericPPAD
 identifyCase ops lam = do
   let t = lambdaReturnType lam
   let d = length t
--- a/tests/ad/issue2208.fut
+++ b/tests/ad/issue2208.fut
@@ -1,0 +1,21 @@
+-- Array-valued scan from #2208. Each derivative is a suffix sum of adj.
+-- ==
+-- tags { autodiff }
+-- entry: scan_arr_add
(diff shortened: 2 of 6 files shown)


athas:
Thanks for the fix. Does this also work in the vector-mode case when using `mjp`? The tests should certainly contain such a variant, like most of the other such tests (e.g. see `tests/ad/scan0.fut`). If it does not immediately work, then the best solution is to use `locallyNonvector` to disable vector-mode for array scans. (I think we may actually do that anyway.)

emecii:
Yes. `diffScan` already wraps the scan reverse pass in `locallyNonvector`, so the array-result PPAD selection also works when reached through `mjp`. I added `scan_arr_add_vec` in `tests/ad/issue2208.fut`, with two independent cotangent batches, in 0ff27fa3f.

The isolated `mjp` case fails with the pre-fix compiler at the AD pass with the original array/scalar type mismatch and passes with this branch. Validation on the updated test: focused C and multicore both pass; full `tests/ad` passes 184/184 on both C and multicore; the fixture style check and `git diff --check` pass.

